PR #66, opened 2014-12-30
Converter 2.0

Lithopsian:
This is a possible replacement for the current converter plugin.  The API is not backward compatible, hence 2.0.  It has an updated UI, multiple threads, enhanced progress window with progress tracking, and fixes several bugs.


Lithopsian:
I'm not sure what code you're testing.  A couple of the widgets lacked spacing (the base folder stuff that was changed very late), but the others are spaced.  The dialog borders are exactly as they were before I started.  The dialog itself does not have a border, but the vbox that contains all the widgets does have a 12px border.

Alexey-Yakovenko:
I'll send you a screenshot tomorrow.

Alexey-Yakovenko:
nvm about border around window. I compared "old and new" + some other dialogs in deadbeef. some of them have 12px border correctly, and some don't, including the original converter gui. I'll fix them all in one swipe later.

PR #69, opened 2015-01-09
converter 2.0

Lithopsian:
Some minor spacing changes, otherwise as before but hopefully synced and mergeable.


Alexey-Yakovenko:
Starting to review this pull request.
Please add a bulleted list of significant changes in the commit message

plugins/converter/callbacks.h
```diff
@@ -1,128 +1 @@
 #include <gtk/gtk.h>
-
```
Alexey-Yakovenko: please delete the callbacks.h, if it only contains this one line

plugins/converter/converter.c
```diff
@@ -1,48 +1,55 @@
...
-#endif
-#if HAVE_SYS_CDEFS_H
-#include <sys/cdefs.h>
-#endif
-#if HAVE_SYS_SYSLIMITS_H
-#include <sys/syslimits.h>
-#endif
+
+#define _FILE_OFFSET_BITS 64
+
```
Alexey-Yakovenko: Please add a comment why this define is needed.

plugins/converter/converter.c
```diff
@@ -1,48 +1,55 @@
...
-#include "converter.h"
+#include <limits.h>
+#include <errno.h>
+#include <libgen.h>
 #include "../../deadbeef.h"
-#include "../../strdupa.h"
+#include "converter.h"
+
+#ifndef P_tmpdir
+    #define P_tmpdir "/tmp"
```
Alexey-Yakovenko: It's ok to just use "/tmp" inline in the code, as having it in the global define only reduces readability.

plugins/converter/converter.c
```diff
@@ -50,171 +57,109 @@
...
-    char str[1024];
-    while (fgets (str, sizeof (str), fp)) {
-        // chomp
-        char *cr = str + strlen (str) - 1;
-        while (*cr == '\n') {
-            cr--;
-        }
-        cr++;
-        *cr = 0;
+    char* dir = strdup(path);
```
Alexey-Yakovenko: strdupa should work better here

plugins/converter/converter.c
```diff
@@ -50,171 +57,109 @@
...
-        else if (!strcmp (str, "tag_flac")) {
-            p->tag_flac = atoi (item);
-        }
-        else if (!strcmp (str, "tag_oggvorbis")) {
-            p->tag_oggvorbis = atoi (item);
-        }
+static char *
+preset_path(const char *dir, const char *subdir, const char *title)
+{
+    char *path = malloc(strlen(dir) + strlen(subdir) + strlen(title) + sizeof("%s/%s/%s.txt"));
```
Alexey-Yakovenko: alloca or variable size array must be used here

plugins/converter/converter.c
```diff
@@ -50,171 +57,109 @@
...
 
-    if (!p->title) {
-        p->title = strdup ("Untitled");
-    }
-    if (!p->ext) {
-        p->ext = strdup ("");
-    }
-    if (!p->encoder) {
-        p->encoder = strdup ("");
+    if (sprintf(path, "%s/%s/%s.txt", dir, subdir, title) < 0 || !ensure_dir(path)) {
```
Alexey-Yakovenko: snprintf should be used instead

plugins/converter/converter.c
```diff
@@ -50,171 +57,109 @@
...
-    }
-    const char *confdir = deadbeef->get_config_dir ();
-    char path[PATH_MAX];
-    if (snprintf (path, sizeof (path), "%s/presets", confdir) < 0) {
-        return -1;
+static void
+encoder_preset_free_strings(ddb_encoder_preset_t *p)
+{
+    if (p->title) {
+        free(p->title);
```
Alexey-Yakovenko: NULL should be assigned to the freed pointer (same to the code below, in the same function)

plugins/converter/converter.c
```diff
@@ -50,171 +57,109 @@
...
-    if (!p->title || !p->title[0]) {
-        fprintf (stderr, "encoder_preset_save: empty title\n");
-        return -1;
-    }
-    const char *confdir = deadbeef->get_config_dir ();
-    char path[PATH_MAX];
-    if (snprintf (path, sizeof (path), "%s/presets", confdir) < 0) {
-        return -1;
+static void
+encoder_preset_free_strings(ddb_encoder_preset_t *p)
```
Alexey-Yakovenko: perhaps a better name for this would be "encoder_preset_free_data"

plugins/converter/converter.c
```diff
@@ -223,1019 +168,1224 @@ encoder_preset_copy (ddb_encoder_preset_t *to, ddb_encoder_preset_t *from) {
...
-    while (prev && prev->next != p) {
-        prev = prev->next;
-    }
-    if (prev) {
-        prev->next = p->next;
-    }
-    else {
-        encoder_presets = p->next;
+static ddb_encoder_preset_t *
+encoder_preset_get(const char *title)
```
Alexey-Yakovenko: space is missing before the "("
this is not really enforced, but is nice to match the rest of the code

plugins/converter/converter.c
```diff
@@ -50,171 +57,109 @@
...
-        if (p->ext) {
-            free (p->ext);
-        }
-        if (p->encoder) {
-            free (p->encoder);
-        }
-        free (p);
+static int
+check_dir(const char *path)
+{
```
Alexey-Yakovenko: the "{" should go on the same line, to match the rest of the code (same in other similar places)

plugins/converter/converter.c
```diff
@@ -223,1019 +168,1224 @@ encoder_preset_copy (ddb_encoder_preset_t *to, ddb_encoder_preset_t *from) {
...
+    uint8_t *wavehdr = outfmt.is_float || outfmt.channels > 2 ? wavehdr_ext : wavehdr_pcm;
+    const size_t wavehdr_size = wavehdr == wavehdr_pcm ? sizeof(wavehdr_pcm) : sizeof(wavehdr_ext);
+
+    const int readsize = CONVERT_SAMPLES * fileinfo->fmt.channels * fileinfo->fmt.bps / 8;
+    read_buffer = malloc(readsize);
+    if (dsp_preset && dsp_preset->chain) {
+        dsp_chain = dsp_plugin_duplicate(dsp_preset->chain);
+        const size_t dspsize = CONVERT_SAMPLES * ceil((double)384000 / fileinfo->fmt.samplerate) * 8 * 32 / 8;
+        write_buffer = malloc(dspsize);
+        dsp_buffer = malloc(dspsize);fprintf(stderr, "readsize=%d, dspsize=%d\n", readsize, dspsize);
```
Alexey-Yakovenko: fprintf should go on a separate line

plugins/converter/converter.c
```diff
@@ -223,1019 +168,1224 @@ encoder_preset_copy (ddb_encoder_preset_t *to, ddb_encoder_preset_t *from) {
...
+        if (!dsp_buffer) {
+            goto error;
+        }
+    }
+    else if (fileinfo->fmt.bps != outfmt.bps || fileinfo->fmt.is_float != outfmt.is_float) {
+        const size_t writesize = CONVERT_SAMPLES * outfmt.channels * outfmt.bps / 8;
+        write_buffer = malloc(writesize);fprintf(stderr, "readsize=%d, writesize=%d\n", readsize, writesize);
```
Alexey-Yakovenko: fprintf should go on a separate line

plugins/converter/converter.c
```diff
@@ -223,1019 +168,1224 @@ encoder_preset_copy (ddb_encoder_preset_t *to, ddb_encoder_preset_t *from) {
...
+    }
+    else {
+        write_buffer = read_buffer;fprintf(stderr, "readsize=%d\n", readsize);
```
Alexey-Yakovenko: fprintf should go on a separate line

plugins/converter/converter.c
```diff
@@ -223,1019 +168,1224 @@ encoder_preset_copy (ddb_encoder_preset_t *to, ddb_encoder_preset_t *from) {
...
+    }
+    if (!write_buffer) {
+        goto error;
+    }
+
+    const size_t callback_interval = deadbeef->pl_get_item_duration(it) * fileinfo->fmt.samplerate / 15;
```
Alexey-Yakovenko: the magic number 15 should be defined as a constant

plugins/converter/converter.c
```diff
@@ -223,1019 +168,1224 @@ encoder_preset_copy (ddb_encoder_preset_t *to, ddb_encoder_preset_t *from) {
...
+    size_t samples = 0;
+    size_t next_sample = 0;
+    time_t next_time = start_time + 1;
+
+    off_t outsize = 0;
+    int sz = decoder->read(fileinfo, read_buffer, readsize);
+    while (*api == DDB_CONVERT_API_CONTINUE && sz > 0) {
```
Alexey-Yakovenko: I find it bad to mention the "API" in the API declarations, except when it's used to talk about the API (e.g. API versions).
Makes sense to call it simply "DDB_CONVERT_CONTINUE"

plugins/converter/converter.c
```diff
@@ -223,1019 +168,1224 @@ encoder_preset_copy (ddb_encoder_preset_t *to, ddb_encoder_preset_t *from) {
...
+            memcpy(&dspfmt, &fileinfo->fmt, sizeof(ddb_waveformat_t));
+            dspfmt.bps = 32;
+            dspfmt.is_float = 1;
+            sz = deadbeef->pcm_convert(&fileinfo->fmt, read_buffer, &dspfmt, dsp_buffer, sz);
+
+            int frames = CONVERT_SAMPLES;
+            for (ddb_dsp_context_t *dsp = dsp_chain; dsp; dsp = dsp->next) {
+                frames = dsp->plugin->process(dsp, dsp_buffer, frames, -1, &dspfmt, NULL);
+                if (frames <= 0) {
+                    *message = "DSP error, please check your preset";
```
Alexey-Yakovenko: Would make sense to localize this: wrap with _("")

plugins/converter/converter.c
```diff
@@ -223,1019 +168,1224 @@ encoder_preset_copy (ddb_encoder_preset_t *to, ddb_encoder_preset_t *from) {
...
+                *(uint16_t *)(wavehdr+44) = LE16(outfmt.is_float ? 3 : 1);
             }
-            else {
-                enc_pipe = popen (enc, "w");
-                if (!enc_pipe) {
-                    fprintf (stderr, "converter: failed to open encoder\n");
-                    goto error;
-                }
+            if (fwrite(wavehdr, wavehdr_size, 1, output) != 1) {
+                *message = "Wave header write error";
```
Alexey-Yakovenko: localization

plugins/converter/converter.c
```diff
@@ -50,171 +57,109 @@
 static ddb_converter_t plugin;
 static DB_functions_t *deadbeef;
 
-static ddb_encoder_preset_t *encoder_presets;
-static ddb_dsp_preset_t *dsp_presets;
+struct encoder_preset_link {
```
Alexey-Yakovenko: please rename into encoder_preset_link_s

plugins/converter/converter.c
```diff
@@ -50,171 +57,109 @@
...
-encoder_preset_alloc (void) {
-    ddb_encoder_preset_t *p = malloc (sizeof (ddb_encoder_preset_t));
-    if (!p) {
-        fprintf (stderr, "failed to alloc ddb_encoder_preset_t\n");
-        return NULL;
-    }
-    memset (p, 0, sizeof (ddb_encoder_preset_t));
-    return p;
-}
+struct dsp_preset_link {
```
Alexey-Yakovenko: please rename into dsp_preset_link_s

plugins/converter/converter.c
```diff
@@ -223,1019 +168,1224 @@ encoder_preset_copy (ddb_encoder_preset_t *to, ddb_encoder_preset_t *from) {
...
+    }
+    return preset_path(deadbeef->get_plugin_dir(), "convpresets", title_copy);
 }
 
-ddb_dsp_preset_t *
-dsp_preset_load (const char *fname) {
-    ddb_dsp_preset_t *p = dsp_preset_alloc ();
-    if (!p) {
+static struct encoder_preset_link *
+encoder_preset_create(const char *fname)
```
Alexey-Yakovenko: This looks more like "encoder_preset_load_from_file"

Alexey-Yakovenko:
I have reviewed about half of the changes so far.

I'll try to finish doing this initial review run today.

The next step to make this merge-able, would be to rearrange the code so that the diff is minimal, and I can actually see what has been changed in the implementations of the functions with the same name, but I will do this after you fix all the stuff from comments.

plugins/converter/converter.c
```diff
@@ -223,1019 +168,1224 @@ encoder_preset_copy (ddb_encoder_preset_t *to, ddb_encoder_preset_t *from) {
...
-    // last part of outfile is the filename
-    get_output_field (it, field, fname, sizeof(fname));
+    return root;
+}
 
-    l = strlen (out);
-    snprintf (out+l, sz-l, "%s.%s", fname, encoder_preset->ext);
-    trace ("converter output file is '%s'\n", out);
+static char *
+escape_filepath(const char *path, char *escaped_path)
```
Alexey-Yakovenko: You removed the output buffer size (used for buffer overflow check). Why?

Alexey-Yakovenko:
I made a new branch, where I fixed some of the things from my comments, and rearranged the code to make it somewhat diff-friendly. the branch is called "converter2" on my repo. Please use pull it before making further changes.

Alexey-Yakovenko:
Another thing: I couldn't figure out from code how do you handle gui events, since the callbacks.c is empty, and there's no obvious g_signal_connect calls. Can you point it out to me?

plugins/converter/converter.c
Lithopsian: How so?  The return value is a pointer to the allocated string.  It it is on the stack, the caller will get a corrupt buffer.

plugins/converter/converter.c
Lithopsian: The buffer is always allocated large enough to contain the string.  snprintf is unnecessary.

plugins/converter/converter.c
Lithopsian: The buffer is always large enough.

Lithopsian:
All GUI events are handled in convgui.c.  Plenty of signals attached in there :)  converter.c contains only the behind-the-scenes code for converting files, isolated behind a separate plugin API.

plugins/converter/converter.c
Lithopsian: check_dir is recursive.  Using alloca to allocate potentially allocate hundreds of kilobyte-scale buffers could blow the stack.  Probably not in practice, but entering a bizarre long and deep path would do it.

Alexey-Yakovenko: The path that you pass to check_dir is writable, it's a copy that you make in ensure_dir.
The check_dir can be modified to accept a non-const pointer, and use dirname on that. Then you don't need any allocations at all.

Alexey-Yakovenko: Further improvement of this function would be to replace recursion with a while loop.

Lithopsian:
Probably not helpful to review this as a diff.  As the stats show, there are more additions then total lines in converter.c.  convgui.c is even closer to completely rewritten.

Alexey-Yakovenko:
If you pull from my converter2 branch, and resubmit a pull request -- it will be easier to review as a diff.

Alexey-Yakovenko:
I submitted a PR with my changes to you. The changes are fixing the formatting, and reordering functions so that they are close to their original placement in the file (mostly converter.c). After everything is fixed and reviewed - we can reorder the functions again to be at new places, but refactoring needs to be done in a separate commit. This is needed because when bugs happen I'm using bisect to find the commit which introduces the bug, then I'm using diff to see what changed. If the file changes the code while moving it around -- it makes it super-difficult to find the cause of the bugs.

Alexey-Yakovenko:
The rules to remember when contributing:

If you're changing the behavior of the code:
Before commiting, run git diff, and make sure that the diff is as minimal as possible. No moving code around, no changing indentations of large blocks, etc.

If you need to change indentation, or move a block of code around -- do it in a separate commit, and (preferably) separate pull request. Don't squash refactoring commits with code change commits.

Alexey-Yakovenko:
Trying to review the convgui.c
It has lots of bugs and coding style issues.
One particular case which is puzzling me is this format string: ```"%3$02u:%02u:%02u"```

man 3 printf says:
```
         An optional field, consisting of a decimal digit string followed by a
         $, specifying the next argument to access.  If this field is not pro-
         vided, the argument following the last argument accessed will be
         used.  Arguments are numbered starting at 1.  If unaccessed arguments
         in the format string are interspersed with ones that are accessed the
         results will be indeterminate.
```

gcc gives a warning: ```missing $ operand number in format [-Wformat]```

It looks like you meant this: ```%3$02u:%1$02u:%2$02u```, but I'm not 100% sure.

plugins/converter/converter.c
Alexey-Yakovenko: Looks like this is relies on glibc behavior, and will have a limitation of 32bit offsets on all the other platforms. I'm removing this define, and will change the code to use correct APIs.

plugins/converter/converter.c
```diff
@@ -223,1019 +168,1224 @@ encoder_preset_copy (ddb_encoder_preset_t *to, ddb_encoder_preset_t *from) {
...
+        fprintf(stderr, "Converter will encode using internal RIFF WAVE writer\n");
+        FILE *wav = fopen(out, "w+b");
+        if (!wav) {
+            *message = "Cannot open output wave file";
+            trace("converter: failed to open output wave file %s\n", out);
+            goto error;
+        }
+        output = wav;
+    }
+    setbuf(output, NULL);
```
Alexey-Yakovenko: Needs explanation why the buffering is disabled here

plugins/converter/converter.c
```diff
@@ -223,1019 +168,1224 @@ encoder_preset_copy (ddb_encoder_preset_t *to, ddb_encoder_preset_t *from) {
...
+    }
 }
 
-int
-converter_cmd (int cmd, ...) {
+static int
+convert(DB_playItem_t *it, const ddb_encoder_preset_t *encoder_preset, const char *out, const ddb_dsp_preset_t *dsp_preset, const int output_bps, const int output_is_float,
+         enum ddb_convert_api *api, char **message, void (* convert_callback)(const time_t, const time_t, const float, void *), void *user_data)
+{
+#ifndef __STDC_IEC_559__
```
Alexey-Yakovenko: I'm removing this check, as deadbeef cannot correctly function on platforms which don't support IEEE 754 floats, no need to put junk checks in the code.

plugins/converter/converter.c
```diff
@@ -223,1019 +168,1224 @@ encoder_preset_copy (ddb_encoder_preset_t *to, ddb_encoder_preset_t *from) {
...
+        const size_t bytes = outfmt.bps / 8;
+        for (size_t i = 0; i < readsize; i += bytes) {
+            for (size_t j = 0; j < bytes; j++) {
+                swap_buffer[i+bytes-j-1] = swap_buffer[i+j];
             }
+        }
+#endif
+        const time_t now = time(NULL);
+        if (samples >= next_sample || now >= next_time) {
+            convert_callback(start_time, now, temp_file_path ? -15 : (float)samples/fileinfo->fmt.samplerate, user_data);
```
Alexey-Yakovenko: trying to figure out the names of callback parameters, found the magic number -15, is it used anywhere else? looks like this is length in seconds...

Lithopsian: Not seconds.  It is a hard-coded percent value.  There is no sensible progress for temporary files, so I just set fixed percentages before and after creating the temporary file.  An alternative might be an indeterminate progress bar, but I think they look dodgy.

Alexey-Yakovenko:
So, if it's negative it's percentage, but positive is length?

plugins/converter/converter.c
Alexey-Yakovenko: So, if it's negative it's percentage, but positive is length?

Alexey-Yakovenko:
I spend many hours (days in fact) trying to refactor this pull request into something mergeable.
Unfortunately, I only kept finding more problems in the new version of the code, and I can't afford enough time to fix it.

* It has a huge amount of changes, which both add new features, and change the existing features
* Many changes in the existing features are unacceptable, they break stuff to the level where it won't compile on some of the supported platforms, or won't work correctly.
* The existing code is reordered too much
* The new stuff often doesn't follow the project's coding conventions

I give up on this PR, and my recommendation is to make your own converter plugin which you can develop and maintain separately, until it becomes stable and well accepted.

Then we can talk about replacing the existing deadbeef's converter with the new one entirely, without quality regressions.

PR #70, opened 2015-01-20
flac: support non-byte-aligned bit depths

Lithopsian:


Alexey-Yakovenko:
did you profile after making the change? to me it looks like performance would be much worse than it was.

Alexey-Yakovenko:
I don't get, why it isn't enough to just tweak bps in the cflac_init?

Lithopsian:
Performance is faster than before :)  On my machine anyway, but probably in all cases.

Lithopsian:
You can't just set the bps and walk away.  Each sample is a certain number of bits sitting at the least significant end of an int32.  The correct bits must be extracted into the correct number of bytes (eg. 24 bit means copying three bytes out of the int32)  Just extracting the bytes will cause 20 bit sample volume to be 16 times too low, so they have to be bit-shifted.

Alexey-Yakovenko:
> Performance is faster than before :) On my machine anyway, but probably in all cases.

I don't see how that's possible, as your inner loop is doing 3 additional comparisons, 4 shifts and 4 &'s per iteration (for the worst case scenario), compared to my implementation which does the checks _before_ running the loop, and does 1 less shift in the best case scenario (8 bit).

I suppose compiler can optimise it, because you cache the real_bps value in an non-aliased variable, but I don't like relying on it, because it makes it not obvious how the code really works.

So what I'd at least keep the original code for the byte-aligned cases, because they can work faster (1 less shift per iteration), and add another out-of-loop "if" for handling the non-byte-aligned cases the way you did it.

Alexey-Yakovenko:
The other changes seem to be fine.

plugins/flac/flac.c
```diff
@@ -126,63 +126,41 @@ FLAC__bool flac_eof_cb (const FLAC__StreamDecoder *decoder, void *client_data) {
...
     }
-    else if (_info->fmt.bps == 8) {
-        for (int i = 0; i <  nsamples; i++) {
-            for (int c = 0; c < _info->fmt.channels; c++) {
-                int32_t sample = inputbuffer[c][i];
-                *bufptr++ = sample&0xff;
-                info->remaining += 1;
+
+    const unsigned real_bps = FLAC__stream_decoder_get_bits_per_sample(decoder);
+    const unsigned shift = (8 - real_bps % 8) % 8;
```
Alexey-Yakovenko: This calculation is really unclear. Did you mean just real_bps%8 here?

plugins/flac/flac.c
```diff
@@ -126,63 +126,41 @@ FLAC__bool flac_eof_cb (const FLAC__StreamDecoder *decoder, void *client_data) {
...
+                    }
+                }
             }
         }
     }
-    if (readbytes > bufsize) {
-        trace ("flac: buffer overflow, distortion will occur\n");
-    //    return FLAC__STREAM_DECODER_WRITE_STATUS_ABORT;
-    }
+    info->remaining = bufptr - info->buffer;
```
Alexey-Yakovenko: This is a clever improvement.

Alexey-Yakovenko:
I made a proper fix for this, using some of your code, and applying all fixes possible. Will test and push soon.

plugins/flac/flac.c
Lithopsian: I think (8 - real_bps % 8) % 8 is correct.  An example is that the bits to shift is just 1 for a 23 bit flac, not 7.  The final modulus is just so that byte-aligned bit depths have a shift of zero, not 8.

plugins/flac/flac.c
Lithopsian: The other thing that made a difference in my performance testing was the local variable for channels.  Not that any of it is critical, the whole thing is a few micro-seconds for 4096 samples.

plugins/flac/flac.c
Alexey-Yakovenko: I changed that to "unsigned shift = _info->fmt.bps - real_bps;"
Gives the same result, but much easier to read.

PR #72, opened 2015-02-03
pltbrowser: add more mouse support, make columns configurable, add duration column, hightlight current playlist

cboxdoerfer:
The patch in selproperties addresses a performance issue when the selection properties widget is visible and the DB_EV_SELCHANGED event is fired to often (e.g. when holding shift and cursor-up or down in playlist).

Playlistbrowser:
- I made the columns configurable and added an duration column showing the total duration of the playlist. Both items and duration column are disabled by default and can be enabled in a popup menu by right-clicking on the header.

- Highlighting the current playlist adds "(playing)" at the end of the active playlist. This is disabled by default.

- Middle mouse button on playlist closes playlist. (disabled by default)

- middle mouse button and double click with left mouse button on empty area opens new playlist, didn't feel the need to make that configurable but of course I'll do so if you like

Looking forward to suggestions.

### DIFF
--- a/plugins/gtkui/widgets.c
+++ b/plugins/gtkui/widgets.c
@@ -126,6 +126,7 @@ typedef struct {
 typedef struct {
     ddb_gtkui_widget_t base;
     GtkWidget *tree;
+    guint refresh_timeout;
 } w_selproperties_t;
 
 typedef struct {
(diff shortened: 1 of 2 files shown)


plugins/gtkui/widgets.c
```diff
@@ -2222,8 +2232,10 @@ selproperties_message (ddb_gtkui_widget_t *w, uint32_t id, uintptr_t ctx, uint32
 }
 
 static void
-w_selproperties_init (struct ddb_gtkui_widget_s *w) {
-    fill_selproperties_cb (w);
+w_selproperties_init (struct ddb_gtkui_widget_s *widget) {
+    w_selproperties_t *w = widget;
+    w->refresh_timeout = 0;
```
Alexey-Yakovenko: refresh_timeout cannot be non-zero when this is called, so this line is not needed.

PR #75, opened 2015-02-25
pltbrowser: add inline editing and F2 shortcut to edit playlist name

cboxdoerfer:


cboxdoerfer:
Build on drone.io fails with this patch. I'll reopen when I have solved that.

Alexey-Yakovenko:
You can simply update the pull request.

cboxdoerfer:
yeah, but at that time I only saw "build failed" and just wanted to be sure you don't pull some mature breakage. Turned out there was just some incompatiblity with the gtk2 version in the static deps.

Alexey-Yakovenko:
I remember, that "event-after" handler was made on purpose, for correctly handling the context menu activation on right click (the case where you hold the mouse button, select the menu item, and then release the button), and the change you made wouldn't work correctly, or maybe break something else. Why did you change this?

Alexey-Yakovenko:
Yeah, I should've written a comment about why it was done this way. But try to play with context menu, and you'll see for yourself.

cboxdoerfer:
Oh, I just saw that this event was fired quite often and thought button-press-event is enough. Of course I"ll change it back.

PR #76, opened 2015-03-02
pltbrowser: avoid flickering on playlistchange, some minor refactoring and fix ddbtabstrip not being updated on column sort

cboxdoerfer:


cboxdoerfer:
Found a bug with this pull request and I'm not sure if this can be fixed nicely.

cboxdoerfer:
i.e. I don't want the treeview to be refilled on playlistswitch because this leads to a short flicker, so I only update the cursor, but since we also use the DB_EV_PLAYLISTSWITCHTED signal when the sort order changed, I can't simply distinguish between a playlist sort change (which needs a refill) and a playlist switch (which doesn't).

cboxdoerfer:
Probably the most elegant solution to that would be an additional signal, e.g. DB_EV_PLAYLISTORDERCHANGED. But I'm not sure if that's worth it, what do you think about that?

Alexey-Yakovenko:
any idea why DB_EV_PLAYLISTSWITCHTED is sent on sort order change?

cboxdoerfer:
because its meaning matches a sort change the closest from the available signals.

cboxdoerfer:
deadbeef.h says:
DB_EV_PLAYLISTCHANGED = 15, // playlist contents were changed

But the contents weren't really touched on a sort change

cboxdoerfer:
Oh, and by playlist sort I don't mean the sort of the playlist content, but the sort order of the playlist, how they are displayed in the tabstrip and pltbrowser

Alexey-Yakovenko:
Ah yes, for this we'd need to add a new event, and I was missing it too on few occasions.

cboxdoerfer:
Hm, somehow your commits show up here too. Any idea why?

cboxdoerfer:
I guess I close that one and open a new one.

PR #1216, opened 2015-05-12
gtkui: use column alignment also on column headers

cboxdoerfer:


Alexey-Yakovenko:
Rejected based on IRC discussion.

PR #1222, opened 2015-05-24
ddbtabstrip: add min/max tab size as #defines in header (no more change to ddbtabstrip.c req'd)

andreas--e:
Hi Alexey

hey this is my first PR :) Moved hardcoded values of min_tab_size / max_tab_size out of `ddbtabstrip.c` and set #defines in the header file accordingly.
Note this is only an example. Maybe it would make sense to also have margin, v-align etc. values as #defines - but I've decided to stick to these two most important (and most prone-for-tweaking) ones.


Alexey-Yakovenko:
Please collapse all 5 commits into 1, and write a meaningful commit message. Thanks.

andreas--e:
can you come to irc for awhile again? I am totally new to this pr stuff here, and most of the buttons are Greek to me.

PR #1240, opened 2015-08-08
vorbis: respect --disable-vorbis

novel:
Do not try to detect libogg when '--disable-vorbis' flag was
specified.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -399,18 +399,20 @@ AS_IF([test "${enable_vorbis}" != "no"], [
 ])
 
 dnl libogg for oggedit
-AS_IF([test "${enable_staticlink}" != "no"], [
-    HAVE_OGG=yes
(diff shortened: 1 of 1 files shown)


Alexey-Yakovenko:
Hi

I see what the problem is and what you tried to fix. Unfortunately it's not correct fix, and would break many other things. I fixed it the way I see it. See commit f58d6e38ac9725fd1fe7760232a36bcefa940ee4. Please let me know if it works for you.

Alexey-Yakovenko:
Explanation in case you wondering...

libogg is used for several other things, not just vorbis plugin. so it's always detected, even if the vorbis plugin is disabled. the bug was there that vorbis plugin was enabled during libogg detection, which was wrong.

novel:
That sounds perfectly reasonable, thanks.

PR #1242, opened 2015-08-10
Support for original release date

ugilio:
Hello,

sometimes I want to be able to use the "original release date" in deadbeef instead of the release date. Example case: I rip a CD that is a remastered edition from 2010 of an album that was released 30 years ago, then I use MusicBrainz Picard to populate the tags, leading to date="2010-XX-YY" and originaldate="1985-something". In the deadbeef playlist and artwork view I want to see the original date and not the remastered date.

The first commit adds more mapping to ORIGINAL_RELEASE_TIME besides the one already existing from ID3 and APE (ffmpeg,flac,vorbis,asf), and also adds ORIGINAL_RELEASE_YEAR. For the mapping I followed what reported by this Picard page:
https://picard.musicbrainz.org/docs/mappings/

The second commit is quite useless for master branch, but it can be applied to the 0.6.2-fixes branch (if the first one is applied before) and adds a %Y replacement string: %Y uses ORIGINAL_RELEASE_TIME, falling back to ORIGINAL_RELEASE_YEAR and then to YEAR.


Alexey-Yakovenko:
Thanks for the code, @ugilio 
Very clean patch.

PR #1243, opened 2015-08-12
KDE5 Plasma system tray support

ugilio:
Hello,

This branch introduces support for KDE5 Plasma system tray, while gracefully falling back to  GtkStatusIcon on systems that don't support it.

KDE5 has dropped support for xembed, so on my system (Fedora 22, KDE5) I can no longer use Deadbeef's system tray icon without awkward workarounds (e.g. wmsystemtray). See also
http://blog.martin-graesslin.com/blog/2014/06/where-are-my-systray-icons/

This implementation makes use of libstatusnotifier library: https://github.com/jjk-jacky/statusnotifier/. Initially I tried to use libappindicator (see also my appindicator branch: https://github.com/ugilio/deadbeef/tree/appindicator), but the big problem is that by design it does not support reacting to mouse clicks (only the popup menu is supported), nor scrolling of mouse wheel on "application type" icons. Libstatusnotifier does not have these limitations.

Libstatusnotifier support can be enabled by --enable-statusnotifier and --enable-statusnotifier=embedded. In the first case it uses the system-provided libstatusnotifier. In the second case it compiles the library inside plugins/ddb_gui_GTK*.so. The embedding is in my opinion better because
* the library is not commonly installed by popular distributions
* by default, the library links to gdk3, so it might be a problem for gtk2-only users willing to compile it manually
* the library does not need to link to new libraries besides the GTK-provided ones, so it does not add new dependencies on ddb_gui_GTK*.so that would prevent the GTK ui from running on certain machines

Possible problem: libstatusnotifier is GPLv3, so this forces GTK ui plugin to be released under GTKv3 too?

With this branch I can use all features of the old tray icon (menu, left click, middle click, scrolling). It might need a little bit of testing on xembed-only systems (I only tried forcing failure of initialization by code, and it correctly falls back to GtkStatusIcon).


Alexey-Yakovenko:
I'm sorry, but I cannot accept this, as it would bump entire project's license to GPL3 (at least on linux/gtk), which I don't want. Please consider shipping the code under GPL3 license as a separate plugin.

Alexey-Yakovenko:
Another possibility would be to negotiate with the author of the statusnotifier lib to relicense it under a different license, preferrably LGPL instead of GPL.

Alexey-Yakovenko:
The response was "not gonna happen" with this link: https://github.com/jjk-jacky/statusnotifier/issues/1

I guess you'll need to convert your patch into a plugin.
Please let me know if you need anything to be changed in deadbeef itself, to make it easier, or any other help.

ugilio:
Okay... a plugin might be done, the drawbacks is that something that should just work out of the box does not (e.g. the user must manually install an extra plugin)

The other option would be to reimplement the thing from scratch. It should not be too hard, i think it's just a matter of sending and receiving some dbus messages. And we don't need a fully fledged library, only what's needed by deadbeef. Now I am a little short of time and don't know anything about dbus (but hey, I didn't know anything about Gtk too!), so I need some time to see what I can do (if the first or the second option).

Alexey-Yakovenko:
There can be another problem with the statusnotifier lib, which I didn't test, but it's quite possible that it won't compile or work with our lowest supported GNU/linux version. When we modify GTKUI, we need to be sure that it still runs on GTK2.12+, and works on ubuntu 8.04, etc. However, I think that in deadbeef 0.7 I'll update the requirements, and move on to GTK 2.16 minimum. That's still quite old, and we need to be careful with the glib and dbus APIs which appeared later.

Alexey-Yakovenko:
@ugilio 
Instead of writing the new library from scratch, how about forking the libappindicator, and adding the missing features?

ugilio:
Yesterday I already started to try to do a new implementation, because I feared libappindicator was something too big, and wanted to try something simpler. I wanted to try to see if it was feasible before answering, sorry for the delay.

I implemented a StatusNotifierItem support library, based on libgio. These APIs are in glib>=2.26, which means GTK 2.20-2.22 I think: for sure ubuntu 10.04, to name one, does not have it.

So indeed I went to the plugin way: machines having an old version of gtk libs for sure don't need the new KDE status notifier support.

The code in the statusnotifier branch works like this (do I have to open another pull request or this is enough?):
* Some systray-related functions are exposed in struct statusicon_functions_t (in gtkui.h, is this the right place?). In gtkui.c, some code is refactored so that the old GtkStatusIcon functions are grouped together in a similar struct. The gtk ui has a pointer to the current struct (by default the gtk one): by changing the pointer you change the systray icon implementation.
* The GTK ui tries to set up the status notifier plugin (if found) during connect and gives it a pointer to the struct. The status notifier plugin sets the pointer to point to its own functions. Nothing is allocated here, only the pointer is set.
* Everything is lazily created on the first call to gtkui_update_status_icon. If for some reason the status notifier setup fails (e.g, we are not running a desktop environment that supports it), the code restores the original pointer to the gtk functions and sets up a GtkStatusIcon.

For this, to work, I needed to export some functions from gtk ui:
`void (*mainwin_toggle_visible)(void);`
`void (*trayicon_do_scroll) (int amount);`
`GtkWidget *(*get_traymenu)(void);`

In theory these fallback mechanisms are enough so that deadbeef can always be compiled with --enable-statusnotifier, but maybe some more testing is needed.
Also, I left some code commented out in plugins/statusnotifier/Makefile.am for the static linking case, because gtk2 2.12 is not enough (we need glib 2.26 for it to work).

Alexey-Yakovenko:
Please open a new pull request. I need to do the code review.

Alexey-Yakovenko:
> Also, I left some code commented out in plugins/statusnotifier/Makefile.am for the static linking case

It should be possible to have newer GTK in the statically-linked version. I will need to verify this too.

PR #1250, opened 2015-08-15
Fix for uninitialized GError

ugilio:
Hopefully, this should fix the problem.
My guess is that the compiler on my machine sets that variable to zero, while the compiler on drone.io did not (it may also be due to different optimization levels).
This should fix the issue.

### DIFF
--- a/plugins/statusnotifier/ddb_statusnotifier.c
+++ b/plugins/statusnotifier/ddb_statusnotifier.c
@@ -123,7 +123,7 @@ static void notifier_initialize_status_icon(char * iconstr) {
 }
 
 static void notifier_create_status_icon_from_file(char * iconfile) {
-    GError *err;
+    GError *err = NULL;
     GdkPixbuf *pixbuf = gdk_pixbuf_new_from_file (iconfile, &err);
     notifier = sn_create_with_icondata("deadbeef-notifier",ApplicationStatus,pixbuf);
     sn_set_tooltip_icondata(notifier,pixbuf);


Alexey-Yakovenko:
Merged. Please make sure to add "statusnotifier: " prefix to all commits related to this plugin, thanks.

ugilio:
Ok, I downloaded the drone.io built package for x86_64 and I can confirm it works as expected.
For the statusnotifier prefix, should I amend previous commit messages? In my statusnotifier branch? Doesn't this cause problems with merged commits?

Alexey-Yakovenko:
No need to amend, I'll fix them.

Alexey-Yakovenko:
Me and two other people are testing this plugin now.

I can confirm the old tray icon still works with lxpanel - which is good.

The new tray icon works in kde 5.3.2 -- also good.

It doesn't work in Unity on ubuntu 12.04, but I don't think it ever worked, and I don't think it should be present (dock replaces the system tray in Unity, similar to OSX).

I'm still waiting for the guys to provide the full test results.

Meanwhile, I need to go through entire diff, and make sure everything is good.

Alexey-Yakovenko:
One guy reports that the system tray icon is not present in XFCE, while it works fine in the build from master branch.

Alexey-Yakovenko:
I reworked the statusnotifier code, to elimitate the cyclic dependency between GTKUI and the statusnotifier plugins.
I have also added an option to disable the new status icon via the plugin settings.
But now there's the problem: the statusnotifier icon is not hidden, when notifier_set_status_icon_visible (notifier, FALSE) is called. Looks like the "Passive" is not the right constanst to send, to hide the icon. I tried destroying the notifier, but that's even worse because it crashes on next use.

Alexey-Yakovenko:
I pushed the version which is using sn_destroy to hide the icon. It is broken, and crashes when attempted to call ```notifier_create_status_icon_from_icon_name```.

Another thing that needs to be figured out is when to call ```notifier_create_status_icon_from_icon_name``` vs ```notifier_create_status_icon_from_icon_name```

ugilio:
Ok, I'm going to look at it.

Alexey-Yakovenko:
@ugilio we should find a better communication channel. Your email address in the commits is a fake one, so I can't use it. Is there a better way of contacting you?

PR #1253, opened 2015-08-17
statusnotifier: fix crash when re-enabling plugin

ugilio:
This fixes the crash when the plugin in disabled and then re-enabled.


Alexey-Yakovenko:
works, merged.

PR #1263, opened 2015-09-08
Add a couple new tf fields

cboxdoerfer:


tf.c
```diff
@@ -788,6 +801,93 @@ tf_eval_int (ddb_tf_context_t *ctx, char *code, int size, char *out, int outlen,
...
                 }
+                // index of track in playlist (zero-padded)
+                else if (!strcmp (name, "list_index")) {
+                    if (it) {
+                        playlist_t *plt = pl_get_playlist (it);
+                        if (plt) {
+                            int total_tracks = plt_get_item_count (plt, ctx->iter);
+                            int digits = log10 (abs (total_tracks)) + 1;
+                            int idx = pl_get_idx_of_iter (it, ctx->iter) + 1;
+                            int len = snprintf (out, outlen, "%0*d", digits, idx);
```
Alexey-Yakovenko: Foobar2000 %list_index% doesn't do padding, even though the title formatting reference says so. Please change to ```... "%d", idx ...```

tf.c
```diff
@@ -788,6 +801,93 @@ tf_eval_int (ddb_tf_context_t *ctx, char *code, int size, char *out, int outlen,
...
                     val = pl_find_meta_raw (it, ":URI");
                 }
+                // index of track in playlist (zero-padded)
+                else if (!strcmp (name, "list_index")) {
+                    if (it) {
+                        playlist_t *plt = pl_get_playlist (it);
+                        if (plt) {
+                            int total_tracks = plt_get_item_count (plt, ctx->iter);
+                            int digits = log10 (abs (total_tracks)) + 1;
+                            int idx = pl_get_idx_of_iter (it, ctx->iter) + 1;
```
Alexey-Yakovenko: This pl_get_idx_of_iter is only needed if ddb_tf_context_t::idx is -1, because this operation is very slow.
The caller is responsible to set the idx to correct value, to make this fast.

tf.c
```diff
@@ -788,6 +801,93 @@ tf_eval_int (ddb_tf_context_t *ctx, char *code, int size, char *out, int outlen,
...
+                    }
+                }
+                // total number of tracks in playlist
+                else if (!strcmp (name, "list_total")) {
+                    int total_tracks = -1;
+                    if (ctx->plt) {
+                        total_tracks = plt_get_item_count ((playlist_t *)ctx->plt, ctx->iter);
+                    }
+                    else if (it) {
+                        playlist_t *plt = pl_get_playlist (it);
```
Alexey-Yakovenko: The foobar2000 ref is not very accurate about this one, but I think we should just use the current playlist here, instead of trying to guess which playlist the track belongs to.

tf.c
```diff
@@ -788,6 +801,93 @@ tf_eval_int (ddb_tf_context_t *ctx, char *code, int size, char *out, int outlen,
                 else if (!strcmp (name, "path")) {
                     val = pl_find_meta_raw (it, ":URI");
                 }
+                // index of track in playlist (zero-padded)
+                else if (!strcmp (name, "list_index")) {
+                    if (it) {
+                        playlist_t *plt = pl_get_playlist (it);
+                        if (plt) {
+                            int total_tracks = plt_get_item_count (plt, ctx->iter);
+                            int digits = log10 (abs (total_tracks)) + 1;
```
Alexey-Yakovenko: log10(0) returns -inf
needs to be handled as special case

Alexey-Yakovenko: something like this should work better

```
int digits = 0;
do {
    x /= 10;
    digits++;
} while (x);
```

PR #1268, opened 2015-09-12
pltbrowser: fix memleak

cboxdoerfer:
Diff looks pretty complex, but this patch just added an if check and missing plt_unref:

```
if (plt) {
    ...
    deadbeef->plt_unref (plt);
}
```


Alexey-Yakovenko:
Is it possible to do something like if (!plt) continue?

cboxdoerfer:
yes sure, I'll update the pull request right away. I also found two other bad big memleaks.

cboxdoerfer:
updated pull request.

PR #1271, opened 2015-09-14
Tab widget: Add placeholder tab when last tab gets deleted, and ignore button press events from child widgets

cboxdoerfer:


plugins/gtkui/widgets.c
```diff
@@ -1757,11 +1791,11 @@ on_tabs_button_press_event (GtkWidget      *notebook,
...
     gboolean       close_tab;
 
-    /* get real window coordinates */
-    int x, y;
-    gdk_window_get_position (event->window, &x, &y);
-    x += event->x;
-    y += event->y;
+    int event_x, event_y;
+    if (!get_event_coordinates_in_widget (notebook, event, &event_x, &event_y)) {
+        // clicked outside the tabstrip (e.g. in one of its childs)
```
Alexey-Yakovenko: should be "children" instead of "childs" :)

Issue #1200, opened 2015-03-24
Non-square covers not properly displayed (bad quality).


Soukyuu:
There seems to be a problem with covers that are not square, the resize result looks as if it was resized with nearest neighbor. Even at (close to) 100%, the image quality is really bad.

Here is an [example cover](https://cloud.githubusercontent.com/assets/2095054/6804477/2ac10568-d23e-11e4-9a16-d3965bb5bf43.jpg), and here is how it [looks in ddb](https://cloud.githubusercontent.com/assets/2095054/6804579/bd90dcec-d23e-11e4-8a96-cc5061c227eb.png) 
(Note: the displacement to the right is due to screenshot cutting).

If I match the panel aspect ratio to cover aspect ratio, the cover looks correct, but the square ones do not. It seems that the album art panel uses the aspect ratio or the panel for the resize, not the aspect ratio of the image itself.

deadbeef-git, arch x64

Alexey-Yakovenko:
Are you talking about cover art widget, or the playlist (album art column), or OSD notification, or all of them?

Soukyuu:
The widget, but now that I looked at the column, it does the same thing as well.
The OSD is affected as well, it doesn't even keep the aspect ratio (rectangular -> stretched to square). I'm on KDE4 if it matters.

Alexey-Yakovenko:
most likely it's kde4's fault for not maintaining the aspect ratio in OSD.
everything else is somewhere in deadbeef code for sure...

Alexey-Yakovenko:
While you're at it -- could you verify if the same thing happens with last stable release 0.6.2?
(you will need to delete the ~/.cache/deadbeef folder before running it)

Soukyuu:
Yes, 0.6.2-3 from the official repos shows the same behavior.

Soukyuu:
The files themselves are correctly extracted to the ./cache/deadbeef/covers folder by the way.

Lithopsian:
The resizing may be done by Deadbeef, or by an external library, depending on the build configuration.  If imlib is present, then it will be used for resizing, and quality should be decent.  Without imlib, resizing is done by Deadbeef code and it used to be poor.  I added code to do it better, not sure if it is in 0.6.2 but should be in trunk.

The OSD thumbnail size can be set in the preferences.  If the system OSD icon size is different then the OSD library will resize again, and that is probably causing your troubles.  Might be an enhancement to have Deadbeef always a supply a square icon?

[...]

PR #1278, opened 2015-09-27
Fix widget artwork scaling for non-square images

Lithopsian:
See bug #1200.  This pull request fixes the large widget scaling.  It allows requesting cover art by both height and width, to avoid the problem of not knowing what scaling to use without knowing the image dimensions and not knowing the image dimensions until the file has been loaded.  There is no change to the external API, although maybe the new call could be exposed.

### DIFF
--- a/plugins/gtkui/coverart.c
+++ b/plugins/gtkui/coverart.c
@@ -252,7 +252,15 @@ cache_qsort(const void *a, const void *b)
     const cached_pixbuf_t *y = (cached_pixbuf_t *)b;
     if (x->pixbuf && y->pixbuf) {
         const int cmp = strcmp(x->fname, y->fname);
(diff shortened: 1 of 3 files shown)


plugins/gtkui/coverart.c
```diff
@@ -248,7 +252,7 @@ cache_qsort(const void *a, const void *b)
     const cached_pixbuf_t *y = (cached_pixbuf_t *)b;
     if (x->pixbuf && y->pixbuf) {
         const int cmp = strcmp(x->fname, y->fname);
-        return cmp ? cmp : y->width - x->width;
+        return cmp ? cmp : y->width != x->width ? y->width - x->width : y->height - x->height;
```
Alexey-Yakovenko: please put some parentheses in there to reduce confusion.

Lithopsian:
Parentheses seemed to look even muddier, so this should be clearer.

PR #1292, opened 2015-10-18
emit compile warning

proudzhu:
"cc1: warning: command line option ‘-fno-rtti’ is valid for C++/ObjC++
but not for C"


Alexey-Yakovenko:
I think this change will break freebsd builds. But let's try and see what happens.

Issue #1301, opened 2015-10-22
Pinned titlebar does not redraw on vertical expose in GTK2


Lithopsian:
Can someone confirm this for me, to make sure I'm not hallucinating?  I'm testing the redraw clip changes I've made and saw something odd, but it seems to have been a longstanding problem.  The fix is simple.  I just wanted to be sure other people see the same problem in case I'm missing something.

Set up a playlist so that a group is pinned, that is the group is partially scrolled but the titlebar is pushed down and remains visible (right click and make sure that the pin groups checkbox is selected).  Now wipe another window vertically over that titlebar so that it is revealed by the top or bottom of the other window.

In GTK2 the titlebar is no longer visible, and the underlying row is drawn.  Expose by moving the window horizontally over the titlebar and it is successfully redrawn.  Scrolls and switching windows will redraw the whole playlist and everything will be fine again.  In GTK3 it works fine, because the GTK3 code on trunk does not pass a clip region to the render function (actually it does, but always the whole draw area), but I have added this and that's when I spotted the problem.

cboxdoerfer:
Can't reproduce it. By "window" do you mean an actual window of another application or a widget in the deadbeef layout?

Alexey-Yakovenko:
I can't reproduce either. But this problem can be WM-dependent, because cmpositing WMs can cache the window contents.

Alexey-Yakovenko:
Yes, the bug is valid in openbox, reproduced by following the steps described above, just drag any window over deadbeef window to force group title be redrawn.

Lithopsian:
I'm using Openbox with no compositing.  You can also reproduce it, probably on any WM, by carefully aligning widgets so that a menu popup exposes the pinned titlebar in the right way.  The problem occurs when the top or bottom edge of the expose region lies within the pinned titlebar.  Anyway, the fix is simple so I'll add it.

Lithopsian:
This is fixed in pull request #1303.

PR #1303, opened 2015-10-23
Playlist artwork and rendering changes

Lithopsian:
Fix playlist artwork sizing (issue #1200).
Fix memory leak, GTK assertions, and possible crash when removing coverart widget during a load.
Support multiple album art columns with different widths.
Fix column type in edit dialog (tf changes).
Fix column minheight calculations to allow for padding (and potentially more complex calculations).
Fix a number of deprecation warnings (playqueue, GTK3 gtk_paint_*, gdk_window_get_pointer, gdk_cairo_set_source_color).
Add clip region to GTK3 draw event calls and make more use of the clip region.
Fix native themed column headers.
Fix pinned header exposes (issue #1301)

### DIFF
--- a/plugins/gtkui/coverart.c
+++ b/plugins/gtkui/coverart.c
@@ -61,7 +61,7 @@ static GdkPixbuf *pixbuf_default;
 static size_t thrash_count;
 
 typedef struct cover_callback_s {
-    cover_avail_callback cb;
+    cover_avail_callback_t cb;
     void *ud;
     struct cover_callback_s *next;
 } cover_callback_t;
@@ -87,7 +87,7 @@ typedef struct {
     char *cache_path;
     int width;
     int height;
-    cover_avail_callback callback;
+    cover_avail_callback_t callback;
     void *user_data;
 } cover_avail_info_t;
 
@@ -139,7 +139,7 @@ gtkui_is_default_pixbuf (GdkPixbuf *pb) {
 }
 
 static cover_callback_t *
-add_callback(cover_avail_callback cb, void *ud)
+add_callback(cover_avail_callback_t cb, void *ud)
 {
     if (!cb) {
         return NULL;
@@ -171,7 +171,7 @@ process_query_callbacks(cover_callback_t *callback, const int send)
 }
 
 static void
-queue_add (cache_type_t cache_type, char *fname, const int width, const int height, cover_avail_callback cb, void *ud)
+queue_add (cache_type_t cache_type, char *fname, const int width, const int height, cover_avail_callback_t cb, void *ud)
 {
     trace("coverart: queue_add %s @ %ix%i pixels\n", fname, width, height);
     load_query_t *q = malloc(sizeof(load_query_t));
@@ -198,7 +198,7 @@ queue_add (cache_type_t cache_type, char *fname, const int width, const int heig
 }
 
 static void
-queue_add_load (cache_type_t cache_type, char *fname, const int width, const int height, cover_avail_callback cb, void *ud)
+queue_add_load (cache_type_t cache_type, char *fname, const int width, const int height, cover_avail_callback_t cb, void *ud)
 {
     for (load_query_t *q = queue; q; q = q->next) {
         if (q->fname && !strcmp (q->fname, fname) && width == q->width && height == q->height) {
@@ -446,7 +446,7 @@ get_pixbuf (cache_type_t cache_type, const char *fname, const int width, const i
 }
 
 void
-queue_cover_callback (cover_avail_callback callback, void *user_data) {
+queue_cover_callback (cover_avail_callback_t callback, void *user_data) {
     if (artwork_plugin && callback) {
         deadbeef->mutex_lock (mutex);
         queue_add(-1, NULL, -1, -1, callback, user_data);
@@ -512,7 +512,7 @@ best_cached_pixbuf(cache_type_t cache_type, const char *path)
 }
 
 static cover_avail_info_t *
-cover_avail_info(cache_type_t cache_type, char *cache_path, const int width, const int height, cover_avail_callback callback, void *user_data)
+cover_avail_info(cache_type_t cache_type, char *cache_path, const int width, const int height, cover_avail_callback_t callback, void *user_data)
 {
     if (cache_path) {
         cover_avail_info_t *dt = malloc(sizeof(cover_avail_info_t));
@@ -534,7 +534,7 @@ cover_avail_info(cache_type_t cache_type, char *cache_path, const int width, con
 }
 
 static GdkPixbuf *
-get_cover_art_int(cache_type_t cache_type, const char *fname, const char *artist, const char *album, int width, int height, cover_avail_callback callback, void *user_data)
+get_cover_art_int(cache_type_t cache_type, const char *fname, const char *artist, const char *album, int width, int height, cover_avail_callback_t callback, void *user_data)
 {
     if (!artwork_plugin) {
         return NULL;
@@ -581,31 +581,31 @@ get_cover_art_int(cache_type_t cache_type, const char *fname, const char *artist
 
 // Deprecated
 GdkPixbuf *
-get_cover_art_callb (const char *fname, const char *artist, const char *album, int width, cover_avail_callback callback, void *user_data)
+get_cover_art_callb (const char *fname, const char *artist, const char *album, int width, cover_avail_callback_t callback, void *user_data)
 {
     return get_cover_art_int(CACHE_TYPE_THUMB, fname, artist, album, width, -1, callback, user_data);
 }
 
 GdkPixbuf *
-get_cover_art_primary (const char *fname, const char *artist, const char *album, int width, cover_avail_callback callback, void *user_data)
+get_cover_art_primary (const char *fname, const char *artist, const char *album, int width, cover_avail_callback_t callback, void *user_data)
 {
     return get_cover_art_int(CACHE_TYPE_PRIMARY, fname, artist, album, width, -1, callback, user_data);
 }
 
 GdkPixbuf *
-get_cover_art_primary_by_size (const char *fname, const char *artist, const char *album, int width, int height, cover_avail_callback callback, void *user_data)
+get_cover_art_primary_by_size (const char *fname, const char *artist, const char *album, int width, int height, cover_avail_callback_t callback, void *user_data)
 {
     return get_cover_art_int(CACHE_TYPE_PRIMARY, fname, artist, album, width, height, callback, user_data);
 }
 
 GdkPixbuf *
-get_cover_art_thumb (const char *fname, const char *artist, const char *album, int width, cover_avail_callback callback, void *user_data)
+get_cover_art_thumb (const char *fname, const char *artist, const char *album, int width, cover_avail_callback_t callback, void *user_data)
 {
     return get_cover_art_int(CACHE_TYPE_THUMB, fname, artist, album, width, -1, callback, user_data);
 }
 
 GdkPixbuf *
-get_cover_art_thumb_by_size (const char *fname, const char *artist, const char *album, int width, int height, cover_avail_callback callback, void *user_data)
+get_cover_art_thumb_by_size (const char *fname, const char *artist, const char *album, int width, int height, cover_avail_callback_t callback, void *user_data)
 {
     return get_cover_art_int(CACHE_TYPE_THUMB, fname, artist, album, width, height, callback, user_data);
 }
--- a/plugins/gtkui/coverart.c
+++ b/plugins/gtkui/coverart.c
@@ -307,7 +307,7 @@ adjust_cache(struct timeval *oldest)
     thrash_count = timeval_older(&now, oldest) ? thrash_count+1 : 0;
 
     /* Grab more space more quickly at small cache sizes */
-    if (1<<thrash_count > thumb_cache_size) {
+    if (thrash_count*2 >= thumb_cache_size) {
         cached_pixbuf_t *new_thumb_cache = realloc(thumb_cache, sizeof(cached_pixbuf_t) * thumb_cache_size * 2);
         if (new_thumb_cache) {
             memset(&new_thumb_cache[thumb_cache_size], '\0', sizeof(cached_pixbuf_t) * thumb_cache_size);
--- a/plugins/gtkui/coverart.c
+++ b/plugins/gtkui/coverart.c
@@ -32,6 +32,7 @@
 #endif
 #include "../artwork/artwork.h"
 #include "gtkui.h"
+#include "coverart.h"
 
 //#define trace(...) { fprintf(stderr, __VA_ARGS__); }
 #define trace(...)
@@ -60,7 +61,7 @@ static GdkPixbuf *pixbuf_default;
 static size_t thrash_count;
 
 typedef struct cover_callback_s {
-    void (*cb)(void *ud);
+    cover_avail_callback cb;
     void *ud;
     struct cover_callback_s *next;
 } cover_callback_t;
@@ -86,7 +87,7 @@ typedef struct {
     char *cache_path;
     int width;
     int height;
-    void (*callback)(void *user_data);
+    cover_avail_callback callback;
     void *user_data;
 } cover_avail_info_t;
 
@@ -138,7 +139,7 @@ gtkui_is_default_pixbuf (GdkPixbuf *pb) {
 }
 
 static cover_callback_t *
-add_callback(void (*cb)(void *), void *ud)
+add_callback(cover_avail_callback cb, void *ud)
 {
     if (!cb) {
         return NULL;
@@ -157,7 +158,20 @@ add_callback(void (*cb)(void *), void *ud)
 }
 
 static void
-queue_add (cache_type_t cache_type, char *fname, const int width, const int height, void (*cb)(void *), void *ud)
+process_query_callbacks(cover_callback_t *callback, const int send)
+{
+    if (callback) {
+        if (send) {
+            trace("coverart: make callback to %p (next=%p)\n", callback->cb, callback->next);
+            callback->cb(callback->ud);
+        }
+        process_query_callbacks(callback->next, send);
+        free(callback);
+    }
+}
+
+static void
+queue_add (cache_type_t cache_type, char *fname, const int width, const int height, cover_avail_callback cb, void *ud)
 {
     trace("coverart: queue_add %s @ %ix%i pixels\n", fname, width, height);
     load_query_t *q = malloc(sizeof(load_query_t));
@@ -184,7 +198,7 @@ queue_add (cache_type_t cache_type, char *fname, const int width, const int heig
 }
 
 static void
-queue_add_load (cache_type_t cache_type, char *fname, const int width, const int height, void (*cb)(void *), void *ud)
+queue_add_load (cache_type_t cache_type, char *fname, const int width, const int height, cover_avail_callback cb, void *ud)
 {
     for (load_query_t *q = queue; q; q = q->next) {
         if (q->fname && !strcmp (q->fname, fname) && width == q->width && height == q->height) {
@@ -205,6 +219,7 @@ queue_add_load (cache_type_t cache_type, char *fname, const int width, const int
 static load_query_t *
 queue_remove(load_query_t *q)
 {
+    process_query_callbacks(q->callback, FALSE);
     load_query_t *next = q->next;
     if (q->fname) {
         free(q->fname);
@@ -219,17 +234,6 @@ queue_pop (void)
     queue = queue_remove(queue);
     if (!queue) {
         tail = NULL;
-    }
-}
-
-static void
-send_query_callbacks(cover_callback_t *callback)
-{
-    if (callback) {
-        trace("coverart: make callback to %p (next=%p)\n", callback->cb, callback->next);
-        callback->cb(callback->ud);
-        send_query_callbacks(callback->next);
-        free(callback);
     }
 }
 
@@ -299,11 +303,11 @@ adjust_cache(struct timeval *oldest)
     /* A possible thrash is when the oldest pixbuf is still fresh */
     struct timeval now;
     gettimeofday(&now, NULL);
-    now.tv_sec -= 10 + thumb_cache_size / 10;
+    now.tv_sec -= 2 + thumb_cache_size / 10;
     thrash_count = timeval_older(&now, oldest) ? thrash_count+1 : 0;
 
-    /* If the whole cache has been thrashed through then we need more space */
-    if (thrash_count > thumb_cache_size) {
+    /* Grab more space more quickly at small cache sizes */
+    if (1<<thrash_count > thumb_cache_size) {
         cached_pixbuf_t *new_thumb_cache = realloc(thumb_cache, sizeof(cached_pixbuf_t) * thumb_cache_size * 2);
         if (new_thumb_cache) {
             memset(&new_thumb_cache[thumb_cache_size], '\0', sizeof(cached_pixbuf_t) * thumb_cache_size);
@@ -405,7 +409,8 @@ loading_thread (void *none) {
                 load_image(queue);
             }
             if (artwork_plugin) {
-                send_query_callbacks(queue->callback);
+                process_query_callbacks(queue->callback, TRUE);
+                queue->callback = NULL;
             }
             queue_pop();
         }
@@ -428,8 +433,12 @@ get_pixbuf (cache_type_t cache_type, const char *fname, const int width, const i
                 gettimeofday(&cache[i].tm, NULL);
                 return cache[i].pixbuf;
             }
-            /* Discard the poxbuf if the disk file modification time doesn't match */
-            evict_pixbuf(cache+i);
+            /* Discard all pixbufs for this file if the disk modification time doesn't match */
+            for (size_t j = 0; j < cache_size && cache[j].pixbuf; j++) {
+                if (!strcmp(cache[j].fname, fname)) {
+                    evict_pixbuf(&cache[j]);
+                }
+            }
             qsort(cache, cache_size, sizeof(cached_pixbuf_t), cache_qsort);
         }
     }
@@ -437,7 +446,7 @@ get_pixbuf (cache_type_t cache_type, const char *fname, const int width, const i
 }
 
 void
-queue_cover_callback (void (*callback)(void *user_data), void *user_data) {
+queue_cover_callback (cover_avail_callback callback, void *user_data) {
     if (artwork_plugin && callback) {
         deadbeef->mutex_lock (mutex);
         queue_add(-1, NULL, -1, -1, callback, user_data);
@@ -446,7 +455,7 @@ queue_cover_callback (void (*callback)(void *user_data), void *user_data) {
 }
 
 static void
-cover_avail_callback(const char *fname, const char *artist, const char *album, void *user_data)
+album_art_avail_callback(const char *fname, const char *artist, const char *album, void *user_data)
 {
     if (!user_data) {
         return;
@@ -457,24 +466,24 @@ cover_avail_callback(const char *fname, const char *artist, const char *album, v
     deadbeef->mutex_lock(mutex);
     if (fname) {
         /* An image file is in the disk cache, load it to the pixbuf cache */
-        trace("cover_avail_callback: add to queue %s, %s\n", fname, dt->cache_path);
+        trace("album_art_avail_callback: add to queue %s, %s\n", fname, dt->cache_path);
         queue_add_load(dt->cache_type, dt->cache_path, dt->width, dt->height, dt->callback, dt->user_data);
     }
     else if (get_pixbuf(dt->cache_type, dt->cache_path, dt->width, dt->height)) {
         /* Pixbuf (usually the default) already cached */
-        trace("cover_avail_callback: pixbuf already in cache, do nothing for %s\n", dt->cache_path);
+        trace("album_art_avail_callback: pixbuf already in cache, do nothing for %s\n", dt->cache_path);
         free(dt->cache_path);
     }
     else {
         /* Put the default pixbuf in the cache because no artwork was found */
         struct stat stat_buf;
         if (!stat(dt->cache_path, &stat_buf)) {
-            trace("cover_avail_callback: cache default pixbuf for %s\n", dt->cache_path);
+            trace("album_art_avail_callback: cache default pixbuf for %s\n", dt->cache_path);
             cache_add(dt->cache_type, cover_get_default_pixbuf(), dt->cache_path, stat_buf.st_mtime, -1, -1);
         }
         else {
             /* Image file unexpectedly missing or not empty (unlucky timing on cache expiry, reset, etc) */
-            trace("cover_avail_callback: file gone, do nothing for %s\n", dt->cache_path);
+            trace("album_art_avail_callback: file gone, do nothing for %s\n", dt->cache_path);
             free(dt->cache_path);
         }
         if (dt->callback) {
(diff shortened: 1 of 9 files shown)


cboxdoerfer:
Noticed a bug right after launch: there are no visible column separators with GTK3 GUI and Adwaita GTK+ theme unless the theme override is enabled.

Lithopsian:
Got it. I was drawing a frame instead of a line and it doesn't show in Adwaita.  Also fixed a one pixel offset with the focus line at the top of a group.

cboxdoerfer:
Thx, it works. Another issue I noticed: Up until now deadbeef uses the same colors in listview for tabs and columns. This gives it a consistent and somewhat unique look. With your patches the default config might use different colors (depending on the gtk theme) and therefore might look strange, e.g. with Adwaita it looks like this: 

And while the default config might have different colors, there's no way for the user to change these colors individually. The only way to change the columns colors is by overriding listview and tabstrip theming and then change the tabstrip colors.

My suggestion: Either continue to use same colors for tabstrip and columns, which means using the gtk column colors also for the tabstrip, or we could add additional config entries for column colors in the Appearence -> Playlist section, so that users are able to configure it individually.

cboxdoerfer:
There are also problems with dragged columns. With most themes I've tried (GTK2 and GTK3) the dragged column is barely visible (especially its text (white text on white or light gray background)) and it's also harder to identify the drop point.

Lithopsian:
I think Adwaita is one of the ugliest blandest themes in existence :)

I could change some of the tabstrip or column colours, such as the font, but overall the tabs and column buttons are different widgets and can't really be set to a similar appearance in GTK.  If the theme styles them differently then they are going to look different.  For example, they might apply a strong gradient or curved effect on the tabs, but flat buttons.

So ideally, a way for users to override themes they don't like would be the answer.  I could comment the code back out again, but that leaves the opposite problem where it can be impossible to make the flat override colours look consistent with the preferred theme.

I see that the override uses tabstrip colours, but I only apply the override when the playlist appearance is selected.  Would changing that to use the tabstrip override if the tabstrip appearance is selected solve the problem?

Lithopsian:
Looking at the drag styling, the font colour can be made more consistent and readable, assuming it is readable at all in the theme.  I'll do that anyway.

I don't know about the drag colours.  What GTK styling would make a button stand out more?  I've used prelight because most themes do some form of styling for it, although not Adwaita or Clearlooks.  Selected style is sometimes a different colour, but some themes don't have a particular style for selected buttons (GTK3 Adwaita :lol:)and then you could end up with white text on a normal button again.  Maybe styling a border also?

The drop zone is the active style, which is generally darker or with depressed shadow.  It might be hard to see, but it is the same style as it was before (the same code in GTK2, new commands based on style context in GTK3).  Perhaps it has a different appearance in some themes because it is now based on an actual column header button instead of a plain widget?  I can see it is particularly ugly in Adwaita, Incidentally, the dragged button and the drop zone always use GTK styling, and always have.  Never the override colours.  That sometimes appears jarring, and also users can't change the appearance if it is unhelpful in their theme.  Maybe I should add override colours for those two cases?  Dark and light from the tabstrip colours?

Again, maybe fixing the override to apply when tabstrip colours are selected would be the answer?  People who want their theme, good or bad, can have it.  People who want something else can have it.

Alexey-Yakovenko:
So what is the conclusion of this PR? Is it something that I need to review and possibly merge, or is it broken? I can't figure out from the comments.

plugins/gtkui/coverart.c
```diff
@@ -86,7 +87,7 @@ typedef struct {
     char *cache_path;
     int width;
     int height;
-    void (*callback)(void *user_data);
+    cover_avail_callback callback;
```
Alexey-Yakovenko: please rename to cover_avail_callback_t

plugins/gtkui/coverart.c
```diff
@@ -157,7 +158,20 @@ add_callback(void (*cb)(void *), void *ud)
 }
 
 static void
-queue_add (cache_type_t cache_type, char *fname, const int width, const int height, void (*cb)(void *), void *ud)
+process_query_callbacks(cover_callback_t *callback, const int send)
```
Alexey-Yakovenko: Please change the "const int" to just "int". The code is the same, but easier to read.

plugins/gtkui/coverart.c
```diff
@@ -157,7 +158,20 @@ add_callback(void (*cb)(void *), void *ud)
...
+            trace("coverart: make callback to %p (next=%p)\n", callback->cb, callback->next);
+            callback->cb(callback->ud);
+        }
+        process_query_callbacks(callback->next, send);
+        free(callback);
+    }
+}
+
+static void
+queue_add (cache_type_t cache_type, char *fname, const int width, const int height, cover_avail_callback cb, void *ud)
```
Alexey-Yakovenko: Same here. I don't like this "const int" everywhere.

plugins/gtkui/ddblistview.c
```diff
@@ -57,17 +57,13 @@
...
-extern GtkWidget *theme_button;
-
 G_DEFINE_TYPE (DdbListview, ddb_listview, GTK_TYPE_TABLE);
 
 struct _DdbListviewColumn {
     char *title;
     int width;
     float fwidth; // only in autoresize mode
-    int minheight;
+    minheight_cb minheight_cb;
```
Alexey-Yakovenko: needs to be minheight_cb_t

plugins/gtkui/ddblistview.c
```diff
@@ -125,7 +121,7 @@ ddb_listview_get_row_pos (DdbListview *listview, int pos);
 
 ////// header functions ////
 void
-ddb_listview_header_render (DdbListview *ps, cairo_t *cr);
+ddb_listview_header_render (DdbListview *ps, cairo_t *cr, const int x1, const int x2);
```
Alexey-Yakovenko: remove const from int arguments.

plugins/gtkui/coverart.c
```diff
@@ -184,7 +198,7 @@ queue_add (cache_type_t cache_type, char *fname, const int width, const int heig
 }
 
 static void
-queue_add_load (cache_type_t cache_type, char *fname, const int width, const int height, void (*cb)(void *), void *ud)
+queue_add_load (cache_type_t cache_type, char *fname, const int width, const int height, cover_avail_callback cb, void *ud)
```
Alexey-Yakovenko: const int

plugins/gtkui/coverart.c
```diff
@@ -503,7 +512,7 @@ best_cached_pixbuf(cache_type_t cache_type, const char *path)
 }
 
 static cover_avail_info_t *
-cover_avail_info(cache_type_t cache_type, char *cache_path, const int width, const int height, void (*callback)(void *), void *user_data)
+cover_avail_info(cache_type_t cache_type, char *cache_path, const int width, const int height, cover_avail_callback callback, void *user_data)
```
Alexey-Yakovenko: const int

plugins/gtkui/ddblistview.c
```diff
@@ -734,165 +730,162 @@ ddb_listview_list_pickpoint_y (DdbListview *listview, int y, DdbListviewGroup **
...
 ddb_listview_list_render (DdbListview *listview, cairo_t *cr, int x, int y, int w, int h) {
-    render_idx = 0;
-    cairo_set_line_width (cr, 1);
-    cairo_set_antialias (cr, CAIRO_ANTIALIAS_NONE);
-    GtkWidget *treeview = theme_treeview;
+    GtkWidget *treeview = listview->theme_treeview;
+    const int scrollx = listview->hscrollpos;
+    const int title_height = listview->grouptitle_height;
+    const int row_height = listview->rowheight;
+    const int total_width = listview->totalwidth;
```
Alexey-Yakovenko: const int 4 times

plugins/gtkui/ddblistview.c
```diff
@@ -734,165 +730,162 @@ ddb_listview_list_pickpoint_y (DdbListview *listview, int y, DdbListviewGroup **
...
-        int grp_height_total = grp->height;
-        int group_idx = idx;
-
-        if (grp_y >= y + h + listview->scrollpos) {
-            break;
-        }
-        listview->binding->ref (it);
+    while (grp && grp_y < y + h) {
+        const int grp_height = title_height + grp->num_items * row_height;
+        const int grp_height_total = grp->height;
```
Alexey-Yakovenko: const int 2 times

plugins/gtkui/ddblistview.c
```diff
@@ -734,165 +730,162 @@ ddb_listview_list_pickpoint_y (DdbListview *listview, int y, DdbListviewGroup **
...
         }
-        idx += grp->num_items + 1;
-        abs_idx += grp->num_items;
+        if (it) {
+            listview->binding->unref(it);
+        }
+        idx += grp->num_items;
 
-        int filler = grp_height_total - (grp_height);
+        const int filler = grp_height_total - grp_height;
```
Alexey-Yakovenko: const int

plugins/gtkui/ddblistview.c
```diff
@@ -734,165 +730,162 @@ ddb_listview_list_pickpoint_y (DdbListview *listview, int y, DdbListviewGroup **
...
         if (filler > 0) {
-            int theming = !gtkui_override_listview_colors ();
-            if (theming) {
+            const int yy = grp_y + grp_height;
```
Alexey-Yakovenko: const int

plugins/gtkui/ddblistview.c
```diff
@@ -734,165 +730,162 @@ ddb_listview_list_pickpoint_y (DdbListview *listview, int y, DdbListviewGroup **
...
+                cairo_set_source_rgb(cr, clr.red/65535., clr.green/65535., clr.blue/65535.);
+                cairo_rectangle(cr, x, yy, w, filler);
+                cairo_fill(cr);
             }
         }
 
         // draw album art
-        ddb_listview_list_render_album_art (listview, cr, grp->head, grp->pinned, grp_next_y - listview->scrollpos, -listview->hscrollpos, grp_y + listview->grouptitle_height - listview->scrollpos, listview->totalwidth, grp_height_total);
-        if (grp->pinned == 1 && gtkui_groups_pinned && y <= 0) {
+        const int grp_next_y = grp_y + grp_height_total;
```
Alexey-Yakovenko: const int

plugins/gtkui/ddblistview.c
```diff
@@ -734,165 +730,162 @@ ddb_listview_list_pickpoint_y (DdbListview *listview, int y, DdbListviewGroup **
...
+        grp = grp->next;
     }
-    if (grp_y < y + h + listview->scrollpos) {
-        int hh = y + h - (grp_y - listview->scrollpos);
-//        gdk_draw_rectangle (listview->list->window, listview->list->style->bg_gc[GTK_STATE_NORMAL], TRUE, x, grp_y - listview->scrollpos, w, hh);
-        int theming = !gtkui_override_listview_colors ();
-        if (theming) {
+
+    if (grp_y < y + h) {
+        const int hh = y + h - grp_y;
```
Alexey-Yakovenko: const int

plugins/gtkui/ddblistview.c
```diff
@@ -2312,185 +2304,158 @@ ddb_listview_list_drag_end                   (GtkWidget       *widget,
     ps->scroll_pointer_y = -1;
 }
 
-// #define HEADERS_GTKTHEME
+static void
+draw_cairo_line(cairo_t *cr, GdkColor *color, const int x1, const int y1, const int x2, const int y2) {
```
Alexey-Yakovenko: const int

plugins/gtkui/ddblistview.c
```diff
@@ -2312,185 +2304,158 @@ ddb_listview_list_drag_end                   (GtkWidget       *widget,
...
+    cairo_set_source_rgb (cr, color->red/65535., color->green/65535., color->blue/65535.);
+    cairo_move_to (cr, x1, y1);
+    cairo_line_to (cr, x2, y2);
+    cairo_stroke (cr);
+}
+
+static void
+draw_header_fg(DdbListview *ps, cairo_t *cr, DdbListviewColumn *c, GdkColor *clr, const int x, const int xx, const int h) {
```
Alexey-Yakovenko: const int

plugins/gtkui/ddblistview.c
```diff
@@ -2312,185 +2304,158 @@ ddb_listview_list_drag_end                   (GtkWidget       *widget,
...
+    int text_width = c->width - 10;
+    if (c->sort_order) {
+        const int arrow_sz = 10;
```
Alexey-Yakovenko: const int

plugins/gtkui/ddblistview.c
```diff
@@ -2312,185 +2304,158 @@ ddb_listview_list_drag_end                   (GtkWidget       *widget,
...
+        text_width -= arrow_sz;
+#if GTK_CHECK_VERSION(3,0,0)
+//                gtk_paint_arrow (gtk_widget_get_style (ps->header), cr, GTK_STATE_NORMAL, GTK_SHADOW_NONE, ps->header, NULL, dir, TRUE, xx-arrow_sz-5, h/2-arrow_sz/2, arrow_sz, arrow_sz);
+        gtk_render_arrow(gtk_widget_get_style_context(ps->theme_treeview), cr, c->sort_order*G_PI, xx-arrow_sz-5, h/2-arrow_sz/2, arrow_sz);
+#else
+        const int dir = c->sort_order == 1 ? GTK_ARROW_DOWN : GTK_ARROW_UP;
```
Alexey-Yakovenko: const int

plugins/gtkui/ddblistview.c
```diff
@@ -2312,185 +2304,158 @@ ddb_listview_list_drag_end                   (GtkWidget       *widget,
...
+    }
+
+    float fg[3] = {clr->red/65535., clr->green/65535., clr->blue/65535.};
+    draw_set_fg_color(&ps->hdrctx, fg);
+    draw_text_custom(&ps->hdrctx, x+2, 3, text_width, 0, DDB_COLUMN_FONT, 0, 0, c->title);
+}
 
 void
-ddb_listview_header_render (DdbListview *ps, cairo_t *cr) {
+ddb_listview_header_render (DdbListview *ps, cairo_t *cr, const int x1, const int x2) {
```
Alexey-Yakovenko: const int

plugins/gtkui/ddblistview.c
```diff
@@ -2312,185 +2304,158 @@ ddb_listview_list_drag_end                   (GtkWidget       *widget,
...
-    GtkWidget *widget = ps->header;
-    int x = -ps->hscrollpos;
-    int w = 100;
     GtkAllocation a;
-    gtk_widget_get_allocation (widget, &a);
-    int h = a.height;
-    const char *detail = "button";
+    gtk_widget_get_allocation (ps->header, &a);
+    draw_begin(&ps->hdrctx, cr);
+    const int h = a.height;
```
Alexey-Yakovenko: const int

plugins/gtkui/ddblistview.c
```diff
@@ -2312,185 +2304,158 @@ ddb_listview_list_drag_end                   (GtkWidget       *widget,
...
     int idx = 0;
-    for (c = ps->columns; c; c = c->next, idx++) {
-        w = c->width;
-        int xx = x;
-#if 0
-        if (colhdr_anim.anim_active) {
-            if (idx == colhdr_anim.c2) {
-                xx = colhdr_anim.ax1;
+    for (DdbListviewColumn *c = ps->columns; c && x < x2; c = c->next, idx++) {
+        const int xx = x + c->width;
```
Alexey-Yakovenko: const int

plugins/gtkui/ddblistview.c
```diff
@@ -2312,185 +2304,158 @@ ddb_listview_list_drag_end                   (GtkWidget       *widget,
...
-                // draw empty slot
-                if (x < a.width) {
+        DdbListviewColumn *c = ps->columns;
+        while (c && idx++ != ps->header_dragging) {
+            x += c->width;
+            c = c->next;
+        }
+
+        // draw empty slot
+        const int w = c->width;
```
Alexey-Yakovenko: const int

plugins/gtkui/ddblistview.c
```diff
@@ -3277,28 +3262,33 @@ ddb_listview_free_groups (DdbListview *listview) {
     }
 }
 
+static int
+ddb_listview_min_group_height(DdbListviewColumn *columns) {
+    int min_height = 0;
+    for (DdbListviewColumn *c = columns; c; c = c->next) {
+        if (c->minheight_cb) {
+            const int col_min_height = c->minheight_cb(c->user_data, c->width);
```
Alexey-Yakovenko: const int

plugins/gtkui/ddblistview.c
```diff
@@ -3277,28 +3262,33 @@ ddb_listview_free_groups (DdbListview *listview) {
...
     deadbeef->pl_lock ();
     int old_height = listview->fullheight;
     listview->groups_build_idx = listview->binding->modification_idx ();
     ddb_listview_free_groups (listview);
-
-    listview->plt = deadbeef->plt_get_curr ();
-
     listview->fullheight = 0;
 
+    const int min_height = ddb_listview_min_group_height(listview->columns);
```
Alexey-Yakovenko: const int

plugins/gtkui/ddblistview.h
```diff
@@ -66,6 +60,7 @@ struct _DdbListviewGroup {
     struct _DdbListviewGroup *next;
 };
 
+typedef int (*minheight_cb) (void *user_data, const int width);
```
Alexey-Yakovenko: minheight_cb_t

plugins/gtkui/ddblistview.c
Alexey-Yakovenko: Please write the description in the pull request, about what this ddb_listview_header_render change is doing. I can't easily understand it from the code changes.

Alexey-Yakovenko: It looks like your intention was to make the list view headers look the same as the gtk themed ones, but unfortunately this will break on many GTK themes. We've already been there. Maybe if you're sure this works well enough in GTK3, you could implement it in the ifdef block, or define HEADERS_GTKTHEME=1 under GTK3.

plugins/gtkui/ddblistview.h
```diff
@@ -129,6 +127,9 @@ struct _DdbListview {
     GtkWidget *header;
     GtkWidget *scrollbar;
     GtkWidget *hscrollbar;
+    GtkWidget *theme_treeview;
+    GtkWidget *theme_button;
+    GtkStyle *button_style;
```
Alexey-Yakovenko: There's no need to duplicate those in each ddblistview instance.

Alexey-Yakovenko:
Conclusion:

* naming convention and the const usage needs to be fixed everywhere
* coverart.c (and the rest of cover art changed) look good
* ddblistview changes (except the artwork column fixes possibly) will get rejected in the current state, because I'm sure it will break on many GTK2 themes
* plcommon.c -- mostly looks good.

Alexey-Yakovenko:
> My suggestion: Either continue to use same colors for tabstrip and columns, which means using the gtk column colors also for the tabstrip (or the other way arround), or we could add additional config entries for column colors in the Appearence -> Playlist section, so that users are able to configure it individually.

This is a very valid point. The idea was always to have the consistent look without visible transition between the tab strip and the list view header. Now it looks really wrong.

Lithopsian:
The previous header rendering code was not based on a button style, let alone a treeview header button.  If you point me to a GTK2 theme that is broken, I'll see if it is or can be fixed in this way.  Otherwise I'll comment it all back out.  Note that the dragged headers have always had GTK styling and still do, never override colours.  Only the normal headers were styled with the tabstrip overrides.  I'll make the tabstrip override preference override these also.

Lithopsian:
ddb_listview_header_render does what it always did, renders the header "buttons".  I'll add some more comments to it.

Alexey-Yakovenko:
> If you point me to a GTK2 theme that is broken, I'll see if it is or can be fixed in this way.

Originally when deadbeef was created, those kinds of problems were found and resolved slowly, mostly via bug reports from the users. I don't have those broken GTK themes from 4-5 years back anymore. If you're going to use GTK theming for anything in deadbeef -- the reason should be very significant, and the changes will need to be tested with at least a few dozens of different GTK2&3 themes before we could ship this.

Lithopsian:
GTK header theming #ifdef'd out in both GTK2 and GTK3.  I think GTK3 theming might work better with the new style context method, but can't be sure without examples to test.

cboxdoerfer:
> I think Adwaita is one of the ugliest blandest themes in existence :)

Well, it's the default gtk theme and lots of distributions and users use it. So I don't think we can ignore it. Appart from that, most of the issues I mentioned apply to most themes I tried (clearlooks, numix, elementary, xfce*, ...).

> I could change some of the tabstrip or column colours, such as the font, but overall the tabs and 
> column buttons are different widgets and can't really be set to a similar appearance in GTK. If the 
> theme styles them differently then they are going to look different. For example, they might apply a 
> strong gradient or curved effect on the tabs, but flat buttons.

But in deadbeef those aren't two completely different widgets, they were designed to look nice together. So I guess the first priority isn't to make each one look as close as it gets to the according gtk widgets, but make them look nice together. Maybe if the user decides to theme them individually we should enable that, but at least in the default config a consistent look is more important.

> I see that the override uses tabstrip colours, but I only apply the override when the playlist 
> appearance is selected. Would changing that to use the tabstrip override if the tabstrip appearance is 
> selected solve the problem?

Imho the biggest problem are the different colors for tab strip and columns in the default config. Maybe applying the same colors of the columns to the tab strip looks nice, but this has to be tested with lots of themes and compared with the current approach.

cboxdoerfer:
btw, Alexey what's wrong with the usage of const qualifiers?

Alexey-Yakovenko:
nothing is wrong, where they do any good. In all of the cases above they just make code bigger and harder to read.

Lithopsian:
Removed const int.
typedef naming is fixed in several places.
#ifdef'd out GTK styling for the column header (note that the divider lines are still GTK themed unless the tabstrip override is selected, and the sort arrows and dragged button are always from the GTK theme)
theme_treeview is still one per listview instance, tidier and safer for a widget that needs to be realised and then hidden.

plugins/gtkui/plcommon.c
```diff
@@ -149,193 +170,148 @@ tf_redraw_cb (gpointer user_data) {
...
-            // destination coordinates and sizes
-            int art_x = x + ART_PADDING_HORZ;
-            int art_y = y + ART_PADDING_VERT;
-            int art_w = listview->cover_size;
-            int art_h = height;
+static void
+cover_draw_cairo (GdkPixbuf *pixbuf, const int x, const int min_y, const int max_y, const int width, const int height, cairo_t *cr, const int filter) {
+    const int pw = gdk_pixbuf_get_width(pixbuf);
+    const int ph = gdk_pixbuf_get_height(pixbuf);
+    const int real_y = min(min_y, max_y - ph);
```
Alexey-Yakovenko: more const ints

Alexey-Yakovenko:
I can see more "const int"s in the patch.

> theme_treeview is still one per listview instance, tidier and safer for a widget that needs to be realised and then hidden.

It is not good enough argument for this change. One thing that you should realise is that having those theming widgets is a hack, from beginning to the end.

This hack is to extract the needed styling information from e.g. GtkTreeView, then apply it to DdbListView (same for the button). The idea is that those widgets are "singletons". It's a really bad idea to duplicate them for each time they are needed.

So I have to disagree -- it's not safer or tidier to apply the hack 2 times instead of 1.

The next thing that you should remember, is that by changing something that _works_, and doing it differently, just for the sake of making it (arguably) more tidy, is a wrong thing. You change the way the code works, so by default I'm assuming that it will break under certain circumstances. This specific code is very sensitive, and breaks _really_ simple. So please put it back the way it was.

Lithopsian:
Back to Hackville City Arizona.  Let me know if you ever want to do this properly.  GTK3 provides everything needed to do this entirely legitimately, without even dummy widgets if you don't want them.

Alexey-Yakovenko:
Thanks, that looks good now. I will need to read-out all the code once again, and test it on a branch, then will try merging.

> GTK3 provides everything needed to do this entirely legitimately, without even dummy widgets if you don't want them.

Depends how much #ifdefs that would add. Maybe we need to abstract this out, and do the styling in a separate module. The biggest concern is to avoid adding add more clutter to ddblistview.c

Lithopsian:
Abstracting the basic treeview background drawing operations makes sense.  At the moment, every time we draw a piece of background, it is wrapped in a GTK3 ifdef and an override check.  They are all identical barring a couple of things like a selected flag and an odd/even flag.

Changing the GTK3 styling would need some initial setup to create the style contexts, but the drawing itself would be the same as now (now meaning as in this pull request, using gtk_render_* instead of gtk_paint_*).

Alexey-Yakovenko:
The column separators look differently in GTK2 vs GTK3 with the same theme. Is that expected?

cboxdoerfer:
Drag and drop indicators also look different:
without PR: http://i.imgur.com/ysVmyE8.png
with PR: http://i.imgur.com/v4oDu83.png

Both screenshots are from GTK3 UI.

plugins/gtkui/ddblistview.c
```diff
@@ -734,165 +724,165 @@ ddb_listview_list_pickpoint_y (DdbListview *listview, int y, DdbListviewGroup **
...
+        DdbListviewIter it = grp->head;
+        listview->binding->ref(it);
+        for (int i = 0, yy = grp_y + title_height; it && i < grp->num_items && yy < y + h; i++, yy += row_height) {
+            if (yy + row_height >= y) {
+// GTK does this already, plus we are going to draw over every inch of it
+//                GtkStyle *st = gtk_widget_get_style (listview->list);
+//                GdkColor *clr = &st->bg[GTK_STATE_NORMAL];
+//                cairo_set_source_rgb (cr, clr->red/65535., clr->green/65535., clr->blue/65535.);
+//                cairo_rectangle (cr, -scrollx, yy, total_width, row_height);
+//                cairo_fill (cr);
```
Alexey-Yakovenko: Commenting this out is wrong. I'm not sure if I remember this correctly 100%, but this code was in there intentionally, perhaps for "translucent" theming support. I'm uncommenting this during merge.

Alexey-Yakovenko:
@cboxdoerfer 
> Drag and drop indicators also look different:

I like the new look more than the old one, so it's good to have. Too bad it's only on GTK3.

Alexey-Yakovenko:
This is now on master, let's use it for a bit and see if it's good enough for 0.7

plugins/gtkui/ddblistview.c
```diff
@@ -2312,185 +2300,174 @@ ddb_listview_list_drag_end                   (GtkWidget       *widget,
...
-                    }
-                }
-                draw_text_custom (&ps->hdrctx, xx + 5, 3, ww, 0, DDB_COLUMN_FONT, 0, 0, c->title);
-            }
-            if (sort) {
-                int dir = sort == 1 ? GTK_ARROW_DOWN : GTK_ARROW_UP;
 #if GTK_CHECK_VERSION(3,0,0)
-                gtk_paint_arrow (gtk_widget_get_style (widget), cr, GTK_STATE_NORMAL, GTK_SHADOW_NONE, widget, NULL, dir, TRUE, xx + w-arrow_sz-5, a.height/2-arrow_sz/2, arrow_sz, arrow_sz);
+                    GtkStyleContext *context = gtk_widget_get_style_context(theme_treeview);
+                    gtk_style_context_add_class(context, GTK_STYLE_CLASS_SEPARATOR);
```
Alexey-Yakovenko: This is not compatible with GTK3.0.0

Alexey-Yakovenko: If I restore the old code, it no longer looks correct (renders as a 1px black line). Any ideas why is that?

Lithopsian:
If I comment out the gtk_paint_arrow line (modified slightly as follows for new variable names), it works in 3.4:
                gtk_paint_arrow (gtk_widget_get_style (ps->header), cr, GTK_STATE_NORMAL, GTK_SHADOW_NONE, ps->header, NULL, c->sort_order ? GTK_ARROW_DOWN : GTK_ARROW_UP, TRUE, xx-arrow_sz-5, h/2-arrow_sz/2, arrow_sz, arrow_sz);

I don't have 3.0.  What happens there?  gtk_render_arrow was available in 3.0, what doesn't work?  Is it a theme thing?

Alexey-Yakovenko:
```ddblistview.c:2382:58: error: ‘GTK_STYLE_CLASS_SEPARATOR’ undeclared (first use in this function)```

Lithopsian:
I didn't understand your inline code.  Is the problem with the separator lines or the arrow?  I thought it was the render_arrow, but no?

Seems like GTK_STYLE_CLASS_SEPARATOR wasn't added until 3.2 (.3)?  Without it, gtk_render_line may draw a plain line instead of a themed separator.  A workaround would be to pass the "separator" string explicitly instead of using the define.

Lithopsian:
The GTK3 drag styling looks slightly unreal to me, like a plain cutout.  In GTK2 it has a shadow.  I think it needs a frame in GTK3, then it would look like "something" instead of a coloured rectangle.

screenshot - 301015 - 14 31 27

Alexey-Yakovenko:
> I didn't understand your inline code. Is the problem with the separator lines or the arrow? I thought it was the render_arrow, but no?

It's not about the arrow, it's about the GTK_STYLE_CLASS_SEPARATOR.

I restored your new code, and replaced the GTK_STYLE_CLASS_SEPARATOR with "separator".
Not sure if that will work correctly on 3.0.0, will need to check if the separator was implemented in there.

Alexey-Yakovenko:
(the code review comments appy to the last line in the quoted code block)

Lithopsian:
Or if you prefer the really flat GTK3 drag look, you can have it in GTK2 by using gtk_paint_flat_box instead of gtk_paint_box.  Or just by removing the shadow.

Alexey-Yakovenko:
Adding a very subtle 1px frame to it would be nice, and then converting GTK2 code to do the same.

Lithopsian:
Was there agreement on this?  Seems like one vote for the new plain look (closer to the override style than the GTK2 style) and one for a frame?

Any border round the dragged header button would depend on the theme, just as the "shadow" in GTK2 is really drawn by the theme.  Usually it would be a 1-2px lightly styled line, but could in theory be almost anything.  The GTK2 code already has a frame, created by asking for a shadow in gtk_paint_box().  As you can see, this is a very subtle effect in most themes, but it could be painted as a completely plain rectangle if that is preferred.

It would also be possible to put a cairo line around the rendered area.  I experimented with this and it looked slightly strange to me, cartoonish.  The themed borders generally have a slight 3-D appearance with a couple of different colours, and it isn't until you draw a completely plain frame in one colour that this is noticeable.

Alexey-Yakovenko:
In the perfect world, the dragged header button is supposed to look the same as the one from the GtkTreeView. I don't have a strong opinion about how it should look in practice. I'm fine with whatever it's now.

Lithopsian:
I can make the dragged header look exactly like a treeview header in GTK3, but is that rally what you want?  Header buttons are often not styled at all.  For example in Adwaita thay are just flat white, not even a border.  Normal buttons are usually styled and might be a better choice.  There is also no "selected" style for buttons, so they won't turn bright blue when they're dragged.  There is usually some "prelight" styling, generally a more subtle highlighting.  Again, real buttons have stronger styling, Adwaita prelight on a treeview header just slightly darkens the text colour.  I'll try it and see what you think, then we can tweak to look less (or more) like a button.

Issue #1304, opened 2015-10-23
Resizing a column sometimes starts a drag instead


Lithopsian:
This has been driving me nuts for a while.  I used to think I was clicking in the wrong place or moving the mouse, but now I think there is a real problem.  I can reproduce easily in GTK3, but never in GTK2.

Place the mouse over a column header divider.  Put it as far to the right as possible on a divider and still have the double arrow cursor.  Now click and move the mouse to try and resize.  A drag starts instead.  Just me?

Lithopsian:
I got confirmation from a couple of people that this can be reproduced.  I also tested on several other machines and OS's.

So a bit of digging finds that it is a rounding error.  Event coordinates are doubles, but in GTK2 they appear to be pre-rounded and never have a fractional part.  In GTK3 they have fractions.  The test for the sizing cursor assigns the x coordinate to an int and then compares to the column boundary, but the test for dragging compares the double directly to the column boundary.

I'll prepare a fix for later.

Alexey-Yakovenko:
great!

Lithopsian:
Starting a resize can also "jump".  The active region for resizing is the last four pixels of each column, and the divider line is drawn approximately 2 pixels to the left of the "official" column edge.  When a resize begins, the edge of the column is immediately moved to the cursor position, which means it jumps up to four pixels making the column smaller.  The visible divider line becomes aligned about 2 pixels to the left of the apparent cursor position.

I plan to fix the column size relative to the cursor position when it is first clicked, the same way that a drag location is exactly relative to the initial click.  There will be no jump in the position of the divider or the size of the column.  The divider will remain at the position relative to the cursor that it was when the resize started.

andreas--e:
Lithopsian, I'm so happy you found a fix for this. I've known this problem on GTK3 since early 0.6 versions, so for very very long. Thanks!

Lithopsian:
The threshold for starting a header drag is also broken.  There is code for not starting a drag operation until the mouse has moved a configurable (default 8) number of pixels, but gtk_drag_check_threshold() is called incorrectly and always returns TRUE.  So header drags start as soon as the mouse moves.  This contributes to the problem by starting drags accidentally, although with the new DND algorithm it is now easy to terminate the drag without anything happening.

I plan on fixing this so that header drags only start after the threshold is exceeded.  This is consistent with playlist dragging, but let me know if you don't want it and I'll just take out the broken code.

[...]

PR #1314, opened 2015-11-01
Playlist

Lithopsian:
Fix issue #1304.
Fix rounding error in GTK3 so drags don't start when the resize cursor is showing.
Fix jumping at the start of a resize, remembering the exact drag start position.
Re-instate threshold checking for starting a drag, must move 8 pixels horizontally.
Do not display the resize cursor during a drag prepare, including while trying to drag from beyond the last column.
Display the resize cursor reliably and promptly (ie. without needing to move it again) when "entering from another window", such as button release events from presses over other widgets.
Prevent sorting if the button release doesn't happen over the header (eg. on the tabs or list).
Prevent accidental sorting when trying to resize or start a drag, or sorting an unexpected column, by having a small dead zone near the separator line.

### DIFF
--- a/plugins/gtkui/ddblistview.c
+++ b/plugins/gtkui/ddblistview.c
@@ -2617,7 +2617,7 @@ ddb_listview_header_realize                      (GtkWidget       *widget,
 static void
 set_header_cursor (DdbListview *listview, gdouble mousex)
 {
-    int x = 0;
+    int x = -listview->hscrollpos;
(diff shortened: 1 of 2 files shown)


Alexey-Yakovenko:
merged

Alexey-Yakovenko:
@Lithopsian looks like something is misconfigured on your machine, as we're getting this email in your commits: `iann@crunchbang`

Issue #1310, opened 2015-10-28
Paging up in a playlist briefly scrolls to the end of the list when reaching the first item


Lithopsian:
1. Create a playlist long enough to have several pages and a scrollbar.
2. Page down, then back up to the top.  Make sure that the last page up keypress will go beyond the first item.

When the cursor would page beyond the first item, the playlist will briefly scroll to the end of the range, then quickly redraw with the first item selected.  This is easier to see with album art since the wrong image will flash up for a moment.If groups are not pinned then the first item in the playlist will be selected, but the list will not be scrolled right to the top.  This is the position reached naturally by selecting the first item when positioned at the end of the list.

I think this is as simple as setting the scroll position to the upper adjustment value instead of the lower when paging up beyond cursor position zero.

Lithopsian:
This shouldn't happen any more.  Paging past the top of the list sets the scroll position to the lower bound instead of the upper.

Alexey-Yakovenko:
Why can't you close your own bug reports after fixing them yourself?
I thought you can do so either by clicking "close" on github, or writing `fixes #XYZ` in the commit message.

Lithopsian:
Sorry, I guess I can close my own issues.  I didn't realise I had such power.

Issue #1308, opened 2015-10-27
Playlist header columns not redrawn after widget resize


Lithopsian:
To reproduce:
1. Size the playlist widget to include a horizontal scrollbar.
2. Scroll the playlist all the way to the left (ie. scrollbar slider to the right).
3. Resize the playlist widget until the horizontal scrollbar disappears (and quite a bit more makes the bug easier to spot).
The playlist column headers are left in a state that is not consistent with the columns themselves.  They are still partially scrolled even though the columns themselves are not.
4. Resize back until the horizontal scrollbar reappears.
The playlist columns and headers are now consistent and set to their starting positions, but the scrollbar indicates that they are still scrolled.  It helps to resize quickly.

Lithopsian:
There is also a drawing artefact on the horizontal scrollbar after some resizes.  I can reproduce it fairly consistently:
1. Make the playlist quite narrow and scroll all the way to the right.
2. Quickly resize the playlist so that the horizontal scrollbar disappears.
3. Gradually resize it narrower until the scrollbar reappears.  The scrollbar slider is drawn over the right hand button, although the button arrow is still visible.

This only seems to happen in GTK2 so could be considered a GTK bug.  It seems to happen because the position value passed to the adjustment constructor is not clamped and can exceed the largest legal value.  I can fix that, and then the bug should go away.  Values set after the adjustment has been added to the scrollbar appear to be clamped correctly in versions I tested.

Lithopsian:
This should be fixed.

Issue #1307, opened 2015-10-27
Playlist horizontal scrolling is desperately slow with the thumb buttons ...


Lithopsian:
... and rather fast with a mouse wheel.

The scroll step is set to 1 pixel, compared to 20 pixels for the vertical scroll bar.  Should they both be 20?  Or something a little less for horizontal scrolling?

Mousewheel scrolling with the mouse over the drawarea also scrolls 20 pixels per step.  With the mouse over the scrollbar itself, a different increment is used.  I believe it is the page size to the power 2/3.  The vertical scrollbar page size is set to half of the drawarea height and it produces nice results with the mousehweel.

The horizontal scrollbar page size is set to the full width of the drawarea and tends to just skip from one end of the range to the other when scrolling, since there isn't typically a huge scroll range.  There is the same effect when clicking on the scrollbar, a very large "page" jump.  Should the page size be smaller?  Half the width like the vertical scrollbar?

Lithopsian:
A slight correction: mousewheel scrolling over the list is 40 pixels at a time (the vertical thumb buttons are 20 pixels, the horizontal ones 1 pixel at a time), about the same as mousewheel scrolling over the vertical scrollbar in a list that is 500 pixels tall, or the horizontal scrollbar in a list 250 pixels wide.  Larger playlist widgets scroll more quickly when wheeling over the scrollbars, but the same when wheeling over the drawing area.  With smooth scrolling in GTK3, the 40 pixel step is modified by delta values which depend on the device.

Alexey-Yakovenko:
How do you make the thumb buttons to emit scroll events? Is that something preconfigured in certain DE/WM? I'm running ubuntu 12.04 + virtualbox, the thumb buttons simply do nothing..

Lithopsian:
I don't know.  Is it a virtualbox thing?  Many GTK3 themes don't have scrollbar buttons, but the ones that do always seem to work.  They don't actually emit scroll events, just change the value of the scrollbar range which we detect as a "value_changed" signal.

Alexey-Yakovenko:
Oh, I misunderstood what you meant by thumb buttons I guess..
I thought you mean those physical buttons on the left side of the mouse, which you press with your thumb.

Lithopsian:
Sorry, I meant the buttons at the top and bottom of the scrollbar.  Or in this case the left and right of the horizontal scrollbar.

[...]

PR #1316, opened 2015-11-02
 gtkui: fix playlist header redraw after resize, fix horizontal scrolllbar misdraw, change scrolling steps

Lithopsian:
Fix header redraw when sizing after horizontal scrolling (issue #1308)
Work round horizontal scrollbar drawing artefact in GTK2 (issue #1308)
Fix scrolling to bottom of playlist momentarily while paging up (issue #1310)
Change scroll steps to be consistent horizontally and vertically, and when using a wheel over the drawing area or scrollbar (issue #1307)
Fingers crossed, caught up and merged all my changes.

### DIFF
--- a/plugins/gtkui/ddblistview.c
+++ b/plugins/gtkui/ddblistview.c
@@ -48,6 +48,7 @@
 #define SCROLL_STEP 20
 #define AUTOSCROLL_UPDATE_FREQ 0.01f
 #define NUM_CHANGED_ROWS_BEFORE_FULL_REDRAW 10
+#define MIN_COLUMN_WIDTH 16
 
 //#define trace(...) { fprintf(stderr, __VA_ARGS__); }
 #define trace(fmt,...)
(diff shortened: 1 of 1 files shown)


Lithopsian:
I notice that gtk_adjustment_get_lower() and related functions are used.  They were introduced in GTK 2.14.  I guess that is about the earliest supported version now?  Pretty old, but there is still some code specifically to support older versions.

plugins/gtkui/ddblistview.c
```diff
@@ -1047,44 +1036,49 @@ ddb_listview_list_expose_event               (GtkWidget       *widget,
 }
 #endif
 
-gboolean
-ddb_listview_vscroll_event               (GtkWidget       *widget,
+static void
+scroll_by (GtkWidget *scrollbar, gdouble delta)
+{
+    GtkRange *range = GTK_RANGE(scrollbar);
+    gdouble step = pow(gtk_adjustment_get_page_size(gtk_range_get_adjustment(range)), 2./3.);
```
Alexey-Yakovenko: what is the math behind this?

Lithopsian:
It is lifted directly from GTK, to match the behaviour of the mouse wheel over the scrollbar itself.  The effect is to mildly reduce the wheel step size as the page size decreases.

Alexey-Yakovenko:
I see.. I think we should make a long term goal of converting all this mess to GtkScrolledWindow, or whatever is appropriate.. Originally I didn't see it as a viable option, but it worked ok with Cocoa, so I don't see why we can't do the same in GTK.

Lithopsian:
I've got a version working well with scrolling by gdk_window_scroll() plus extra redraws for pinned groups.  The same logic could probably be extended to a GtkScrolledWindow with a viewport.  Maybe see how step 1 works out before going the whole way ...

Did you do the same sort of thing in Coacoa?  Or something completely different?

Alexey-Yakovenko:
Yeah, I ported the DdbListview to cocoa. But used the scrolled view for scrolling, instead of manual scrollbar handling.

Issue #1318, opened 2015-11-04
Play item text shows behind pinned title in themes with transparent row background


Lithopsian:
The treeview row background colour on some themes is transparent.  These themes include HighContrast and the default (ie. broken) theme in GTK3.  On these themes, a pinned group title background does not hide any underlying text.  Selected row backgrounds and the focus box can also show across the album art column for the same reason.

![2015-11-04--1446665680_1280x1024_scrot](https://cloud.githubusercontent.com/assets/7702742/10949526/4109891e-832b-11e5-9720-5bcd60978402.png)

Alexey-Yakovenko:
I'm also getting black text on black background with the same use case, plus the problem you described above (GTK 3.4.0, ubuntu 12.04)

Lithopsian:
Where is the black text on black background showing?  Is that HighContrast or HighContrastInverse?

cboxdoerfer:
Works fine here with both GTK2 and GTK3 ui. Tried it with Adwaita, HighContrast, HighContrastInverse and all other themes I have installed (numix, elementary, arc, ...). The only exception, and that has been known already, is the default Mint theme with the GTK2 ui which triggers that bug.

@Lithopsian which version of GTK3 are you using? I'm running 3.18.

Alexey-Yakovenko:
> Where is the black text on black background showing? Is that HighContrast or HighContrastInverse?

entire playlist is black except the column headers and the selected track(s).
It's on high contrast.

Lithopsian:
Solid black indicates a theme failure, something not defined and coming out black.  Happens a lot in GTK3 as features are changed or added.  Strange though, because I have GTK 3.4.2 here and HighContrast looks basically fine.

I don't have GTK 3.18!  Maybe they updated the standard themes to have a defined background colour for treeview rows?  Do you see the transparent backgrounds with Mint-X theme in GTK2?  In Mint or something else?  I couldn't find a GTK2 theme that did this.

I could work round this fairly easily.  Whenever the playlist draws a treeview background that it expects to hide something already drawn, just paint it with the window background colour first.  It should only happen with the pinned title and the album art column, that I can think of.  Do you think it is worth it?

cboxdoerfer:
Yes, with Mint-X themes and GTK2 UI I can see the transparent backgrounds, but all other themes I tried seem to be fine. I'm running Debian Testing and Arch Linux, both with GTK 3.18 and GTK 2.24.28.

[...]

Issue #1306, opened 2015-10-26
Text drawing may overflow the specified width


Lithopsian:
The drawing functions (eg. draw_text_custom) ellipsize text to keep its length within the specified width parameter.  However, the ellipsis itself is never clipped and can exceed the maximum allowed width.  This causes the ellipsis character to run over the boundary of narrow columns onto the next column header.  I noticed it when resizing the headers, but it can happen on the item rows also if the font is large enough.

I guess this is effectively a Pango bug.  There is a hard-coded minimum column size, but that cannot cover the possible size of an ellipsis character in any font or font size.  It would be tricky to dynamically determine a minimum column size, since it could change after a column size has been set.

A simple solution would be to clip the cairo region being used to draw the text, so even when Pango renders text wider than the column it will not appear outside the column boundaries.

Lithopsian:
The PR uses cairo to clip the drawing area to the width that the text is expected to be cropped to.  This is applied to column headers and column contents.

This clipping isn't done for the group title.  The group title is never ellipsized, just cropped to the width of the playlist.  It cannot overflow the boundaries of the playlist.

PR #1322, opened 2015-11-06
Playlist column clipping, abstract treeview rendering and button styling

Lithopsian:
Clip column text so that the ellipsis cannot overflow a column or header (issue #1306).
Abstract treeview background rendering to confine all the #ifdefs to one function (issue #1318 about transparent backgrounds is not fixed yet).
Make GTK3 styling for dragged headers more button-like.  They are styled like default buttons (not true column header buttons), pre-lit and focused, which is about the strongest button styling available in most themes.  In GTK2 the dragged headers still have the flat "selected" colouring.  Non-dragged buttons are still flat using the override colours.


Lithopsian:
It is supposed to look like a treeview background.  The actual DdbListview widget background is different, and I'm going to have to draw that separately to fix the transparency issues.  I can call it whatever you want though.

Issue #1330, opened 2015-11-13
Search playlist fonts don't update when preferences are changed


Lithopsian:
When the preferences for playlist fonts are changed, to and from overrides or to a different override font, the main playlist updates to match.  The search playlist is redrawn so that background colours change, and text colours change, but the fonts themselves and font sizes are not updated.  If the search dialog window is resized, the fonts will then update.

Lithopsian:
The search list also doesn't respond when a playlist is paused or unpaused.

While I'm here, does anybody know of other things the search window doesn't keep up with?

Lithopsian:
Should the sorted column persist between search windows?  At the moment, the same window is just hidden and shown again so the sorted column remains, but it seems like it should return to a plain initial state.

Lithopsian:
SELECTIONFOCUS also doesn't work properly.  The main window playlist will scroll to the first selected item, but doesn't move the cursor.  It looks like a random pointer is being passed as a ddb_playlist_t.  I don't know what is getting set behind the scenes, but it isn't the requested cursor.

Issue #1328, opened 2015-11-09
Playlist auto-resize doesn't maintain accurate column widths


Lithopsian:
In auto-resize mode, the horizontal scrollbar can still appear when reducing the playlist width, even when the original widths were set to have no scrollbar, which sort of defeats the purpose.  It looks like a rounding error, but it isn't.  I think the fraction calculation is based on a width including the vertical scrollbar, but the vertical scrollbar doesn't shrink so eventually the other columns plus the scrollbar become too big for the widget.  There may be rounding issues on top of that, since sometimes the scrollbar will appear, then disappear when the widget gets slightly narrower, then reappear again.

Also, the columns don't respect the minimum width in auto-resize mode.  This does (or would, if the first part of this bug is fixed) prevent the horizontal scrollbar from ever appearing, but it does sort of make it pointless having a minimum column width.  Or possibly the calculation could be modified to still maintain the desired total width even when columns start hitting their minimum widths.

Lithopsian:
I've been playing with this - very tricky to make something that always behaves nicely and smart enough to replace a human being, but I think it can be improved:

1. Avoid randomly showing and hiding the horizontal scrollbar.  Losing or gaining a pixel or two through rounding isn't a problem unless it shouts out by changing the scrollbar.  To be reliable at all sizes and all rounding situations, the code needs to detect the scrollbar before resizing and keep adjusting until it has a suitable size to keep it that way (ie. stay on or off).

2. Resize when the vertical scrollbar appears or disappears.  This is a particular and common cause of case 1, but needs care to avoid feedback loops (mostly from the album art column resizing the group heights) where a resize reverses the change that caused it, and either the vertical scrollbar or both scrollbars will flash on and off.  The danger zone for such feedback needs to be calculated and the album art column not resized until the widget changes enough to permanently show or hide the scrollbar.

3. Maintain minimum column widths.  Not 100% essential, but manual sizing has a minimum column width, so auto-sizing probably should too.  Depending on the horizontal scrollbar state, it may be necessary to steal back pixels from larger columns when a minimum size is reached.  Ultimately, it will become impossible and the horizontal scrollbar will be forced on when all columns reach their minimum size.  I think this would be better than just keep shrinking until the columns reach zero width.

[...]

PR #1333, opened 2015-11-17
gtkui: playlist autoresize 

Lithopsian:
Change autoresize in line with issue #1328.  This addresses scrollbar and rounding issues when resizing, and introduces minimum column widths in autoresize mode.

Font updates are moved out of the configure_event handler, but issue #1330 is only partially fixed.

### DIFF
--- a/plugins/gtkui/ddblistview.c
+++ b/plugins/gtkui/ddblistview.c
@@ -183,7 +183,7 @@ ddb_listview_header_expose_event                 (GtkWidget       *widget,
                                         gpointer         user_data);
 #endif
 
-gboolean
+static gboolean
 ddb_listview_header_configure_event              (GtkWidget       *widget,
                                         GdkEventConfigure *event,
                                         gpointer         user_data);
@@ -210,7 +210,7 @@ ddb_listview_header_button_release_event         (GtkWidget       *widget,
 static gboolean
 ddb_listview_header_enter (GtkWidget *widget, GdkEventCrossing *event, gpointer user_data);
 
-gboolean
+static gboolean
 ddb_listview_list_configure_event            (GtkWidget       *widget,
                                         GdkEventConfigure *event,
                                         gpointer         user_data);
@@ -357,7 +357,6 @@ ddb_listview_init(DdbListview *listview)
     listview->groups = NULL;
     listview->plt = NULL;
 
-    listview->block_redraw_on_scroll = 0;
     listview->calculated_grouptitle_height = DEFAULT_GROUP_TITLE_HEIGHT;
 
     listview->cursor_sz = NULL;
@@ -576,6 +575,10 @@ ddb_listview_destroy(GObject *object)
 
 void
 ddb_listview_refresh (DdbListview *listview, uint32_t flags) {
+    if (flags & DDB_REFRESH_FONTS) {
+        ddb_listview_update_fonts (listview);
+        ddb_listview_header_update_fonts (listview);
+    }
     if (flags & DDB_LIST_CHANGED) {
         ddb_listview_build_groups (listview);
     }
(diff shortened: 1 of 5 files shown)


plugins/gtkui/ddblistview.h
```diff
@@ -254,6 +253,7 @@ enum {
     DDB_REFRESH_VSCROLL = 4,
     DDB_REFRESH_LIST    = 8,
     DDB_LIST_CHANGED    = 16,
+    DDB_REFRESH_FONTS   = 32,
```
Alexey-Yakovenko: I don't like this being in the APIs.
This is very GTK-specific.
It can be handled on GTK level by sending e.g. configure signal when the font configuration changes, which in turn can be detected in the DB_EV_CONFIGCHANGED handler somewhere.

Lithopsian: REFRESH_STYLING?  REFRESH_CONFIG would be even more generic?  It is called for any change in the config, even just for the preferences dialog being opened.

Alexey-Yakovenko:
A general suggestion... When you feel like refactoring is needed, e.g. extracting a piece of code into a separate function, would be nice to do that in a separate PR.

Doing this will make it possible reading the actual code changes for me.

Right now it's very difficult, because the code has not just changed, but also has been reformatted, and moved to a new location.

Alexey-Yakovenko:
An example of this is right here..

The code looked like this
```
assert (c);
listview->columns = c->next;
ddb_listview_column_free (listview, c);
listview->binding->columns_changed (listview);
```

now it looks like this, but it has been also moved to the new place
```
assert(c);
set_column_width(listview, c, 0);
*next = c->next;
ddb_listview_column_free(listview, c);
listview->binding->columns_changed(listview);
```

It's very hard to actually see what changed, and in what way.

What you should have done instead:

* Either refactor the code first without changing in, then apply the code fixes later
* Or apply the code fixes, without refactoring, then do the refactoring later

Either way would make it easy for both of us.

plugins/gtkui/ddblistview.h
Alexey-Yakovenko: well, then call it CONFIG_CHANGED -- that's the most appropriate, and reflects exactly what it's reacting to.

plugins/gtkui/ddblistview.c
```diff
@@ -2617,52 +2540,143 @@ ddb_listview_update_scroll_ref_point (DdbListview *ps)
...
+    int full_height = 0;
+    deadbeef->pl_lock();
+    for (DdbListviewGroup *grp = listview->groups; grp; grp = grp->next) {
+        full_height += listview->grouptitle_height + max(grp->num_items * listview->rowheight, min_height);
+    }
+    deadbeef->pl_unlock();
+    return full_height;
+}
+
+// Calculate if this width would cause auto-resize thrashing
```
Alexey-Yakovenko: I could only guess that you meant "whether" instead of "if" in the comment.

So the comment should possibly be "Check whether this column's width would cause autoresize thrashing"

But this doesn't really explain what this function is doing.

The function name is very uninformative too.

Please fix.

I can't really give any concrete advise, because I can't understand what this code is doing.

plugins/gtkui/ddblistview.c
```diff
@@ -3316,100 +3293,81 @@ ddb_listview_min_group_height(DdbListviewColumn *columns) {
...
+}
+
+static void
+ddb_listview_build_groups (DdbListview *listview) {
+    listview->groups_build_idx = listview->binding->modification_idx();
+    ddb_listview_free_groups(listview);
+    listview->plt = deadbeef->plt_get_curr();
+
+    deadbeef->pl_lock();
+    int height = build_groups(listview);
```
Alexey-Yakovenko: I think I already commented on this, but will write again.. This extraction doesn't make any logical sense, and doesn't improve code structure in any way. Please move the build_groups content in here. This at least will help me see what have you changed.

Alexey-Yakovenko:
That should be it for now. Please fix everything I wrote in the comments, and I'll do a second round of review on the updated PR.
I tried running the build off this branch, and played around with autoresizing columns. I couldn't spot any regressions, and I think it works very well in general.

Lithopsian:
Merge branch scroll into master

PR #1339, opened 2015-11-21
Fix precedence of album artist variants, add BAND

saivert:
This should keep it more in line with foobar2000 I think. Comments welcome!


Alexey-Yakovenko:
merged

PR #1343, opened 2015-11-23
fix playitem copy, allow playitem drop onto tab

Lithopsian:
Changed plt_copy_items() so that it copies the correct items when inserted items change the indices of items still to be copied.
Made the target strings and info numbers match between ddblistview and ddbtabstrip so that drags from the playlist to a tab work.

### DIFF
--- a/playlist.c
+++ b/playlist.c
@@ -3508,7 +3508,7 @@ void
 plt_copy_items (playlist_t *to, int iter, playlist_t *from, playItem_t *before, uint32_t *indices, int cnt) {
     pl_lock ();
 
-    if (!from || !to | cnt == 0) {
+    if (!from || !to || cnt == 0) {
         pl_unlock ();
         return;
     }
--- a/playlist.c
+++ b/playlist.c
@@ -115,7 +115,7 @@ static uintptr_t mutex;
 // used at startup to prevent crashes
 static playlist_t dummy_playlist = {
     .refc = 1
-}; 
+};
 
 static int pl_order = -1; // mirrors "playback.order" config variable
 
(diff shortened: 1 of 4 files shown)


playlist.c
```diff
@@ -3513,6 +3513,7 @@ plt_copy_items (playlist_t *to, int iter, playlist_t *from, playItem_t *before,
         return;
     }
 
+    playItem_t *items[cnt];
```
Alexey-Yakovenko: If you try to drag&drop many items in one go, this will cause stack allocation to fail, and items will be 0.
On linux, stack size is usually very large (8M I think it is), while on OSX it's 512K, and on android it's 1M.
So to crash this on OSX, I'd need to call this with 64K items or more.
Please change this to use malloc/free.

Alexey-Yakovenko:
> Changed plt_copy_items() so that it copies the correct items when inserted items change the indices of items still to be copied.

Can you provide an example how to reproduce this problem? I don't get what you're trying to solve.

Alexey-Yakovenko:
Another way of solving this problem without excessive memory allocation is to first form the list of items as a separate list, then insert in the actual playlist at the given point.

Lithopsian:
Use heap.
Also fix the drop position when dragging from one tab to another, so that it isn't moved to beyond any selected items.  That only makes sense for drops within the same tab.

playlist.c
```diff
@@ -3508,30 +3508,34 @@ void
 plt_copy_items (playlist_t *to, int iter, playlist_t *from, playItem_t *before, uint32_t *indices, int cnt) {
     pl_lock ();
 
-    if (!from || !to) {
+    if (!from || !to | cnt == 0) {
```
Alexey-Yakovenko: did you mean `||`?

Lithopsian: Well, it works, but definitely not what I intended ;)

Alexey-Yakovenko:
merged

Issue #1353, opened 2015-11-26
linux static build fails to start with AFS homedir


zao:
On my work workstation the application fails to start up with the following output.

    ∫ /opt/deadbeef-0.6.2/deadbeef 
    starting deadbeef 0.6.2 [static] [portable]
    server_start
    INFO: unlink socket: No such file or directory
    bind: Operation not permitted

I suspect this is due to the home directory being on an OpenAFS filesystem, which is known for being restrictive in what operations it permits. (One famous feature is lack of hardlinks between files in different directories.)

Setting `XDG_CONFIG_HOME` to a local filesystem allows the application to start properly but is a workaround.

`strace` on the binary results in the following output, with a bunch of `ENOENT` and `EPERM` on local sockets created in the config directory.

Maybe session-specific things like local sockets and other per-machine state shall be in local locations like the XDG-suggested `$XDG_RUNTIME_DIR`?

    execve("/opt/deadbeef-0.6.2/deadbeef", ["/opt/deadbeef-0.6.2/deadbeef"], [/* 84 vars */]) = 0
    brk(0)                                  = 0x7eb000
    access("/etc/ld.so.nohwcap", F_OK)      = -1 ENOENT (No such file or directory)
    mmap(NULL, 8192, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7fa269325000
    readlink("/proc/self/exe", "/opt/deadbeef-0.6.2/deadbeef", 4096) = 28
    access("/etc/ld.so.preload", R_OK)      = -1 ENOENT (No such file or directory)
    open("/opt/deadbeef-0.6.2/../lib/tls/x86_64/libdl.so.2", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
    stat("/opt/deadbeef-0.6.2/../lib/tls/x86_64", 0x7ffe84f20e70) = -1 ENOENT (No such file or directory)
    open("/opt/deadbeef-0.6.2/../lib/tls/libdl.so.2", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
    stat("/opt/deadbeef-0.6.2/../lib/tls", 0x7ffe84f20e70) = -1 ENOENT (No such file or directory)
[...]

PR #1355, opened 2015-11-26
Use XDG_RUNTIME_DIR for IPC socket (Bug #1353)

saivert:
Fixes bug #1353

### DIFF
--- a/main.c
+++ b/main.c
@@ -92,6 +92,7 @@ char dbdocdir[PATH_MAX]; // see deadbeef->get_doc_dir
 char dbplugindir[PATH_MAX]; // see deadbeef->get_plugin_dir
 char dbpixmapdir[PATH_MAX]; // see deadbeef->get_pixmap_dir
 char dbcachedir[PATH_MAX];
+char dbruntimedir[PATH_MAX]; // /run/user/<uid>/deadbeef
 
 char use_gui_plugin[100];
 
@@ -403,7 +404,7 @@ server_start (void) {
     int len = offsetof(struct sockaddr_un, sun_path) + sizeof (server_id)-1;
 #else
     char *socketdirenv = getenv ("DDB_SOCKET_DIR");
-    snprintf (srv_local.sun_path, sizeof (srv_local.sun_path), "%s/socket", socketdirenv ? socketdirenv : dbconfdir);
+    snprintf (srv_local.sun_path, sizeof (srv_local.sun_path), "%s/socket", socketdirenv ? socketdirenv : dbruntimedir);
     if (unlink(srv_local.sun_path) < 0) {
         perror ("INFO: unlink socket");
     }
@@ -880,6 +881,19 @@ main (int argc, char *argv[]) {
         fprintf (stderr, "fatal: too long cache path %s\n", dbcachedir);
         return -1;
     }
+
+    const char *xdg_runtime = getenv("XDG_RUNTIME_DIR");
+    if (xdg_runtime)
+    {
+        if (snprintf(dbruntimedir, sizeof (dbruntimedir), "%s/deadbeef/", xdg_runtime) >= sizeof (dbruntimedir)) {
+            fprintf (stderr, "fatal: too long cache path %s\n", dbruntimedir);
+            return -1;
+        }
+        mkdir (dbruntimedir, 0755);
+    }
+    else {
+        strcpy (dbruntimedir, dbconfdir);
+    }
 #endif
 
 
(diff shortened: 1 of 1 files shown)


main.c
```diff
@@ -880,6 +881,17 @@ main (int argc, char *argv[]) {
         fprintf (stderr, "fatal: too long cache path %s\n", dbcachedir);
         return -1;
     }
+
+    const char *xdg_runtime = getenv("XDG_RUNTIME_DIR");
+    if (!xdg_runtime) {
+        fprintf (stderr, "Warning: XDG_RUNTIME_DIR not set falling back to /tmp/deadbeef\n");
```
Alexey-Yakovenko: wouldn't this be printing this warning on pretty much every computer?

main.c
```diff
@@ -880,6 +881,17 @@ main (int argc, char *argv[]) {
...
+    }
+    const char *runtime_root = xdg_runtime ? xdg_runtime : "/tmp";
+    if (snprintf(dbruntimedir, sizeof (dbruntimedir), xdg_runtime ? "%s/deadbeef/" : "%s/deadbeef/", runtime_root) >= sizeof (dbruntimedir)) {
```
Alexey-Yakovenko: this will break when 2 users try to use deadbeef at the same time

main.c
```diff
@@ -880,6 +881,17 @@ main (int argc, char *argv[]) {
         fprintf (stderr, "fatal: too long cache path %s\n", dbcachedir);
         return -1;
     }
+
+    const char *xdg_runtime = getenv("XDG_RUNTIME_DIR");
+    if (!xdg_runtime) {
+        fprintf (stderr, "Warning: XDG_RUNTIME_DIR not set falling back to /tmp/deadbeef\n");
+    }
+    const char *runtime_root = xdg_runtime ? xdg_runtime : "/tmp";
```
Alexey-Yakovenko: this (/tmp/deadbeef) will break when 2 users try to use deadbeef at the same time

main.c
```diff
@@ -92,6 +92,7 @@ char dbdocdir[PATH_MAX]; // see deadbeef->get_doc_dir
 char dbplugindir[PATH_MAX]; // see deadbeef->get_plugin_dir
 char dbpixmapdir[PATH_MAX]; // see deadbeef->get_pixmap_dir
 char dbcachedir[PATH_MAX];
+char dbruntimedir[PATH_MAX]; // /run/user/<uid>/deadbeef
```
Alexey-Yakovenko: this should be a local variable -- it's not needed anywhere else.

Alexey-Yakovenko: (nvm the last comment, deleted)

Alexey-Yakovenko:
what should happen instead:

Check if XDG_RUNTIME_DIR is set, and use it. Otherwise keep the old behavior.

Printing the warning for 99.9% users is annoying, so please don't do that (yeah, I know it's recommended in the FDO paper -- they're wrong).

saivert:
Better?

main.c
```diff
@@ -977,7 +991,7 @@ main (int argc, char *argv[]) {
     len = offsetof(struct sockaddr_un, sun_path) + sizeof (server_id)-1;
 #else
     char *socketdirenv = getenv ("DDB_SOCKET_DIR");
-    snprintf (remote.sun_path, sizeof (remote.sun_path), "%s/socket", socketdirenv ? socketdirenv : dbconfdir);
+    snprintf (remote.sun_path, sizeof (remote.sun_path), "%s/socket", socketdirenv ? socketdirenv : dbruntimedir);
```
Alexey-Yakovenko: looks like the dbruntimedir variable will get used here before being initialized.

Alexey-Yakovenko: I'm wrong, nvm.

Alexey-Yakovenko:
I think we should just use this:
```
--- a/main.c
+++ b/main.c
@@ -402,7 +402,7 @@ server_start (void) {
     memcpy (srv_local.sun_path, server_id, sizeof (server_id));
     int len = offsetof(struct sockaddr_un, sun_path) + sizeof (server_id)-1;
 #else
-    char *socketdirenv = getenv ("DDB_SOCKET_DIR");
+    char *socketdirenv = getenv ("XDG_RUNTIME_DIR");
     snprintf (srv_local.sun_path, sizeof (srv_local.sun_path), "%s/socket", socketdirenv ? socketdirenv : dbconfdir);
     if (unlink(srv_local.sun_path) < 0) {
         perror ("INFO: unlink socket");
@@ -976,7 +976,7 @@ main (int argc, char *argv[]) {
     memcpy (remote.sun_path, server_id, sizeof (server_id));
     len = offsetof(struct sockaddr_un, sun_path) + sizeof (server_id)-1;
 #else
-    char *socketdirenv = getenv ("DDB_SOCKET_DIR");
+    char *socketdirenv = getenv ("XDG_RUNTIME_DIR");
     snprintf (remote.sun_path, sizeof (remote.sun_path), "%s/socket", socketdirenv ? socketdirenv : dbconfdir);
     len = offsetof(struct sockaddr_un, sun_path) + strlen (remote.sun_path);
 #endif
```

saivert:
Yes except it has to create a directory for the socket file otherwise the socket file ends up in XDG_RUNTIME_DIR dir.
Unless you rename it to deadbeefsocket or something. I think it is a lot cleaner with a separate directory for deadbeef though.

Alexey-Yakovenko:
haha yeah that's what I meant by "or similar" after the edit. anyway, let's keep it the way you made it for now.

merged.

PR #1359, opened 2015-11-27
pulseaudio: fix double free in pulse_setformat() 

### LINKED ISSUE #1358: Pulseaudio: crash if two file have different samplerate
Arch Linux 64bit, deadbeef-git in the AUR
deadbeef will get a SIGABRT if the samplerate of two file different

it seems that pulse_setformat() called twice
and pulse_stop() called twice(21a4bda93772f447598594967d346f8be9371458)
then pulse_free() called twice(c43f4189849b1e94040abc63b3e2b1f27afbde91)

pingplug:
fix #1358 
add mutex on pulse_free() and pulse_init()
use a different mutex for pulse_thread()
remove pulse_set_spec() in pulse_setformat() as pulse_set_spec() can be called by pulse_play() and pulse_unpause()
prevent double pulse_stop() cause double pulse_free()

### DIFF
--- a/plugins/pulse/pulse.c
+++ b/plugins/pulse/pulse.c
@@ -53,6 +53,7 @@ static pa_sample_spec ss;
 static ddb_waveformat_t requested_fmt;
 static int state;
 static uintptr_t mutex;
+static uintptr_t mutex_thread;
 
 static int buffer_size;
 
@@ -162,7 +163,7 @@ static int pulse_set_spec(ddb_waveformat_t *fmt)
 static int pulse_init(void)
 {
     trace ("pulse_init\n");
-    state = OUTPUT_STATE_STOPPED;
+    deadbeef->mutex_lock(mutex);
     pulse_terminate = 0;
 
     if (requested_fmt.samplerate != 0) {
@@ -174,6 +175,7 @@ static int pulse_init(void)
     }
 
     pulse_tid = deadbeef->thread_start(pulse_thread, NULL);
+    deadbeef->mutex_unlock(mutex);
 
     return 0;
 }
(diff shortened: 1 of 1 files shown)


plugins/pulse/pulse.c
```diff
@@ -162,7 +163,7 @@ static int pulse_set_spec(ddb_waveformat_t *fmt)
 static int pulse_init(void)
 {
     trace ("pulse_init\n");
-    state = OUTPUT_STATE_STOPPED;
+    deadbeef->mutex_lock(mutex);
```
Alexey-Yakovenko: This would deadlock the mutex if pulse_set_spec returns non-zero.

plugins/pulse/pulse.c
```diff
@@ -308,9 +313,9 @@ static void pulse_thread(void *context)
         pulse_callback (buf, sizeof (buf));
         int error;
 
-        deadbeef->mutex_lock(mutex);
+        deadbeef->mutex_lock(mutex_thread);
         int res = pa_simple_write(s, buf, sizeof (buf), &error);
```
Alexey-Yakovenko: What will happen if e.g. pa_simple_free gets called while this thread is inside of the pa_simple_write?
And the mutex_thread is bogus, because it's only used in one place, which is here -- doesn't make any sense to me.
Explain?

pingplug: It's my fault. I thought the mutex in pulse_thread will cause deadblock.

plugins/pulse/pulse.c
pingplug: So is it right if I unlock before return?

Alexey-Yakovenko: yes

plugins/pulse/pulse.c
Alexey-Yakovenko: I never heard about it causing deadlocks. I think you should remove mutex_thread, and keep using the mutex around pa_simple_write, unless there's the real reason to change this.

pingplug:
OK, fixed

Alexey-Yakovenko:
merged

PR #1366, opened 2015-11-29
Fix for segfault on "refresh coverart"

tvn87:
If the cache folder (e.g. ~/.cache/deadbeef) does not exist a comparison
between size_t and int lead to segmentation fault.


Alexey-Yakovenko:
This fix is too much a hack... I think the code expects at this point, that the folder exists. So the fix would be:

* Make sure that the folder is created before this is called.
* Still handle the error correctly, i.e. `if (scaled_dirs_count < 0) ....`

Alexey-Yakovenko:
Ok, I see what this place in the code is doing. It doesn't need to be creating the folder. It just needs to handle the error correctly. I'll make the fix.

tvn87:
Alright, so you're doing the fix? I would just change the loop condition, so it checks explicitly if `scaled_dirs` is a null pointer.

Alexey-Yakovenko:
should be fixed now.

Issue #1389, opened 2015-12-10
gtkui: listview scrollpos jumps when resizing album art column with the listview scrolled to the end


cboxdoerfer:
Steps to reproduce:
1. Use a rather big playlist with some groups to trigger the bug more easily
2. Scroll to the end of the playlist
3. Change the size of the album art column

While making the album art column smaller and bigger the scroll position of the playlist jumps around.
I tracked it down and this bug was introduced with this patch: 9868fa5b855db3c482ca0288fe92fb4eed519309

cboxdoerfer:
Also fixed with https://github.com/Alexey-Yakovenko/deadbeef/commit/d5cd956cfddbae4900df22d202c98ca99a9632b3.

PR #1393, opened 2015-12-11
gtkui: fix jerky scrolling when resizing at the end of the listview

Lithopsian:
This is to fix issue #1389.


Lithopsian:
Just pushed new commits and they've piled on to this PR.  Let me know if this complicates things too much and I'll un-push them.

The new commits are a fix for issue #1394 and making the edit and remove column items on the header context menu insensitive when not over a column.

PR #1399, opened 2015-12-15
gtkui: (re)fix trkproperties to allow the deletion of fields longer than 4,999 characters

Lithopsian:
I think this was done once, but it isn't working now.

### DIFF
--- a/plugins/gtkui/trkproperties.c
+++ b/plugins/gtkui/trkproperties.c
@@ -266,7 +266,7 @@ on_metadata_edited (GtkCellRendererText *renderer, gchar *path, gchar *new_text,
     gtk_tree_model_get_value (GTK_TREE_MODEL (store), &iter, 3, &mult);
     const char *svalue = g_value_get_string (&value);
     int imult = g_value_get_int (&mult);
-    if (strcmp (svalue, new_text) && (!imult || strlen (new_text))) {
+    if (strcmp (svalue, new_text) && (!imult || strlen (new_text) == 0)) {
         gtk_list_store_set (store, &iter, 1, new_text, 3, 0, -1);
         trkproperties_modified = 1;
     }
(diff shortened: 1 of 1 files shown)


Alexey-Yakovenko:
I don't remember merging this before.. what makes you think it happened?

plugins/gtkui/trkproperties.c
```diff
@@ -266,7 +266,7 @@ on_metadata_edited (GtkCellRendererText *renderer, gchar *path, gchar *new_text,
     gtk_tree_model_get_value (GTK_TREE_MODEL (store), &iter, 3, &mult);
     const char *svalue = g_value_get_string (&value);
     int imult = g_value_get_int (&mult);
-    if (strcmp (svalue, new_text) && (!imult || strlen (new_text))) {
+    if (strcmp (svalue, new_text) && (!imult || strlen (new_text) == 0)) {
```
Alexey-Yakovenko: This looks like a serious bug fix.. how to trigger the bug? I'm trying to figure out what's changed.. I will need to apply this fix to cocoaui too, I think (if it's valid)

I'd like to reproduce this, and be able to verify the fix.

Lithopsian:
The bug is that once you have, or create, a field with more than 5,000 characters, you can't get rid of it.  Minor editing is prevented because the result might be a corruption, but removing the contents entirely should be possible so you can start again.

There was a commit a long time ago that created the line you highlight.  It was intended to allow removal of these fields.  I don't know if it worked then, but it doesn't work now so I make this change.

Lithopsian:
@Alexey-Yakovenko is this whole workaround still needed?  If it is, is a tf field with more than 5,000 characters going to be a problem?  For example, a custom listview column for a long tag.

Alexey-Yakovenko:
It's a workaround for GtkTreeView, and it's still needed. It crashes on some gtk versions when text is longer than certain amount of bytes or chars, far more than 5K though.

DdbListView is unaffected, therefore the workaround is not necessary there.

PR #1400, opened 2015-12-15
playlist: only clear existing selection in plt_search_process2 if requested

Lithopsian:
Selected search items are also cleared inside plt_search_reset_int(), so pass the flag in there.


Alexey-Yakovenko:
oops my bad, sorry man.

PR #1408, opened 2015-12-18
converter: add preview of output files to gtk ui

cboxdoerfer:
Screenshot:

### DIFF
--- a/plugins/converter/convgui.c
+++ b/plugins/converter/convgui.c
@@ -77,6 +77,8 @@ enum {
     PRESET_TYPE_DSP
 };
 
+static const char *default_format = "%album artist% - %title%";
+
 static void
 fill_presets (GtkListStore *mdl, ddb_preset_t *head, int type) {
     ddb_preset_t *p = head;
@@ -99,6 +101,14 @@ on_converter_progress_cancel (GtkDialog *dialog, gint response_id, gpointer user
     converter_ctx_t *ctx = user_data;
     ctx->cancelled = 1;
 }
+
+void
+on_output_file_changed                 (GtkEntry        *entry,
+                                        gpointer         user_data);
+
+void
+on_converter_realize                 (GtkWidget        *widget,
+                                        gpointer         user_data);
 
 typedef struct {
     GtkWidget *entry;
@@ -266,7 +276,7 @@ converter_process (converter_ctx_t *conv)
     conv->outfolder = strdup (gtk_entry_get_text (GTK_ENTRY (lookup_widget (conv->converter, "output_folder"))));
     const char *outfile = gtk_entry_get_text (GTK_ENTRY (lookup_widget (conv->converter, "output_file")));
     if (outfile[0] == 0) {
-        outfile = "%album artist% - %title%";
+        outfile = default_format;
     }
     conv->outfile = strdup (outfile);
     conv->preserve_folder_structure = gtk_toggle_button_get_active (GTK_TOGGLE_BUTTON (lookup_widget (conv->converter, "preserve_folders")));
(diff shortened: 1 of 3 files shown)


cboxdoerfer:
Close PR for now because of huge memory usage when preview list has lots of items

Alexey-Yakovenko:
I think capping the list at ~20 items or so would make sense.

cboxdoerfer:
Well, the problem only happens with GTK3. With GTK2 everything is fine and there's no major memory impact with ~10.000 tracks in the preview. With Adwaita and GTK3 the memory usage however increases to 300-400 MB and with other themes I even get:

```Gdk-ERROR **: The program 'deadbeef' received an X Window System error.
This probably reflects a bug in the program.
The error was 'BadAlloc (insufficient resources for operation)'.
```

So I guess there's probably something wrong with my code, or GTK3 is flawed or buggy.

Alexey-Yakovenko:
GTK list view was never suitable for working with large arrays of data.

cboxdoerfer:
Yeah, unfortunately. For now I limited the number of items. But I guess there's really something wrong. I tried it with one of the gtk example applications from the gtk devs and the same happens there. Up to about 1000 rows everything is fine, with no significant memory usage, but anything larger will result in BadAlloc in gdk. Will talk to the gtk devs about that.

Alexey-Yakovenko:
I don't like the interface.

I think it will look cleaner if you remove the frame with the Preview text, and put "Preview" into the column header, instead of the "Filename".

Horizontal scrollbar is missing, or the column doesn't auto-expand to the content, so when text is long -- no way to see its tail, except by expanding the window itself (which is not always possible).

The window width is larger than 800px now, and can't be reduced -- an easy fix, so should be done.

plugins/converter/convgui.c
```diff
@@ -342,6 +350,121 @@ converter_process (converter_ctx_t *conv)
     return 0;
 }
 
+static int preview_delay_timer = 0;
+
+static gboolean
+update_preview (gpointer user_data)
```
Alexey-Yakovenko: It's better to name this "preview_update", as this is the most common notation in deadbeef sources (object_verb_subject).

Alexey-Yakovenko: It's even better to suffix this with a "_cb", as this the common way to mark callback functions.

plugins/converter/convgui.c
```diff
@@ -342,6 +350,121 @@ converter_process (converter_ctx_t *conv)
...
+    }
+
+    const char *format = user_data;
+    if (!format) {
+        // get filename format
+        GtkEntry *output_file = GTK_ENTRY (lookup_widget (conv->converter, "output_file"));
+        format = gtk_entry_get_text (output_file);
+        if (!format || format[0] == 0) {
+            // use default format when entry is empty
+            format = "%album artist% - %title%";
```
Alexey-Yakovenko: It would make sense to put it into some header file, because now it's used from several places.

plugins/converter/convgui.c
```diff
@@ -342,6 +350,121 @@ converter_process (converter_ctx_t *conv)
...
+    char *tf = deadbeef->tf_compile (format);
+    if (!tf) {
+        return FALSE;
+    }
+
+    // detach model from treeview while adding lots of items to increase performance
+    g_object_ref (store);
+    gtk_tree_view_set_model (tree_view, NULL);
+
+    // FIXME: limit number of items because of huge memory usage with GTK3
```
Alexey-Yakovenko: This should be a plain comment, or a NOTE, but not a FIXME.

cboxdoerfer:
Found the bug with the memory usage. It was caused by a widget placement inside the scrolled window. Happend probably when copying/pasting the wrong widget in glade. So memory usage is no problem anymore, even with 50.000 tracks.

And of course I will fix the issues you mentioned.

plugins/converter/convgui.c
```diff
@@ -342,6 +350,121 @@ converter_process (converter_ctx_t *conv)
...
+    }
+    // reattach model to treeview
+    gtk_tree_view_set_model (tree_view, GTK_TREE_MODEL (store));
+    g_object_unref (store);
+    deadbeef->tf_free (tf);
+    return FALSE;
+}
+
+static void
+add_preview_timeout (const char *format)
```
Alexey-Yakovenko: same here, should be called "preview_timeout_add"

Alexey-Yakovenko:
> So memory usage is no problem anymore, even with 50.000 tracks.

what about performance? does it render only the visible tracks, not all of them I hope?

cboxdoerfer:
Performance is good. Scrolling is smooth, both in GTK2 and GTK3 and memory usage is barely noticeable (mostly depending on the string lengths).

PR #1410, opened 2015-12-20
gtkui: make DB_EV_FOCUS_SELECTION work properly again, fix some cursor and scroll edge cases

Lithopsian:
I gave up trying to hack everything onto the DB_EV_FOCUS_SELECTION message.  Now this does what it used to, move the cursor and scroll to the first selected item in the current playlist in all listviews.  There is a new structured message for precisely setting the cursor to synchronise between the main and search listviews.  The first selected item is not always the one that should receive the cursor.

Playback following should now work in all cases, including in a background tab where playback stops, moves to a different tab, or the preferences are changed, and jump to track where there is no main playlist widget.  Playback following is done in gtkui.c for background tabs and if there is no listview widget to do it.  Otherwise all messages are acted on independently by the main and search listviews, the only synchronisation is for GUI actions on one or the other.

Playback following no longer occurs in the current playlist if it is being selected or dragged, basically the whole time between the left mouse button being pressed on a listview and released.

### DIFF
--- a/deadbeef.h
+++ b/deadbeef.h
@@ -428,6 +428,10 @@ enum {
     DB_EV_TRACKFOCUSCURRENT = 1006, // user wants to highlight/find the current playing track
 #endif
 
+#if (DDB_API_LEVEL >= 9)
+    DB_EV_CURSOR_MOVED = 1007, // for synchronising listviews, p1 = playlist iter sending the message
+#endif
+
     DB_EV_MAX
 };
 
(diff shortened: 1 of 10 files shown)


deadbeef.h
```diff
@@ -428,6 +428,10 @@ enum {
     DB_EV_TRACKFOCUSCURRENT = 1006, // user wants to highlight/find the current playing track
 #endif
 
+#if (DDB_API_LEVEL >= 9)
+    DB_EV_CURSOR_MOVED = 1007, // for synchronising listviews, p1 = playlist iter sending the message
```
Alexey-Yakovenko: Please change the comment to this:
`used for syncing playlists, p1 = PL_MAIN or PL_SEARCH, ctx is a ddb_event_track_t, containing the new track under cursor`

Alexey-Yakovenko:
nvm, fixed it myself and merged.

PR #1417, opened 2015-12-21
gtkui: just draw coverart once at song change

Lithopsian:


Alexey-Yakovenko:
what happens if you just press "stop"? will "songchanged" be emitted anyway?

Lithopsian:
Yes, there is always a DB_EV_SONGCHANGED event, with either DB_EV_SONGSTARTED, DB_EV_SONGFINISHED, or both.  In fact two DB_EV_SONGFINISHED events when playback moves itself from one track to another, a few seconds apart.

Alexey-Yakovenko:
that's kinda strange.., but ok.

Issue #1381, opened 2015-12-08
Playlist group titles with a leading space are often incorrectly ellipsized


Lithopsian:
1. Set custom styling with a group header font of Courier 10 Pitch Italic.
2. Group by a custom format with a leading space.  This can be a hard-coded string, including an empty string, or a format such as %length% which will be formatted with a leading space.

The title is shown ellipsized, which is never intended to happen for the group title.  The special case of the group by string " - ", caused by missing metadata, seems to display just an ellipsis in all fonts, with the group title line over the top of it.  This doesn't always happen.  It is more common with italic fonts.  Short strings such as the " - " example are wrong in almost every font.

draw_get_text_extents() in gdkdrawing.c appears to returning only the ink extents, not the logical extents.  This means that leading and trailing whitespace is ignored.  For example, the ink extents for " + " are 7x7 pixels, while the logical extents are 29x15 pixels.  If this is the desired behaviour, then whitespace must be stripped from the group header before it is rendered.  I think this is probably not desirable for alignment reasons.

Note that custom formats have leading whitespace stripped, but only when they are saved.  This means that a format with leading whitespace will have it removed the next time Deadbeef starts.

Lithopsian:
There is a similar problem in tab titles.  Leading, or trailing, whitespace ellipsizes the title instead of making the tab wider.

PR #1422, opened 2015-12-22
Text drawing changes

Lithopsian:
This is to fix issue #1381.  The extent of text was incorrectly calculated, especially with leading or trailing whitespace.

It also preserves leading and trailing whitespace in group formats  At the moment, it is saved for a while, then lost on restart.

### DIFF
--- a/plugins/gtkui/ddblistview.c
+++ b/plugins/gtkui/ddblistview.c
@@ -120,7 +120,7 @@ ddb_listview_list_render_row_background (DdbListview *ps, cairo_t *cr, DdbListvi
 static void
 ddb_listview_list_render_row_foreground (DdbListview *ps, cairo_t *cr, DdbListviewIter it, int idx, int y, int w, int h, int x1, int x2);
 static void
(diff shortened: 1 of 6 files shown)


Alexey-Yakovenko:
I don't get the commit named `gtkui: preserve leading and trailing whitespace in group formats`.

The way it looks is that you take the group format string, escape it, put in quotes, and send to groups_changed.

Then before using, you unescape it in the pl_common_set_group_format.

And there's no single line of explanation why is this needed.

Regarding the code itself, I reviewed the commits individually, and [this commit](https://github.com/Lithopsian/deadbeef/commit/32fc3b0467dab0aec12d8c366e23910748e291ac) needs some more work.

Alexey-Yakovenko:
@Lithopsian 
I merged this PR because it fixes a bug, but now I'm trying to clean out the changes from it which don't seem necessary, unless they can be explained.

In particular, I can't see any reason to use the parser_escape/unescape stuff.

Can you explain what it's used for?

PR #1434, opened 2016-01-02
Fix useless warnings for non-existing menus.

saivert:
No need to use the generic lookup_widget() function. We know what we are dealing with
so can do the right thing.


Alexey-Yakovenko:
merged

PR #1441, opened 2016-01-07
hotkeys: reload metadata from playlist tabs now reloads whole playlist (fixes #1424)

### LINKED ISSUE #1424: Reload metadata in playlist context menu only reloads metadata for one track
Namely the one selected in the playlist. That would be fine if the user opens the context menu on one of the selected tracks, but in this case it's done on the selected playlist entry, so the expected outcome would be to update the _whole_ playlist. Otherwise the entry is kind of useless...

spikeh:
Hello, I made a small change to fix bug #1424. I kept the changes to a minimum, but I think this whole function can be refactored to handle the different contexts better.


Alexey-Yakovenko:
merged

PR #1448, opened 2016-01-12
osx: Removing define for '/usr/local/bin/' encoder path in OSX.

carlosanunes:
This removes the define for '/usr/local/bin/' encoder path in OSX. 
Execution of encoders now relies only in the PATH env variable when no full path is given (like in Linux). Set PATH variable value in LSEnvironment,deadbeef-Info.plist to '/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin' (if no value is set, the default is '/usr/bin:/bin:/usr/sbin:/sbin'). This is only applicable when launched through Launch Services. If executed directly from the command line the env variables of the shell are used and the LSEnvironmnt dictionary is ignored.
This last part is also applicable to Xcode. When running in Xcode the path variable must be set in the schema preferences.

https://developer.apple.com/library/mac/documentation/General/Reference/InfoPlistKeyReference/Articles/LaunchServicesKeys.html#//apple_ref/doc/uid/20001431-106825

### DIFF
--- a/plugins/cocoaui/deadbeef-Info.plist
+++ b/plugins/cocoaui/deadbeef-Info.plist
@@ -2,6 +2,11 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>LSEnvironment</key>
+	<dict>
+		<key>PATH</key>
+		<string>/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin</string>
+	</dict>
 	<key>CFBundleDevelopmentRegion</key>
 	<string>en</string>
 	<key>CFBundleExecutable</key>
--- a/plugins/converter/converter.c
+++ b/plugins/converter/converter.c
@@ -928,11 +928,6 @@ convert (DB_playItem_t *it, const char *out, int output_bps, int output_is_float
             char *o = enc;
             *o = 0;
 
-#ifdef __APPLE__
-            strcpy (o, "/usr/local/bin/");
-            o += 15;
-#endif
-
             int len = sizeof (enc);
             while (e && *e) {
                 if (len <= 0) {


Alexey-Yakovenko:
I have to remove this change, because it doesn't work on my computer. Any ideas why this could be?

carlosanunes:
Sure it works, you're hardcoding the path. :p 

When you were launching it from Xcode how did you configure the schema ?

A bit more strange that when you launch thru finder, but will check.

Alexey-Yakovenko:
ok, I wasn't setting the PATH in xcode, that's probably the reason. but doesn't explain why it doesn't work when running from Dock. I'll try again, just to be sure.

Alexey-Yakovenko:
Tried again. Running from Dock still doesn't work (with the fix applied).

carlosanunes:
What is the OSX version you're using ?

Alexey-Yakovenko:
The latest 10.11

Alexey-Yakovenko:
can try on 10.9 now if needed

carlosanunes:
No need.

(just to confirm, after you changed the schema in XCode, it worked ?)

Alexey-Yakovenko:
I haven't tried this. Will try now, but have only access to 10.8 atm.

Alexey-Yakovenko:
Yes, it works after adding `PATH=$PATH:/usr/loca/bin` to the scheme

Alexey-Yakovenko:
Weird, here it also works when launched from the Dock.

carlosanunes:
With my patch ?

Alexey-Yakovenko:
yes

carlosanunes:
I am unable to reproduce the behaviour in OSX version 10.11.3. :(

This is not working for any executable in /usr/local/bin but works for encoders in a different location ?

Alexey-Yakovenko:
I will try to investigate, later today, or perhaps tomorrow.

Alexey-Yakovenko:
Ok, I figured what was going on. Indeed it worked, but the destination file already existed, and the error message box is not implemented yet, so it looked like converter "skips" the conversion for unknown reason, which I thought to be the missing encoder error.

I restored the commit. Really sorry for the confusion.

Alexey-Yakovenko:
Even worse, the error message box actually works. I just had my converter settings set to When file exists: Skip. Feel so embarassed now...

carlosanunes:
It's all good. :)

Wish all problems would be solved by changing a dropdown option.

PR #1453, opened 2016-01-21
Work on cocoaui: File association in finder and converter panel filename preview

carlosanunes:
Deadbeef is now associated with the filetypes it can decode (for the current osx build) in finder. That is user can open a file with Deadbeef from finder;
In the converter panel of deadbeef the filename preview field is filled.

### DIFF
--- a/main.c
+++ b/main.c
@@ -351,58 +351,74 @@ server_exec_command_line (const char *cmdline, int len, char *sendback, int sbsi
         parg++;
     }
     if (parg < pend) {
-        if (conf_get_int ("cli_add_to_specific_playlist", 1)) {
-            char str[200];
-            conf_get_str ("cli_add_playlist_name", "Default", str, sizeof (str));
-            int idx = plt_find (str);
-            if (idx < 0) {
-                idx = plt_add (plt_get_count (), str);
-            }
-            if (idx >= 0) {
-                plt_set_curr_idx (idx);
-            }
+        if (add_paths(parg, pend - parg, queue, sendback, sbsize) > 0) {
+            return 0; // files not loaded, but continue normally
         }
-        playlist_t *curr_plt = plt_get_curr ();
-        if (plt_add_files_begin (curr_plt, 0) != 0) {
-            plt_unref (curr_plt);
-            snprintf (sendback, sbsize, "it's not allowed to add files to playlist right now, because another file adding operation is in progress. please try again later.");
-            return 0;
-        }
-        // add files
         if (!queue) {
-            plt_clear (curr_plt);
-            messagepump_push (DB_EV_PLAYLISTCHANGED, 0, DDB_PLAYLIST_CHANGE_CONTENT, 0);
-            plt_reset_cursor (curr_plt);
-        }
-        while (parg < pend) {
-            char resolved[PATH_MAX];
-            const char *pname;
-            if (realpath (parg, resolved)) {
-                pname = resolved;
-            }
-            else {
-                pname = parg;
-            }
-            if (deadbeef->plt_add_dir2 (0, (ddb_playlist_t*)curr_plt, pname, NULL, NULL) < 0) {
-                if (deadbeef->plt_add_file2 (0, (ddb_playlist_t*)curr_plt, pname, NULL, NULL) < 0) {
-                    int ab = 0;
-                    playItem_t *it = plt_load2 (0, curr_plt, NULL, pname, &ab, NULL, NULL);
-                    if (!it) {
-                        fprintf (stderr, "failed to add file or folder %s\n", pname);
-                    }
-                }
-            }
-            parg += strlen (parg);
-            parg++;
-        }
-        pl_save_current ();
-        messagepump_push (DB_EV_PLAYLISTCHANGED, 0, DDB_PLAYLIST_CHANGE_CONTENT, 0);
-        plt_add_files_end (curr_plt, 0);
-        plt_unref (curr_plt);
-        if (!queue) {
-            messagepump_push (DB_EV_PLAY_NUM, 0, 0, 0);
             return 2; // don't reload playlist at startup
         }
+    }
+    return 0;
+}
+
+// parses a list of paths and adds them to deadbeef
+// 0 - no error, files loaded
+// 1 - no error, but files not loaded
+int
+add_paths(const char *paths, int len, int queue, char *sendback, int sbsize) {
+    const uint8_t *parg = (const uint8_t *)paths;
+    const uint8_t *pend = paths + len;
+
+    if (conf_get_int ("cli_add_to_specific_playlist", 1)) {
+        char str[200];
+        conf_get_str ("cli_add_playlist_name", "Default", str, sizeof (str));
+        int idx = plt_find (str);
+        if (idx < 0) {
+            idx = plt_add (plt_get_count (), str);
+        }
+        if (idx >= 0) {
+            plt_set_curr_idx (idx);
+        }
+    }
+    playlist_t *curr_plt = plt_get_curr ();
+    if (plt_add_files_begin (curr_plt, 0) != 0) {
+        plt_unref (curr_plt);
+        snprintf (sendback, sbsize, "it's not allowed to add files to playlist right now, because another file adding operation is in progress. please try again later.");
+        return 1;
+    }
+    // add files
+    if (!queue) {
+        plt_clear (curr_plt);
+        messagepump_push (DB_EV_PLAYLISTCHANGED, 0, DDB_PLAYLIST_CHANGE_CONTENT, 0);
+        plt_reset_cursor (curr_plt);
+    }
+    while (parg < pend) {
+        char resolved[PATH_MAX];
+        const char *pname;
+        if (realpath (parg, resolved)) {
+            pname = resolved;
+        }
+        else {
+            pname = parg;
+        }
+        if (deadbeef->plt_add_dir2 (0, (ddb_playlist_t*)curr_plt, pname, NULL, NULL) < 0) {
+            if (deadbeef->plt_add_file2 (0, (ddb_playlist_t*)curr_plt, pname, NULL, NULL) < 0) {
+                int ab = 0;
+                playItem_t *it = plt_load2 (0, curr_plt, NULL, pname, &ab, NULL, NULL);
+                if (!it) {
+                    fprintf (stderr, "failed to add file or folder %s\n", pname);
+                }
+            }
+        }
+        parg += strlen (parg);
+        parg++;
+    }
+    pl_save_current ();
+    messagepump_push (DB_EV_PLAYLISTCHANGED, 0, DDB_PLAYLIST_CHANGE_CONTENT, 0);
+    plt_add_files_end (curr_plt, 0);
+    plt_unref (curr_plt);
+    if (!queue) {
+        messagepump_push (DB_EV_PLAY_NUM, 0, 0, 0);
     }
     return 0;
 }
--- a/plugins/cocoaui/AppDelegate.m
+++ b/plugins/cocoaui/AppDelegate.m
@@ -207,7 +207,8 @@ static int file_added (ddb_fileadd_data_t *data, void *user_data) {
     dispatch_queue_t aQueue = dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0);
     dispatch_async(aQueue, ^{
         char str[100];
-        NSString *paths =[ filenames componentsJoinedByString:@" "];
+        // building single paths string for the deadbeef function, paths must be separated by '\0'
+        NSString *paths =[filenames componentsJoinedByString:@"\0"];
         add_paths([paths UTF8String], [paths length], 0, str, 100);
     });
 }
--- a/plugins/cocoaui/AppDelegate.m
+++ b/plugins/cocoaui/AppDelegate.m
@@ -194,32 +194,22 @@ static int file_added (ddb_fileadd_data_t *data, void *user_data) {
 }
 
 - (BOOL)application:(NSApplication *)theApplication openFile:(NSString *)filename {
-    NSString *playlist = nil;
-    NSURL *url = [NSURL fileURLWithPath:filename];
-    // add to specific list if property set
-    if (conf_get_int ("cli_add_to_specific_playlist", 1)) {
-        char str[200];
-        conf_get_str ("cli_add_playlist_name", "Default", str, sizeof (str));
-        playlist = [NSString stringWithUTF8String:str];
-    }
-    [self addFiles:YES play:YES files:[NSArray arrayWithObject:url] playlist:nil];
+    dispatch_queue_t aQueue = dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0);
+    dispatch_async(aQueue, ^{
+        char str[100];
+        add_paths([filename UTF8String], [filename length], 0, str, 100);
+    });
     return YES; // assume that everything went ok
 }
 
 
 - (void)application:(NSApplication *)sender openFiles:(NSArray *)filenames {
-    NSArray *urls = [NSArray array];
-    NSString *playlist = nil;
-    for (id filename in filenames) {
-        urls = [urls arrayByAddingObject:[NSURL fileURLWithPath:filename]];
-    }
-    // add to specific list if property set
-    if (conf_get_int ("cli_add_to_specific_playlist", 1)) {
-        char str[200];
-        conf_get_str ("cli_add_playlist_name", "Default", str, sizeof (str));
-        playlist = [NSString stringWithUTF8String:str];
-    }
-    [self addFiles:YES play:YES files:urls playlist:playlist];
+    dispatch_queue_t aQueue = dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0);
+    dispatch_async(aQueue, ^{
+        char str[100];
+        NSString *paths =[ filenames componentsJoinedByString:@" "];
+        add_paths([paths UTF8String], [paths length], 0, str, 100);
+    });
 }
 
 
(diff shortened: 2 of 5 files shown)


Alexey-Yakovenko:
It's not the right way to do it... This will always add files to the current playlist, disregarding the settings, e.g. the `cli_add_playlist_name` and `cli_add_to_specific_playlist`.

We should extract the implementation of this from main.c server_exec_command_line into a utility function.

It might also be necessary to hook the file addition progress dialog the same way as GTKUI does it (I can't remember if this is already done or not, but most likely it's not done).

The other one, about converted filenames preview, is not quite right either.
We should keep stuff as much close as possible between GTKUI and CocoaUI.
You should see how this preview works in GTKUI on master.
Also you used old title formatting string with new title formatting API -- this won't work correctly.

Regarding the hardcoded filetypes.. I'm not sure whether it's a good or bad thing to do..
Will try it out. The obviously bad thing is that adding new formats would be a pain, especially with external plugins, like opus.

carlosanunes:
"It's not the right way to do it... This will always add files to the current playlist, disregarding the settings, e.g. the cli_add_playlist_name and cli_add_to_specific_playlist. (...)"

Will comply.

"The other one, about converted filenames preview, is not quite right either.
We should keep stuff as much close as possible between GTKUI and CocoaUI.
You should see how this preview works in GTKUI on master.
Also you used old title formatting string with new title formatting API -- this won't work correctly."

Looked at the okAction() function in ConverterWindowController.m to get the title formatting string. But will redo looking at the gtkui.

"Regarding the hardcoded filetypes.. I'm not sure whether it's a good or bad thing to do..
Will try it out. The obviously bad thing is that adding new formats would be a pain, especially with external plugins, like opus."

I wrote a bash script to help me with generating the xml for the info.plist but it is still not quite right. That is, it still leaves a bit of work for the dev.
I thought up of other ways to lessen the work, but always defining the filetypes in the info.plist file. OSX looks too hardwired to expect this in order for a clean circumvent - rewriting the info.plist at runtime is completely unrecommended, and APIs that would allow a app to edit the application/file association OSX registry are deprecated.
There could be ways to later on package external plugins with their own info.plist... but would need to do some research on this.
The other alternative would be to just make deadbeef associated to every public file, leaving for the user the task of knowing if it should use deadbeef to open it or not... :/

Alexey-Yakovenko:
> APIs that would allow a app to edit the application/file association OSX registry are deprecated.

Which APIs are those? Usually when something is deprecated, there should be a replacement API.

> The other alternative would be to just make deadbeef associated to every public file, leaving for the user the task of knowing if it should use deadbeef to open it or not... :/

Hehe.. no, don't do this :)

Alexey-Yakovenko:
If it's not possible to do in any other way, we'll just use the static plist, yeah.. We do the same on linux and android anyway :)

Alexey-Yakovenko:
> Looked at the okAction() function in ConverterWindowController.m to get the title formatting string.

It's out of date. Needs fixing as well.

carlosanunes:
https://developer.apple.com/library/mac/documentation/Carbon/Reference/LaunchServicesReference/ - Scratch that, I am not so sure now that even with the deprecated API there could be an easy way to register file extensions at runtime.

Alexey-Yakovenko:
This doesn't look deprecated to me.

Alexey-Yakovenko:
All these APIs are not deprecated:

```
LSCopyAllRoleHandlersForContentType
LSCopyDefaultRoleHandlerForContentType
LSSetDefaultRoleHandlerForContentType
LSCopyAllHandlersForURLScheme
LSCopyDefaultHandlerForURLScheme
LSSetDefaultHandlerForURLScheme```

Are they insufficient?

Alexey-Yakovenko:
anyway.. I haven't tried too hard, but there are several solutions on google which I'd try before sticking with a static plist.

carlosanunes:
> Are they insufficient?

I can't see a way to register filetypes with them. 

> I haven't tried too hard, but there are several solutions on google

Can you paste some of them here ?

Alexey-Yakovenko:
This discussion looked the most promising to me: http://stackoverflow.com/questions/3344157/setting-default-application-for-given-file-extension-on-mac-os-x-from-code

(I don't know if it would work)

carlosanunes:
Hmm... was missing the https://developer.apple.com/library/mac/documentation/MobileCoreServices/Reference/UTTypeRef/ part.

Could work... the bad part with that method is that deadbeef would always be marked as the default application.

carlosanunes:
Later tonight I'll make the first two changes you requested. But i'll leave the file association as it is for now.

Alexey-Yakovenko:
Yes, the file association needs to be researched.

carlosanunes:
Pushed the changes.

Alexey-Yakovenko:
I don't like the way opening files is implemented.

* It duplicates the code, and does things differently than what's in the main.c / GTKUI - we should reuse existing code instead of making new code which is doing exactly the same thing but differently.
* It's not memory-efficient (duplicates strings in memory for no real reason)
* It has what looks like bugs, e.g. it gets the playlist name from config, converts into NSString, then doesn't use it

Alexey-Yakovenko:
Converter changes would probably have been accepted, if they were in a separate PR.

carlosanunes:
> It has what looks like bugs, e.g. it gets the playlist name from config, converts into NSString, then doesn't use it

Checked, in the openFile method the bug is there. The openFiles method is clean.

> It duplicates the code, and does things differently than what's in the main.c / GTKUI - we should reuse existing code instead of making new code which is doing exactly the same thing but differently.
> It's not memory-efficient (duplicates strings in memory for no real reason)

Code is/was already duplicated. :p 
The open file and open dir functions are basically the same for example. I, albeit wrongly, just exercised the same margin.
Proposal for revision in the last section. (please give me your feedback on it, and/or counter :) )

I looked at the function you indicated in main.c, but it was built considering cli commands, unless there are different aspects of it that I am missing, most of it has no obvious use in an cocoa application. Or the intention is, for whatever reason, to allow the user to open a terminal, cd to the deadbeef bundle, and use cli commands if he so desires ?
I would understand it better if I could actually use the function in main.c instead of copying it.

Revision taking into account code duplication:
* openFile and openFiles would just call addFiles function (BOOL clear, BOOL play, NSArray files)
* addFiles would read the cli_add_to_specific_playlist and cli_add_playlist_name config and call addToPlaylist function (BOOL clear, BOOL play, NSArray files, NSString playlist)
* addToPlaylist would insert files

Alexey-Yakovenko:
> I would understand it better if I could actually use the function in main.c instead of copying it.

Yes, that's what I meant, and I wrote it in the original feedback.

This:

> We should extract the implementation of this from main.c server_exec_command_line into a utility function.
> It might also be necessary to hook the file addition progress dialog the same way as GTKUI does it (I can't remember if this is already done or not, but most likely it's not done).

basically, I mean that we should make 1 function, and use it from both places, and reuse as much code as possible.

but anyway.. I will accept PR which does things your way, but it needs a bit more work.
I suggest that addToPlaylist function should take `ddb_playlist_t  *` pointer instead of NSString playlist name.

Alexey-Yakovenko:
> Or the intention is, for whatever reason, to allow the user to open a terminal, cd to the deadbeef bundle, and use cli commands if he so desires ?

I'm not sure what do you mean by this question, but yes -- this is supposed to work the same way as in Linux.

carlosanunes:
> basically, I mean that we should make 1 function, and use it from both places, and reuse as much code as possible.

Sorry, I was limiting in my mind any changes just to the cocoaui plugin code. I did not interpret that you were suggesting altering main.c itself.
Completely agree now. Will comply.

carlosanunes:
For review. :)

Alexey-Yakovenko:
It's almost all good, except the path list should be separated with zeroes instead of spaces.

Alexey-Yakovenko:
@carlosanunes I compiled deadbeef off your branch, to try and test how the file associations work. They simply don't. Finder -> Open With doesn't show deadbeef as supported, and even if I'm trying to force it, still shows this:

image

What did I miss?

Alexey-Yakovenko:
Converter preview: works, but I think it should refresh the preview while typing (can be added in another PR).

carlosanunes:
> What did I miss?

screen shot 2016-01-27 at 13 48 06

screen shot 2016-01-27 at 13 54 41

Testing it a bit more thoroughly, I am not getting the same behaviour as you, but I noticed that although I can 'force' file association, Finder is not automagically adding DeadBeef to all the filetypes I defined in the 'Open With...' menu.

Did you drop deadbeef bundle in the Application folder ?

> It's almost all good, except the path list should be separated with zeroes instead of spaces.

Failed to test it with multiple files... Will fix.

Alexey-Yakovenko:
@carlosanunes Yeah I dropped into /Applications.

I even made sure to run deadbeef from /Applications before trying to open with Finder -> Open with...

I will try at home later today on 10.11 (am using 10.9.5 now) -- maybe a compatibility issue..

carlosanunes:
I'll later on check if the plist data is correct, or if some tag is missing... can you confirm if the problem you're having also affects .mp3 files ?

(tested everything in 10.11)

Alexey-Yakovenko:
@carlosanunes yeah, I tested on mp3, flac and ogg -- same thing.

carlosanunes:
Ok, so I did some more testing, can you confirm that the file itself is being correctly identified in Finder ? That is, Finder 'knows' that it is an OggVorbis Audio file ? 

Finder seems to be using a different criteria to populate the file description from the one it uses to populate the Open With list. The custom UTIs that I created might not have all the actual required information or be conflicting with something else.
But still cannot explain why DeadBeef is not recognised as a viable app for .ogg files in your system.

Alexey-Yakovenko:
@carlosanunes tested on 10.11.3:

Finder `Open With` menu still doesn't have deadbeef.
Something is obviously wrong.

But clicking `Open With -> Other...` let's me choose deadbeef, and shows that it can open the files (I tried FLAC).

Should we perhaps look at how this is implemented in VLC? It works correctly in both 10.9 and 10.10.

Alexey-Yakovenko:
> Ok, so I did some more testing, can you confirm that the file itself is being correctly identified in Finder ? 

Will have a look at this tomorrow (but I tried with mp3 and flac as well with the same result, and I'm pretty sure at least mp3 can be identified by Finder)

carlosanunes:
@Alexey-Yakovenko 
Already had looked at it before starting this, and checked it again. VLC uses the deprecated method (CFBundleTypeExtensions)...

https://developer.apple.com/library/mac/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html

Alexey-Yakovenko:
Deprecated is fine if it works better than the new stuff.

carlosanunes:
@Alexey-Yakovenko 
Can you check if Finder 'know' what .dbpl files are ? That is, in the description. 
screen shot 2016-01-27 at 21 41 20

> Deprecated is fine if it works better than the new stuff.

Well... if it was deprecated there is probably a reason for this... I am going to read up on this, see if there is something in the Apple Doc that I missed...

Alexey-Yakovenko:
Yes, will get back to you tomorrow.

Alexey-Yakovenko:
> Can you check if Finder 'knows' what .dbpl files are ? That is, in the description. 

No. It's recognized as a Document, opens with TextEdit. The same in "Open With".

Alexey-Yakovenko:
@carlosanunes Something interesting.. It looks I missed trying mp3 on 10.9 yesterday.

Tried it now, and it works in `Open With -> deadbeef`. I first thought "it fixed itself", but then I see FLAC and OGG still don't work.

carlosanunes:
Very strange... last night I read a bit more on this, but at least we should be getting similar behaviour.

Later tonight I'll do some changes to the plist and if it 'helps' OSX to more rapidly include file association into its db.

carlosanunes:
Checked something obvious... the iTunes Info.plist file. It is a bit different, but still uses the same method as we do.
I am going to change the deadbeef plist to match the iTunes plist.

carlosanunes:
Pushed changes.

screen_shot

Although I removed some keys and added some more descriptors to the UTTypes (for example, the ogg type), my main change was moving all the uttypes I declared to be under the UTExportedTypeDeclarations key, against what the documentation suggests but what iTunes follows (https://developer.apple.com/library/mac/documentation/General/Reference/InfoPlistKeyReference/Articles/CocoaKeys.html#//apple_ref/doc/uid/TP40009251-SW9). OSX seems to ignore UTTypes declared under the UTImportedTypeDeclarations key if a matching one was declared by another application outside of it (the key).

Please let me know if it is working on your side as well.

PS: Please be patient, it takes a while, and make sure there are no multiple deadbeef bundles on the system, as this seems to confuse OSX a bit.

Alexey-Yakovenko:
> Example, if I create a bogus gme audio file (the filetype is completely unknown by any other application in my system), it is immediately associated with Deadbeef.

Tried this -- indeed, this is how it works. NSF files do get associated with deadbeef.
I'll try deleting VLC and see what happens.

Alexey-Yakovenko:
The latest batch of changes didn't fix anything, deadbeef doesn't get picked up at all for WV, but works for NSF (VLC installed).

Alexey-Yakovenko:
(google chrome is using CFBundleTypeExtensions just like VLC)

carlosanunes:
Rebooted as you did... now seeing the same behaviour...

Yeah, I had checked chrome's info.plist as well. What I don't get is that iTunes does not use it and still gets picked up... Going to recheck it again, and add valid icon for the filetype.

Alexey-Yakovenko:
That's probably because iTunes is preinstalled (i.e. before VLC), so it takes priority somehow.

What I don't get is how VLC works, even if installed e.g. after deadbeef.

carlosanunes:
Yes, it would be nice to find out how OSX prioritises this stuff...

carlosanunes:
PS: Haven't tested it, but probably a quick fix, although I suspect you won't be very fond of it, would be to just associate deadbeef with the public.audio 'supertype' like QuickTime does...

Alexey-Yakovenko:
My preferred solution would be the one that Chrome and VLC are using. Because this seems to work.

carlosanunes:
Was afraid of that answer...

I am not too comfortable in using deprecated stuff... Maybe hopeless but I'll try to get this to work... again... iTunes using it and having no problem with it bugs me...

In the mean time, I can make separate PR for the other stuff.

Alexey-Yakovenko:
Everything becomes deprecated eventually. Sometimes deprecated is the only way to go.

carlosanunes:
That is true...

Please indulge me a bit longer... At end of the week I'll have time to check this more thoroughly.

Alexey-Yakovenko:
No problem mate, take your time.

carlosanunes:
Removed more keys that were not necessary were otherwise already ignored, and fixed some identifier mismatch issues.

Tested in two different OSX systems, rebooted. Uninstalled VLC and installed VLC, rebooted. Seems to be working now...

Alexey-Yakovenko:
Tried the latest one. Nothing has changed. Deadbeef doesn't appear as an app which can open FLAC files. Perhaps I'm doing something wrong?

I build in XCode, then drag-n-drop deadbeef.app to /Applications from XCode.

Then I right click various files, and I can see that deadbeef can open some file formats, like NSF. But still can't open e.g. .MOD, or FLAC, or APE.

However it can open MP3, but I remember you mentioned that "existing" file types work differently than the new ones.

Alexey-Yakovenko:
The only way to get this PR merged is if it works on my machine without removing VLC or doing any other manual tweaks.

carlosanunes:
>Tried the latest one (OSX 10.9.5). Nothing has changed. Deadbeef doesn't appear as an app which >can open FLAC files. Perhaps I'm doing something wrong?

>I build in XCode, then drag-n-drop deadbeef.app to /Applications from XCode.
>After that I run deadbeef once, then logout, login, and run deadbeef again.
>Is that enough to make OSX register the new data?

You don't need to copy deadbeef.app, actually it's probably a good idea that you do not. In OSX there is a service (lsregister I think) that searches the filesystem for new app bundles, after finding one it goes through the data in the plist.info file and adds it to the system db.

>Then I right click various files, and I can see that deadbeef can open some file formats, like NSF. But >still can't open e.g. .MOD, or FLAC, or APE.

>However it can open MP3, but I remember you mentioned that "existing" file types work differently than >the new ones.

File association in OSX works by UTIs (unique type identifiers). There is an already a set of system declared UTI ( https://developer.apple.com/library/mac/documentation/Miscellaneous/Reference/UTIRef/Articles/System-DeclaredUniformTypeIdentifiers.html ), MP3 for example is one of them. So for MP3, I only state that deadbeef can handle files with that UTI and that is it.
The problem here seems to lie with UTIs (for example the one I wrote for FLAC) that I declared in the plist.info file that somehow conflict with UTIs defined by other applications, for example VLC.

I am working with OSX v10.11.3. Repeated the test (uninstalled and installed VLC) and still works...

carlosanunes:
If you can, please test again but by not copying this time... just to see if there being two identical bundles is what's messing up OSX.

Alexey-Yakovenko:
I removed all copies of deadbeef.app from disk, and kept only one (the release build in xcode temp folder).

Then I ran deadbeef, signed out, signed back in, and ran deadbeef again.

The result:

Now deadbeef doesn't show up in the list of applications to open MP3 (Open With -> Other), but it shows up in the Open With menu for MP3.

Alexey-Yakovenko:
> Repeated the test (uninstalled and installed VLC) and still works...

You don't need to uninstall and install VLC. You need to have VLC installed, then remove all deadbeef.app instances, reboot/signout, then install deadbeef back.

The condition here is that VLC is present before deadbeef is installed.

If deadbeef is installed before VLC -- it works (I wrote it in [this comment](https://github.com/Alexey-Yakovenko/deadbeef/pull/1453#issuecomment-177878027) already).

But I don't expect anyone to remove VLC to make deadbeef file associations work.

carlosanunes:
> You don't need to uninstall and install VLC. You need to have VLC installed, then remove all deadbeef.app instances, reboot/signout, then install deadbeef back.

Did not reboot when tested this... Yup, still not showing up... Arrghhh...

> But I don't expect anyone to remove VLC to make deadbeef file associations work.

I do not, just part of the tests I made to make sure a 'new' app would not override existing one.

I am going to remove the file association stuff from this PR until I get this to work.

Alexey-Yakovenko:
kk

carlosanunes:
Removed the file association commits associated with the plist file. 

Will start over, first just with UTIs that are already known to the system and therefore not a problem (e.g.: MP3, WMA, WAV, etc.). And then tackle the UTI conflicts issue separately, using the 'deprecated' stuff...

Sounds good ?

Alexey-Yakovenko:
Why not just do what VLC and Chrome are doing? To me it looks like you're trying to solve an easy problem in a complicated way. You'll have to use deprecated functions anyway in the end, so using 2 different key names sounds bad to me. We know that using non-deprecated key names won't bring any value.

carlosanunes:
> Why not just do what VLC and Chrome are doing?

Chrome and VLC actually use (explicit) UTIs, VLC for volume drop and Chrome for its own chrome extensions format. But can't say for sure for what reason... 

> We know that using non-deprecated key names won't bring any value.

OSX seems to prioritise applications that explicitly use UTI in the 'Open With' list. (pedantic answer :P)
A less pedantic answer is that it is less xml we need to use and less type definition from our part. The identifiers are already there, at least, for the system defined UTIs.

This might be me being stubborn... I am just not too comfortable in using deprecated stuff with undocumented behaviour when I don't really have to. Weird behaviour can popup on anytime (more than with the supported stuff at least :D ).

Issue #1432, opened 2016-01-01
deadbeef doesn't play nice with tiling window managers


orbea:
On spectrwm 2.7.2 I am running the gtk2 deadbeef 0.6.2 compiled from source with this patch (Which it won't compile without). https://projects.archlinux.org/svntogit/community.git/tree/trunk/plugins_vfs_zip_vfs_zip_c.patch?h=packages/deadbeef

To illustrate the issue here is my current deadbeef layout.
![1](http://ks392457.kimsufi.com/orbea/stuff/pics/scrots/deadbeef/2015-12-31-233111_748x1040_scrot.png)

When I open a new terminal deadbeef resizes to this, which is not really an issue.
![2](http://ks392457.kimsufi.com/orbea/stuff/pics/scrots/deadbeef/2015-12-31-233122_1680x1050_scrot.png)

Now the problem is that when I close the new terminal deadbeef resizes to its original size, but fails to resize its layout appropriately leaving it for the user to do so manually, which quickly becomes tedious.
![3](http://ks392457.kimsufi.com/orbea/stuff/pics/scrots/deadbeef/2015-12-31-233152_748x1040_scrot.png)
![4](http://ks392457.kimsufi.com/orbea/stuff/pics/scrots/deadbeef/2015-12-31-233234_748x1040_scrot.png)

Alexey-Yakovenko:
Other related issues: #975, #1378

cboxdoerfer:
@orbea, This should be fixed on master. Would be great if you could verify that.

orbea:
Yes, I can confirm this is fixed, thanks!

PR #1488, opened 2016-02-18
gtkui: splitter: add children always with resize property set to FALSE

### LINKED ISSUE #1465: DeaDBeeF 0.7 resets user settings of geometry of widgets
This is how it should be: [config_user.txt](https://github.com/Alexey-Yakovenko/deadbeef/files/112604/config_user.txt)
This is what you get after restart player [config.txt](https://github.com/Alexey-Yakovenko/deadbeef/files/112624/config.txt)

In screenshots:
User settings (approximately):
before

After restart:
after

cboxdoerfer:
This makes the splitter widget behave like the foobar2000 one when the movement isn't locked. Instead of only resizing the right/bottom pane, we now resize both panes proportionally to their width/height.

In order to fix some race conditions with different gtk versions, different window managers and other factors when deadbeef launches in maximized mode, I also had to add an ugly hack to ensure that the splitter position is always restored correctly. The fix is similar to the one of the auto column resizing: We now also store the width/height of the splitter, so we can calculate the correct splitter position even if the window dimensions changed between stop and start of the player. So for each splitter widget I wait for the first expose event (at which point the widget should have a valid dimension) and adapt the splitter position according to the new window dimensions.

This should fix #1465 and #1432

### DIFF
--- a/plugins/gtkui/widgets.c
+++ b/plugins/gtkui/widgets.c
@@ -96,6 +96,9 @@ typedef struct {
     GtkWidget *box;
     int position;
     int locked;
+    int width;
+    int height;
+    gulong signal_id;
 } w_splitter_t;
 
 typedef struct {
@@ -595,7 +598,7 @@ w_draw_event (GtkWidget *widget, cairo_t *cr, gpointer user_data) {
 
         if (!gtk_widget_get_has_window (widget)) {
 #if GTK_CHECK_VERSION(3,0,0)
-        cairo_translate (cr, -allocation.x, -allocation.y);
+            cairo_translate (cr, -allocation.x, -allocation.y);
 #endif
             cairo_reset_clip (cr);
             cairo_rectangle (cr, allocation.x, allocation.y, allocation.width, allocation.height);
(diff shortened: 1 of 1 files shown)


Alexey-Yakovenko:
I'll need to test this before merging. Can do in about a week or so.

Alexey-Yakovenko:
Works, but unstable.

Resizing window many times leads to loss of the relative splitter position at the same window size.

I.e. imaging that the window originally has size W, and splitter has the position X.

After resizing the window many times, and restoring it back to position W, the splitter position may become X-10.

Let's discuss how to fix it.

Possibly remember the starting position and window size, when the resizing starts, and "stick" to it, by using fractional values?

cboxdoerfer:
>Possibly remember the starting position and window size, when the resizing starts, and "stick" to it, by using fractional values?

Yes, this should work. Will try it asap. 

Btw. here are two GTK+ bug reports that are related to the issues we are having:
https://bugzilla.gnome.org/show_bug.cgi?id=58952
https://bugzilla.gnome.org/show_bug.cgi?id=733638

And there seems to be also some work on a new paned widget that might get added to GTK+ at some point: https://github.com/chergert/panel-gtk

Alexey-Yakovenko:
We can't rely on this, as it would only work in latest (future) GTK3 versions.

Unless we could backport it to GTK2.12 somehow (or, well, 2.16 when we drop 2.12 support).

cboxdoerfer:
Yes, I know. Just wanted to mention it in case we decide to use a custom/modified widget and need some code for reference.

Do have an idea how we can detect changes in size of the gtkpaned widget? I can't get it to work since it doesn't receive any configure events.

Alexey-Yakovenko:
Nope, sorry.. Didn't really dig the GtkPaned that much

Alexey-Yakovenko:
@cboxdoerfer I guess you're planning to post a new PR when you finish with the new widget impl? Let me know if you wanted to update this one, then I'll reopen.

cboxdoerfer:
Yes, I'm going to open a new one. Then we might also use a separate branch for that.

PR #1495, opened 2016-03-03
Assorted FreeBSD fixes

novel:


Alexey-Yakovenko:
There are more places which use syslimits.h:

```
$ grep --include "*.[ch]" -r syslimits .
./plugins/artwork/artwork.c:    #include <sys/syslimits.h>
./plugins/converter/converter.c:#include <sys/syslimits.h>
./plugins/converter/convgui.c:#include <sys/syslimits.h>
./plugins/liboggedit/oggedit_internal.h:#include <sys/syslimits.h>
```

Where did you get the idea that syslimits shouldn't be used in the user applications?
What problem this change is supposed to fix?

novel:
> Where did you get the idea that syslimits shouldn't be used in the user applications?

In the commit message I provided this link: https://svnweb.freebsd.org/base/head/sys/sys/syslimits.h?revision=194498. If you open it, you''ll notice a piece like that:
```
#if !defined(_KERNEL) && !defined(_LIMITS_H_) && !defined(_SYS_PARAM_H_)
#ifndef _SYS_CDEFS_H_
#error this file needs sys/cdefs.h as a prerequisite
#endif
#ifdef __CC_SUPPORTS_WARNING
#warning "No user-serviceable parts inside."
#endif
#endif
```

Do you think my understanding of this is wrong?

> What problem this change is supposed to fix?

```
gmake[2]: Entering directory '/usr/home/novel/code/deadbeef/plugins/artwork'
  CC       artwork.lo
In file included from artwork.c:46:0:
/usr/include/sys/syslimits.h:41:2: warning: #warning "No user-serviceable parts inside." [-Wcpp]
 #warning "No user-serviceable parts inside."
  ^
  CC       cache.lo
cache.c: In function 'remove_cache_item':
cache.c:85:26: error: 'PATH_MAX' undeclared (first use in this function)
     char cache_root_path[PATH_MAX];
                          ^
cache.c:85:26: note: each undeclared identifier is reported only once for each function it appears in
cache.c: In function 'path_ok':
cache.c:110:93: error: 'PATH_MAX' undeclared (first use in this function)
     return strcmp (entry, ".") && strcmp (entry, "..") && dir_length + strlen (entry) + 1 < PATH_MAX;
                                                                                             ^
cache.c: In function 'cache_cleaner_thread':
cache.c:117:22: error: 'PATH_MAX' undeclared (first use in this function)
     char covers_path[PATH_MAX];
                      ^
Makefile:622: recipe for target 'cache.lo' failed
```

Alexey-Yakovenko:
> Do you think my understanding of this is wrong?

Yes.

novel:
okthnx

Alexey-Yakovenko:
I applied the fixes for both problems you found. Thanks for the help.

novel:
Thanks, but it's not quite there yet:

```
gmake[2]: Entering directory '/usr/home/novel/code/deadbeef/plugins/artwork'
  CC       artwork.lo
artwork.c: In function 'make_cache_dir_path':
artwork.c:891:21: error: 'NAME_MAX' undeclared (first use in this function)
     char esc_artist[NAME_MAX+1];
                     ^
artwork.c:891:21: note: each undeclared identifier is reported only once for each function it appears in
artwork.c: In function 'make_cache_path2':
artwork.c:944:41: error: 'NAME_MAX' undeclared (first use in this function)
     if (make_cache_dir_path (path, size-NAME_MAX, artist, img_size)) {
                                         ^
artwork.c: In function 'process_scaled_query':
artwork.c:1547:24: error: 'PATH_MAX' undeclared (first use in this function)
     char unscaled_path[PATH_MAX];
                        ^
artwork.c: In function 'process_query':
artwork.c:1630:21: error: 'PATH_MAX' undeclared (first use in this function)
     char cache_path[PATH_MAX];
                     ^
artwork.c: In function 'get_album_art':
artwork.c:1847:21: error: 'PATH_MAX' undeclared (first use in this function)
     char cache_path[PATH_MAX];
                     ^
artwork.c: In function 'invalidate_playitem_cache':
artwork.c:2035:29: error: 'PATH_MAX' undeclared (first use in this function)
             char cache_path[PATH_MAX];
                             ^
Makefile:622: recipe for target 'artwork.lo' failed
```

Needs something like this to work:

```
diff --git a/plugins/artwork/artwork.c b/plugins/artwork/artwork.c
index 89b3905..0679af4 100644
--- a/plugins/artwork/artwork.c
+++ b/plugins/artwork/artwork.c
@@ -51,6 +51,7 @@
     #include <jpeglib.h>
     #include <png.h>
 #endif
+#include <limits.h>
 #include "../../deadbeef.h"
 #include "artwork_internal.h"
 #include "lastfm.h"
diff --git a/plugins/artwork/artwork_internal.c b/plugins/artwork/artwork_internal.c
index b7c47bc..3b2d663 100644
--- a/plugins/artwork/artwork_internal.c
+++ b/plugins/artwork/artwork_internal.c
@@ -32,6 +32,7 @@
 #include <dirent.h>
 #include <unistd.h>
 #include <sys/stat.h>
+#include <limits.h>
 #include "../../deadbeef.h"
 #include "artwork_internal.h"

```

Also I'm wondering if you still need AC_CHECK_HEADERS([sys/syslimits.h]) bit in configure.ac

PS It'd be really helpful if you elaborate your responses rather than giving curt answers.

Alexey-Yakovenko:
fxd artwork.c and artwork_internal.c too. cheers.

> PS It'd be really helpful if you elaborate your responses rather than giving curt answers.

The only thing that I wrote is that I think that your understanding of the code in syslimits.h is wrong.

I don't get what do I need to elaborate? Didn't mean to offend you.

This is not relevant to fixing the problem anyway.

Your patch was not doing the right thing anyway, and missed out another similar potential error with closedir.

> Also I'm wondering if you still need AC_CHECK_HEADERS([sys/syslimits.h]) bit in configure.ac

not really, but it doesn't hurt anything.

Alexey-Yakovenko:
Another potential problem with this.. I think that the sys/syslimits.h usage was intentional, and was required on FreeBSD for some reason. that is why BSD is mentioned in configure.ac.

It can happen that something will break on older FreeBSD versions.

novel:
> The only thing that I wrote is that I think that your understanding of the code in syslimits.h is wrong.

Yeah, so I'm curios to know what is your understanding.

> It can happen that something will break on older FreeBSD versions.

Yep, I noticed that it was intentional, but it's not obvious why approach with syslimits.h was chosen in the first place.

Anyway, it builds on FreeBSD 9.x, and earlier versions are not supported by FreeBSD itself, I'm not sure if you care about those older versions at all.

Alexey-Yakovenko:
> Yeah, so I'm curios to know what is your understanding.

My understanding is that this file is not covered by the C standard, and therefore the way it's supposed to be used it undefined (i.e. depends on the needs of the software developer's code).

I don't know what the warning in BSD syslimits.h is supposed to mean.

Alexey-Yakovenko:
> Anyway, it builds on FreeBSD 9.x, and earlier versions are not supported by FreeBSD itself, I'm not sure if you care about those older versions at all.

Not really.

novel:
> Not really.

Good, then we're fine with this, I think. Thanks.

novel:
You probably can consider the following change:

```

index d534ed1..e178d81 100644
--- a/plugins/artwork/cache.c
+++ b/plugins/artwork/cache.c
@@ -167,7 +167,7 @@ cache_cleaner_thread (void *none)
             usleep (100000);
             deadbeef->mutex_lock (thread_mutex);
         }
-        if (covers_subdir) {
+        if (covers_dir) {
             closedir (covers_dir);
             covers_dir = NULL;
         }
```

Alexey-Yakovenko:
oops :) thanks for spotting

Alexey-Yakovenko:
fxd

novel:
Thanks! I'll test.
Meanwhile, I'm seeing another crash, I suppose in ddb_gui_GTK2.so. It doesn't happen often (maybe a 1 or 2 times / day max), though I had no luck to properly catch it (either running version compiled without debugging or forgetting to run in lldb). I'll let you know if I'll finally manage to chase it down.

Alexey-Yakovenko:
Sure. Just make sure to open another issue.

PR #1516, opened 2016-03-22
Tokenised playlist search

FauxFaux:
This patch changes the format of the search query, used in the "find" dialog.

Previously, it searched for the entire input string in any single metadata tag.

Now, it tokenises the string into words, and looks for each word in each metadata tag.  Quoting is supported, as is specifying the tag to look in.

Examples:

search term | action
--- | ---
foo bar | searches for "foo" and "bar" independently, so matches "01. Bar Foo - Quux"
"foo bar" | searches for "foo bar" as a whole, like the current search, so doesn't match the above
foo "bar baz" | searches for "foo" and "bar baz" independently, so matches "baz bar foo", but not "foo bar and baz"
title:foo | searches for a track with "foo" in the title, but doesn't look in any other metadata tag
title:"foo bar" | searches for a track with "foo bar" in the title, but doesn't look in any other metadata tag
title:foo bar | searches for a track with "foo" in the title, and "bar" in any metadata tag

This appears to work for me, but I'm sure there are problems.  Major things for review:

 * I don't understand what `cmpidx` is doing at all.
 * Using `strstr` to match the metadata key names is almost certainly not the best way to do it.
 * There's sure to be a bug in the parser for queries; there always is.
 * I don't understand why that code was duplicated before (or was it not?), and if the order mattered around cmpidx.  See the "extract method" and "eliminate duplicated" commits.
 * Where's the appropriate place to document this?

github has done a poor attempt of displaying the diff in the pull request; please consider looking at the commits, or using a proper diff viewer.


Alexey-Yakovenko:
Have you compared performance between new and old code?

cboxdoerfer:
I'm not sure if the syntax is a good idea. Why not make use of the title formatting? The user is already familiar with that and it is much more flexible. Also this would allow us to implement the query syntax of foobar2000 in the future (or something similar) without breaking compatibility.

Alexey-Yakovenko:
cmpidx: that's an optimisation, which allows to reduce the number of comparisons, because lots of strings are duplicated in many songs of the same album. We store the comparison result, as positive or negative integer, in each compared string. Then when we compare again -- we can quickly find out the result, during the same plt_search_process call.

FauxFaux:
Right, I think I understand `cmpidx` now.  The way I'm using it is (very) wrong, which explains why I sometimes get incorrect matches (which I hadn't noticed last night).  To apply the same technique to multiple (`n`) search terms I'd need `n` extra bytes of storage, which I presume don't exist before `m->value`.

I'll have a go at taking this performance optimisation out, which I don't feel would have that much effect, and I'll gather some other benchmarks too, if performance is a concern.  For the simple case (where you quote the input), the amount of work done is exactly the same as in the old code, excluding a couple of pointer dereferences and the parsing step; i.e. I'm not expecting much difference.


For the search syntax: I was copying what I see as the search syntax absolutely everywhere: Github issues (`is:issue`), Google/gmail (`filetype:pdf`), Kibana (`tags:foo`), Spotify (`title:"native american"`), ...
I feel that this is what people are familiar with.  (I suspect most of these have the same syntax as they're backed onto exactly the same technology.)

I can't even imagine what you mean by using title formatting.  The only other option I considered was doing a full string match across a formatted version of the title; let the user specify `%artist% - %title%` then match "yonce - Single La" as a full text string, which is not currently possible.

Maybe add some functions like `$search(%artist%, beyo)`, and interpret the result as a boolean?  Seems very arcane.  What am I missing?

FauxFaux:
The new search isn't significantly slower than the crippled old search:

Crippled old search:

````
search for '', time: 0ms
search for 't', time: 47ms
search for 'th', time: 211ms
search for 'thi', time: 295ms
search for 'this', time: 304ms
...
search for 'this gets', time: 316ms
...
search for 'this gets quite a bit slower as the length goes up', time: 421ms
````

New search:
````
search for '"', time: 14ms
search for '"t', time: 38ms
search for '"th', time: 203ms
search for '"thi', time: 301ms
search for '"this', time: 310ms
...
search for '"this gets', time: 325ms
...
search for '"this gets quite a bit slower as the length goes up', time: 432ms
````

There is, however, a significant (but limited) improvement from the `cmpidx` optimisation:

Real old search:

````
search for 't', time: 31ms
search for 'th', time: 141ms
search for 'thi', time: 151ms
search for 'this', time: 135ms
search for 'this ', time: 136ms
...
search for 'this gets', time: 138ms
...
search for 'this gets quite a bit slower as the length goes up', time: 155ms
````

I believe the 150ms per keypress is still annoyingly too slow, though. I would rather:

 * add some debouncing
 * only scanning cheap metadata fields configurably, instead of just skipping `cuesheet` and `log`
 * e.g. new search can do:

````
search for 'title:angel', time: 26ms
search for 'title:angel artist:robbie', time: 76ms
````

Even for two matches, it's annihilating the original search (which can't even express the same query).

(No, I didn't really mean to push the timing commit into this pull request, but maybe it's useful anyway.)

(Test playlist: 15,000 tracks, mixed ogg/mp3 (and tiny numbers of minority formats), heavy in MusicBrainz metadata.  23MB as a `.dbpl`.  i7-3770k.)

FauxFaux:
Debounce makes it almost usable for long queries, but hilights how annoying it is.  I experimented with bounces in the 100-300ms range: https://github.com/FauxFaux/deadbeef/commit/aafc3c996fd26e7e0f69685b7f67e8a8d39508d0

FauxFaux:
I changed the search to default to searching only the "artist", "title" and "album" fields, unless you explicitly ask it to search for something else.

This is probably quite a bit more controversial.

e.g. `ppk` finds nothing, but `artistsort:ppk` finds the ППК tracks, as expected.

cboxdoerfer:
>I can't even imagine what you mean by using title formatting. The only other option I considered was doing a full string match across a formatted version of the title; let the user specify %artist% - %title% then match "yonce - Single La" as a full text string, which is not currently possible.

Well, use title formatting for accessing fields instead of introducing a new syntax for field access. foobar2000 does it like that: `%artist% HAS test`

This would give the user much more flexibility because they can access arbitrary fields either with `%field%` or `$meta(field)`, and not just the ones we add to the search syntax. And not forget all the functions in title formatting, e.g. in foobar2000 you can do something like that `"$strcmp(%artist%,%album%)" IS 1`. This would search for all tracks where %artist% and %album% are identical.

FauxFaux:
That makes sense.

The simple case you've highlighted is functionally equivalent to what is implemented: `artist:test`.  There looks to be some value in using the field name lookup from the `tf_` code, which maps some common cases together.. but they have spaces in, which messes up my beautiful search syntax.

I don't think I like:
`(%artist% HAS "foo bar") AND (%track number% IS 6)` when:
`artist:"foo bar" tracknum:6`
... works (so long as you know the magic internal field names).  Either way, I don't think I care enough to go rewriting the `tf_` parser to deal with this kind of thing.

Maybe just don't document the "foo:bar" syntax (or take it out!), and leave the search at words or quoted phrases?  The `cmpidx` hack could be put back in for the first search term, or the first two, or ... hmm.

cboxdoerfer:
> I don't think I like:
(%artist% HAS "foo bar") AND (%track number% IS 6) when:
artist:"foo bar" tracknum:6
... works (so long as you know the magic internal field names). 

Of course it works, but if we decide to implement OR operators, or all the fancy keywords foobar2000 [supports](http://wiki.hydrogenaud.io/index.php?title=Foobar2000:Query_syntax#Advanced_search) it might break. So at least we should consider this in advance and decide if we want this kind of flexibility in the future and design the query syntax according to that.

> Either way, I don't think I care enough to go rewriting the tf_ parser to deal with this kind of thing.

Why would you need to rewrite the tf parser? Just use the existing one.

> Maybe just don't document the "foo:bar" syntax (or take it out!), and leave the search at words or quoted phrases? The cmpidx hack could be put back in for the first search term, or the first two, or ... hmm.

It's not that I don't like the style of the syntax. All I'm saying is: we should discuss in advance where we want to go with the search and design the query syntax according to that. If all we want from the search is accessing a couple of selected metadata fields and a basic AND-condition we certainly don't need anything like foobar2000 has, but if not the query syntax should at least be flexible enough so we can add those features later.

Alexey-Yakovenko:
I agree with what @cboxdoerfer is saying. We should be working towards implementing the proper query syntax. Alternative search implementations can be in a separate plugin, or on a branch, or wherever.

Therefore, we should start a conversation about making an API for changing the standard search implementation (I'm not talking about monkey-patching the plt_search_process pointer).

Does this even sound like a good idea in the long run?

Alexey-Yakovenko:
> we should discuss in advance where we want to go

This is especially important. Before changing _anything_ - the starting point is to chat with other devs, and get the right direction.

Alexey-Yakovenko:
@FauxFaux please let me know if you're interested in releasing your search implementation as a plugin. Then we can figure out what APIs are needed to be exposed on deadbeef side. Unfortunately, I can't make this the default search implementation, because it doesn't align with the project direction. Therefore I'm closing the PR.

PR #1519, opened 2016-03-23
Update deadbeef.desktop.in

laichiaheng:
Add some Chinese keywords.

### DIFF
--- a/deadbeef.desktop.in
+++ b/deadbeef.desktop.in
@@ -19,7 +19,7 @@ X-PulseAudio-Properties=media.role=music
 X-Ayatana-Desktop-Shortcuts=Play;Pause;Stop;Next;Prev
 
 Keywords=Sound;Music;Audio;Player;Musicplayer;MP3
-Keywords[zh_TW]=音樂;音樂播放器;播放器;音訊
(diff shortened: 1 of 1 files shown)


Alexey-Yakovenko:
Some explanation is needed about what this change fixes or adds.

laichiaheng:
Add the Chinese keywords for searching this app on menu.

Alexey-Yakovenko:
Putting multiple localisations in one huge line doesn't seem to scale well in the long run.

Is there anything like Keywords[LANGUAGE_CODE] syntax, same as for other entries in the desktop file?

laichiaheng:
Done.

laichiaheng:
Complete

Alexey-Yakovenko:
Hmm.. Do you still have to duplicate english in every localisation?
I would expect that english is included automatically..

Btw, is there any way for me to test how this works?

I mean, what's the use case for that, which DE supports it, etc?

laichiaheng:
I'm using Unity desktop, I edit "/usr/share/applications/deadbeef.desktop" to the last one I posted, and it works for me, and the English still needs to be added for other languages.

Alexey-Yakovenko:
ok, I got it. will merge later (need to collapse the commits, so can't do it via github interface)

PR #1524, opened 2016-03-29
Support webm in ffmpeg plugin.

### LINKED ISSUE #1511: [Support] Playing webm files
Sorry for putting this up here but I wasn't able to find another place to ask for support. 
The issue is Deadbeef doesn't detect *.webm audio files. I looked up in the plugins page and didn't find a related plugin. 

Thanks for any help.

PS - Foobar supports webm files.

xinyazhang:
This fixes #1511

### DIFF
--- a/plugins/ffmpeg/ffmpeg.c
+++ b/plugins/ffmpeg/ffmpeg.c
@@ -68,7 +68,7 @@
 static DB_decoder_t plugin;
 static DB_functions_t *deadbeef;
 
-#define DEFAULT_EXTS "aa3;oma;ac3;vqf;amr;opus;tak;dsf;dff"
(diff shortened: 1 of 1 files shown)


Alexey-Yakovenko:
Why webm specifically? Why not add all file extensions that ffmpeg can support?

Alexey-Yakovenko:
(That's actually an interesting idea -- enumerate all supported file extensions, and expose them all)

xinyazhang:
The problem is ffmpeg will support more formats in later version.
Asking ffmpeg for the format it supports is a good idea.
Maybe I can look into this.

xinyazhang:
Closed since we have a better way to do this.

PR #1527, opened 2016-03-31
ffmpeg: enumerate all extensions supported by ffmpeg.

xinyazhang:
This patch lets the ffmpeg plugin enumerate all extensions supported by ffmpeg. It basically works.
(tested by disabling most decoders, works for mp3, m4a, wv and flac).

However there are some issues to solve so I send this PR mainly for comments.

1. Conflicts with other plugins.
Although ffmpeg works for most formats but it's not the best if we already support some format with specific library.
For example, I cannot play APE flawlessly with ffmpeg on my platform.
Currently I'm planning to have a "blacklist" that disables some extensions for ffmpeg plugin, e.g. ape and other common extensions.
Does this sound OK for you?

2. GUI changes
In this patch I've already added an option to enable this feature, but I think if we are going to use the "blacklist", then it's better to allow users editing this list.

3. What extensions do we support?
Err, ffmpeg supports tons of formats. Is there an elegant way to show the list?

4. Anything else I forgot?

Thanks.

### DIFF
--- a/plugins/ffmpeg/ffmpeg.c
+++ b/plugins/ffmpeg/ffmpeg.c
@@ -68,9 +68,15 @@
 static DB_decoder_t plugin;
 static DB_functions_t *deadbeef;
 
+#define ENABLE_ALL_EXT 1
+
+#if ENABLE_ALL_EXT
+#define EXCLUDE_EXTS "" // TODO: add extensions supported by other plugins
+#endif
+
 #define DEFAULT_EXTS "aa3;oma;ac3;vqf;amr;opus;tak;dsf;dff"
 
-#define EXT_MAX 100
+#define EXT_MAX 256
 
 #define FFMPEG_MAX_ANALYZE_DURATION 500000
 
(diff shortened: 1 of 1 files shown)


Alexey-Yakovenko:
1. I think this is quite easy to solve by making sure that ffmpeg plugin is always at the end of the plugin list, and only used when no other plugin can open a file. I was planning to add this anyway.

I don't think that blacklist is a good solution.

3. There's no elegant way to show / edit the list. Only a string with extensions separated with semicolon.

4. Enumerating _encoder_ file extensions is not going to work, because deadbeef ships ffmpeg without any encoders (built with --disable-encoders). If only _some_ decoders don't claim their supported extensions -- drop them. It actually sounds like a bug in ffmpeg to me.

Alexey-Yakovenko:
I like what you did with the option "Use all extensions", than what I had in mind.

This way, we don't even need to show the user which extensions are supported.

We can even add a hack, to allow ffmpeg to probe ANY files, without ever generating a list.

E.g. `exts = { "*", NULL }`, and treat the "*" as a special indicator, that the plugin can probe the files itself.

That should solve most of the hassle. What do you think?

Alexey-Yakovenko:
I just pushed the necessary changes to master - both appending ffmpeg to the end of plugin list, and supporting "*" as file extension.

xinyazhang:
Adding all extensions ("*") works for drag-drop case, but I don't think it's suitable when using "Open files" dialog, which will show every file. In that case we still need an extension list explicitly.

xinyazhang:
A lot of FFmpeg decoders do not include an extension list because .read_probe is the preferred way and it's more reliable.
However for the GUI we need the list of extensions in the dialog.

But making ffmpeg the last one to load really helps!

Alexey-Yakovenko:
Ahh... I totally forgot that we need the list of file extensions in the file dialogs..

Well, then we just have to use what we have.

This is only a problem with file dialog, but not with folder dialog, or drag-n-drop, or command line.

Alexey-Yakovenko:
Ok, let's keep the detection the way you do it, and kill the blacklist (it's not required).

Cleanup the code, and I will merge.

xinyazhang:
Got it, I'll do it later.

Alexey-Yakovenko:
ENABLE_ALL_EXT - why is this needed? I think you should just make this enabled by default, without any ifdefs.

Alexey-Yakovenko:
Other problems:
* your editor seems to be inserting tabs instead of spaces (This is shown as broken indentation in the diff)
* coding style of the new code doesn't match the existing style, e.g. return type of the functions should be on the separate lines, and there should be space between the function names and argument list opening `(`

There are a few less important problems, but I can fix them post-merge.

xinyazhang:
Updated code pushed. (I rebased so the commit message doesn't change)

xinyazhang:
Hello?

Alexey-Yakovenko:
Sorry, forgot to get back to it. It looks good now.

Alexey-Yakovenko:
merged

PR #1540, opened 2016-04-10
tf: Implement $num(n,len) (fixes #1279)

### LINKED ISSUE #1279: [tf 2.0] Support for $num() fb2k string function
waker, during a short chat with one of your main developers, he too said this function is not just another gimmick, but a valuable addition to TF 2.0.

Let me requote the most important bit from fb2k docs:

`num(n,len)`

Formats the integer number n in decimal notation with len characters.(...) If n is not numeric, it is treated as zero. Examples:

    $num(123,5) → 00123
    $num(-123,5) → -0123
    $num(4.8,5) → 00004
    $num(A1,5) → 00000

Basic idea: there is neither a rounding function in ddb nor in fb2k. So back in my Win2k days, I've simply "misused" this function to turn a '123.47' value (think of BPM!) into a better readable '123'.

I can't deny, that there is lots of __room for improvement towards fb2k__. So deadbeef might, in theory, go ahead and offer a `$num(n, len, digits)` variant that would even allow a `digits` amount of digits (fixed).
That would turn a pretty silly, overexact BPM value of 123.874 into '123.8'.
Equivalent functions in other languages will be `trunc()` or, IIRC from my 80s BASIC days, `FIX()`. :)

Thanks for considering in advance.

Serranya:
I hope everything is alright. I am no professional c developer.

### DIFF
--- a/tf.c
+++ b/tf.c
@@ -204,6 +204,63 @@ tf_func_strcmp (ddb_tf_context_t *ctx, int argc, const char *arglens, const char
     return !res;
 }
 
+// $num(n,len) Formats the integer number n in decimal notation with len characters. Pads with zeros
+// from the left if necessary. len includes the dash when the number is negative. If n is not numeric, it is treated as zero.
+int
+tf_func_num (ddb_tf_context_t *ctx, int argc, const char *arglens, const char *args, char *out, int outlen, int fail_on_undef) {
+    const char *arg = args;
+    int bool_out = 0;
+    int len;
+
+    if (argc != 2) {
+        return -1;
+    }
+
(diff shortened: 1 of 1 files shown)


Alexey-Yakovenko:
malloc, free are unacceptable.

snprintf is.. undesirable :)

Serranya:
So should I use arrays instead of malloc/free ?

And maybe manually pad the number string instead of using snprintf ? Or just a different function I dont know yet ?

Alexey-Yakovenko:
Please see e.g. how the `$hex` is implemented -- it does a very similar thing.

Serranya:
I got rid of the snprintf and dynamic memory allocation.

I would appreciate another review from you.

Issue #1469, opened 2016-02-02
Hiding/restoring maximized window by clicking tray icon doesn't preserve widget layout


Alexey-Yakovenko:
Continued from the bug #1465.

Fixing the issue with tray is different than fixing the issue on start/quit, therefore the new bug.

Details: The window gets "restored" from maximized state at some point, and the widgets get "shrunk" in the process. That's especially a problem with Splitter widgets, which don't automatically adopt to window resizing, but clip to the max possible size instead.

Alexey-Yakovenko:
The problem: we can't really prevent GTK from "unmaximizing" the window during restoration from notification area.

The proposed solution: implement proportional scaling for splitters, in reaction to configure event. That's quite a lot of work, so can't be done for 0.7.1, and should go on master.

cboxdoerfer:
This should be fixed with the recent splitter changes on master.

Issue #1378, opened 2015-12-07
gtkui: splitters: allow locking size of one side, leaving another dynamic


Soukyuu:
Currently, it's only possible to lock the splitter position, which leads to all sorts of problems when resizing the window (most of the time, they don't resize or break layout). It would be much nicer if we could lock one side of the splitter, fixing one component's size while allowing the other side to be sized dynamically.

While it's possible to achieve that (kind of) with leaving the splitters unlocked, it leads to several side issues. For example, use this:
```
 gtkui.layout.0.6.2 vsplitter pos=897 locked=0 {hsplitter pos=1341 locked=1 {tabs active=0 num_tabs=2 tab000="Now Playing" tab001="Library" {playlist hideheaders=0 {} placeholder {} } vsplitter pos=527 locked=1 {tabs active=0 num_tabs=2 tab000="Cover" tab001="Properties" {coverart {} selproperties {} } vsplitter pos=25 locked=1 {hsplitter pos=157 locked=1 {playtb {} hbox expand="1 1 1 1" fill="1 1 1 1" homogeneous=1 {button color="#000000" textcolor="#000000" label="SAC" action="toggle_stop_after_current" use_color=0 use_textcolor=0 {} button color="#000000" textcolor="#000000" label="Linear" action="order_linear" use_color=0 use_textcolor=0 {} button color="#000000" textcolor="#000000" label="Shuffle" action="order_shuffle" use_color=0 use_textcolor=0 {} button color="#000000" textcolor="#000000" label="Repeat" action="loop_cycle" use_color=0 use_textcolor=0 {} } } tabs active=1 num_tabs=2 tab000="Visualization" tab001="Playlists" {musical_spectrum {} pltbrowser {} } } } } waveform_seekbar {} } 
```

With unlocked splitters, the waveform part of the splitter keeps resizing itself on every start, even though I'm trying to keep it at about half the size: [intended](https://dl.dropboxusercontent.com/u/19330332/ddb_intended.jpg) and [after restart](https://dl.dropboxusercontent.com/u/19330332/ddb_unlocked_splitters.jpg)

[...]

Issue #975, opened 2015-03-12
Not remembering the positioning of viewports\windows on restart


Alexey-Yakovenko:
Original [issue 1073](https://code.google.com/p/ddb/issues/detail?id=1073) created by Alexey-Yakovenko on 2014-03-09T04:40:10.000Z:

<b>What steps will reproduce the problem?</b>
<b>Какие шаги приводят к воспроизведению</b>
<b>проблемы?</b>
1. Design a custom layout
Ex: 2 horizontal splits (3 slices)
   Left: File Browser (Plugin)
   Middle: Playlist with Tabs
   Right: Horizontal Split
      Top: Selection Properties
      Bottom: Spectrum Analyser
2. Close Deadbeef
3. Start Deadbeef

<b>What is the expected output? What do you see instead?</b>
<b>Какой ожидаемый вывод? Что вы видите</b>
<b>вместо него?</b>

All the window\viewports position (size only) has changed, when I would expect it to open the same way I opened it.  I have reposition\expand\etc the different windows to make them usable again.

<b>What version of the product are you using? On what operating system and CPU</b>
<b>architecture?</b>
<b>Какую версию продукта вы используете? На</b>
<b>какой операционной системе и</b>
<b>архитектуре CPU?</b>

UbuntuGnome 14.04 (64) - 3.13.0-10-generic
DeadBeef version 0.6.1
Installed: debubuntu package amd64

<b>How did you install the product?</b>
<b>Как вы установили продукт?</b>
install manager

<b>Please provide any additional information below.</b>
<b>Пожалуйста, предоставьте любую</b>
<b>дополнительную информацию ниже.</b>

cboxdoerfer:
This should be fixed with the recent splitter changes on master.

PR #1544, opened 2016-04-17
gtkui: add new splitter widget

cboxdoerfer:
This pull request adds a new container widget (DdbSplitter, similar to GtkPaned) to the gtkui that's now used internally by the horizontal and vertical splitter (instead of GtkPaned and GtkBox). It works similar to the foobar2000 splitter widget.

The splitter widget now has the following config entries:
- `locked`
  - `0`: proportional scaling
  - `1`: lock size of first child
  - `2`: lock size of second child (new)
- `prop` (new)
  - `float_value` `[0.0, 1.0]`: the proportion value used in proportional scaling
- `size_c1` (new)
  - the size of the first child, this is the same as the old `pos` value and used to determine the size when child 1 is locked
- `size_c2` (new)
  - the size of the second child, used when child 2 is locked

The container has the following features:
- ability to hold up to two child widgets
- proportional scaling
- ability to lock the size of the first or second widget
- horizontal or vertical layout
- if no child is locked double click on splitter handle resets the handle pos to half the splitter width/height

This should fix the following issues: #1432 #975 #1378 #1469

### DIFF
--- a/plugins/gtkui/widgets.c
+++ b/plugins/gtkui/widgets.c
@@ -47,6 +47,7 @@
 #include "ddbvolumebar.h"
 #include "callbacks.h"
 #include "drawing.h"
+#include "ddb_splitter.h"
 
 //#define trace(...) { fprintf(stderr, __VA_ARGS__); }
 #define trace(fmt,...)
@@ -94,7 +95,9 @@ typedef struct {
 typedef struct {
     ddb_gtkui_widget_t base;
     GtkWidget *box;
-    int position;
+    int size_c1;
+    int size_c2;
+    float prop;
     int locked;
 } w_splitter_t;
 
@@ -1170,11 +1173,17 @@ w_splitter_load (struct ddb_gtkui_widget_s *w, const char *type, const char *s) 
     for (;;) {
         get_keyvalue (s,key,val);
 
-        if (!strcmp (key, "pos")) {
-            ((w_splitter_t *)w)->position = atoi (val);
-        }
-        else if (!strcmp (key, "locked")) {
+        if (!strcmp (key, "locked")) {
             ((w_splitter_t *)w)->locked = atoi (val);
+        }
+        else if (!strcmp (key, "prop")) {
+            ((w_splitter_t *)w)->prop = atof (val);
+        }
+        else if (!strcmp (key, "size_c1")) {
+            ((w_splitter_t *)w)->size_c1 = atoi (val);
+        }
+        else if (!strcmp (key, "size_c2")) {
+            ((w_splitter_t *)w)->size_c2 = atoi (val);
         }
     }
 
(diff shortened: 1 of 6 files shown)


plugins/gtkui/widgets.c
```diff
@@ -94,7 +95,9 @@ typedef struct {
 typedef struct {
     ddb_gtkui_widget_t base;
     GtkWidget *box;
-    int position;
+    int size_c1;
```
Alexey-Yakovenko: what "c1" and "c2" means? wouldn't it be easier to read/write if it's just size_left, and size_right?

plugins/gtkui/widgets.c
```diff
@@ -1170,41 +1173,41 @@ w_splitter_load (struct ddb_gtkui_widget_s *w, const char *type, const char *s)
...
+        }
+        else if (!strcmp (key, "size_c1")) {
+            ((w_splitter_t *)w)->size_c1 = atoi (val);
+        }
+        else if (!strcmp (key, "size_c2")) {
+            ((w_splitter_t *)w)->size_c2 = atoi (val);
+        }
+        else if (!strcmp (key, "pos")) {
+            // import old pos value
+            ((w_splitter_t *)w)->size_c1 = atoi (val);
```
Alexey-Yakovenko: There's one concern about this:

I guess, that when you write the configuration back, you don't write the "pos", which means if you need to run older deadbeef version - it won't get the correct configuration..

To handle this issue, I suggest keeping the "pos" (and write it back), but ignore it if size_c1 is present.

Alexey-Yakovenko: Another possibility is to just rename "size_c1" into "pos", because they seem to mean the same thing -- then there's no issues with compatibility at all, and no duplicate data.

plugins/gtkui/widgets.c
```diff
@@ -94,7 +95,9 @@ typedef struct {
 typedef struct {
     ddb_gtkui_widget_t base;
     GtkWidget *box;
-    int position;
+    int size_c1;
+    int size_c2;
+    float prop;
```
Alexey-Yakovenko: The use of word "prop" for "proportion" is very confusing to me. Perhaps the word "ratio" is a more common term for this.

Alexey-Yakovenko: Even if we keep using "proportion" there -- I believe it's better to use the whole word everywhere across the code, instead of "prop".

plugins/gtkui/widgets.c
cboxdoerfer: child1 and child2. But yes, of course we can use size1 and size2 instead. I'll fix that.

plugins/gtkui/widgets.c
cboxdoerfer: Yes, I'll go with the second suggestion.

plugins/gtkui/widgets.c
cboxdoerfer: Sure np.

cboxdoerfer:
Fixed those issues.

Issue #1543, opened 2016-04-13
Use video frame as cover artwork


xinyazhang:
I have several m4a files tagged by iTunes, but I found iTunes embedded cover artworks as the video track.
For example, ffprobe outputs the following info for one of these m4a files.
```
  Duration: 00:02:06.90, start: 0.000000, bitrate: 149 kb/s
    Stream #0:0(eng): Audio: aac (LC) (mp4a / 0x6134706D), 44100 Hz, stereo, fltp, 142 kb/s (default)
    Metadata:
      creation_time   : 2036-09-18 22:48:48
    Stream #0:1: Video: mjpeg, yuvj420p(pc, bt470bg/unknown/unknown), 490x500 [SAR 1:1 DAR 49:50], 90k tbr, 90k tbn, 90k tbc
```

Can we parse the video track and use its image as the artwork?

P.S. I'll handle this if nobody else are working on it.

Alexey-Yakovenko:
I'm fairly sure that it is possible to get the content of this "video" track, and it will be a simple JPEG file. But it's hard to say for sure without having the file.

Anyway, no, deadbeef can't read album art from mp4 files (yet). But should be possible.

xinyazhang:
After taking a look of the plugins there are two approaches to implement this.

The first one is to call ffmpeg to address this issue
Pros: handles all formats, e.g. an MKV file with XviD video stream. (for real video stream I'm planning to sample an arbitrary frame)
Cons: replying on ffmpeg to work, if user disabled or doesn't have ffmpeg then it doesn't work.

The other one is to relay the request to aac plugin, which parses the m4a file.
Pros: Works fine for users who wants aac artwork support but doesn't have ffmpeg.
Cons: only works for m4a files, only works for limited image formats.

I prefer the first one but OK with both.
What do you think of it?

Alexey-Yakovenko:
How about solution #3: call libmp4ff directly to get to the album art?

Alexey-Yakovenko:
Do you have a real example of an audio file with album art encoded as a video stream?

xinyazhang:
Yes I have, but all of them are copyrighted, but I think I can craft one with iTunes.

xinyazhang:
Here it is:
[01 Short1.m4a.zip](https://github.com/Alexey-Yakovenko/deadbeef/files/219879/01.Short1.m4a.zip)

[...]

PR #1546, opened 2016-04-17
artwork: fix album art retrieval for m4a files.

xinyazhang:
1. Add function to mp4ff for handling binary blobs.
	- Previously mp4ff assume all tags are C string, which is not true for
	  album arts.
2. Add mp4_extract_art to artwork plugin to utilize mp4ff.
3. Link mp4ff to artwork plugin, now libmp4ff is mandatory for artwork.

Additional notes for PR:
It seems mp4ff uses a mixed coding style so I kept the original tabs unchanged.

### DIFF
--- a/plugins/artwork/artwork.c
+++ b/plugins/artwork/artwork.c
@@ -60,6 +60,7 @@
 #include "wos.h"
 #include "cache.h"
 #include "artwork.h"
+#include "mp4ff.h"
 
 //#define trace(...) { fprintf (stderr, __VA_ARGS__); }
 #define trace(...)
@@ -1507,6 +1508,56 @@ apev2_extract_art (const char *fname, const char *outname) {
     return err;
 }
 
+static uint32_t
+mp4_fp_read (void *user_data, void *buffer, uint32_t length) {
+    DB_FILE* stream = user_data;
+    uint32_t ret = (uint32_t)deadbeef->fread (buffer, 1, length, stream);
+    return ret;
+}
+
+static uint32_t
+mp4_fp_seek (void *user_data, uint64_t position) {
+    DB_FILE* stream = user_data;
+    return deadbeef->fseek (stream, (int64_t)position, SEEK_SET);
+}
+
+static int
+mp4_extract_art (const char *fname, const char *outname) {
+    int ret = 0;
+    DB_FILE* fp = deadbeef->fopen (fname);
+    if (!fp)
+        return -1;
+
+    mp4ff_callback_t cb = {
+        .read = mp4_fp_read,
+        .write = NULL,
+        .seek = mp4_fp_seek,
+        .truncate = NULL,
+        .user_data = fp
+    };
+    mp4ff_t *mp4 = mp4ff_open_read_coveronly(&cb);
+    if (!mp4) {
+        deadbeef->fclose (fp);
+        return -1;
+    }
+
+    mp4ff_cover_art_t* art_list = mp4ff_cover_get (mp4);
+    // FIXME: What shall we do if we have multiple covers in the same file?
+    if (!art_list->tail) {
+        ret = -1;
+    } else {
+        uint32_t sz = art_list->tail->size;
+        char* image_blob = art_list->tail->data;
+        trace (stderr, "will write mp4 cover art (%u bytes) into %s\n", sz, outname);
+        write_file (outname, image_blob, sz);
+    }
+
+    mp4ff_close (mp4);
+    deadbeef->fclose (fp);
+    return ret;
+}
+
+
 static int
 web_lookups (const char *artist, const char *album, const char *cache_path)
 {
@@ -1663,6 +1714,12 @@ process_query (const cover_query_t *query)
         // try to load embedded from apev2
         trace ("trying to load artwork from apev2 tag for %s\n", query->fname);
         if (!apev2_extract_art (query->fname, cache_path)) {
+            return 1;
+        }
+
+        // try to load embedded from mp4
+        trace ("trying to load artwork from mp4 tag for %s\n", query->fname);
+        if (!mp4_extract_art (query->fname, cache_path)) {
             return 1;
         }
     }
--- a/plugins/artwork/Makefile.am
+++ b/plugins/artwork/Makefile.am
@@ -31,6 +31,6 @@ FLAC_DEPS=$(FLAC_LIBS)
 flac_cflags=-DUSE_METAFLAC $(FLAC_CFLAGS)
 endif
 
-AM_CFLAGS = $(CFLAGS) $(ARTWORK_CFLAGS) $(flac_cflags) $(artwork_net_cflags) $(ogg_def) -std=c99
-artwork_la_LIBADD = $(LDADD) $(ARTWORK_DEPS) $(FLAC_DEPS) $(ogg_libs)
+AM_CFLAGS = $(CFLAGS) $(ARTWORK_CFLAGS) $(flac_cflags) $(artwork_net_cflags) $(ogg_def) -DUSE_TAGGING -I@top_srcdir@/plugins/libmp4ff -std=c99
+artwork_la_LIBADD = $(LDADD) $(ARTWORK_DEPS) $(FLAC_DEPS) $(ogg_libs) ../libmp4ff/libmp4ff.a
 endif


xinyazhang:
This should close #1543

plugins/artwork/artwork.c
```diff
@@ -1507,6 +1508,56 @@ apev2_extract_art (const char *fname, const char *outname) {
...
+    const char* image_blob;
+    uint32_t sz = mp4ff_meta_get_coverart (mp4, &image_blob);
+    if (!image_blob) {
+        ret = -1;
+        goto closemp4;
+    }
+    trace (stderr, "will write mp4 cover art (%u bytes) into %s\n", sz, outname);
+    write_file (outname, image_blob, sz);
+
+closemp4:
```
Alexey-Yakovenko: When a 20-line function has 2 gotos -- something is suspicious about it.

That's one concern.. another concern being that you're using stdio file API, while you should be using the deadbeef's VFS, which supports reading from archives, etc.

plugins/libmp4ff/mp4ff.h
```diff
@@ -102,7 +102,17 @@ int mp4ff_meta_get_totaltracks(const mp4ff_t *f, char **value);
 int mp4ff_meta_get_totaldiscs(const mp4ff_t *f, char **value);
 int mp4ff_meta_get_compilation(const mp4ff_t *f, char **value);
 int mp4ff_meta_get_tempo(const mp4ff_t *f, char **value);
-int32_t mp4ff_meta_get_coverart(const mp4ff_t *f, char **value);
```
Alexey-Yakovenko: Instead of changing the existing public API, you should add a new API.

plugins/libmp4ff/mp4ff.h
```diff
@@ -102,7 +102,17 @@ int mp4ff_meta_get_totaltracks(const mp4ff_t *f, char **value);
...
+
+/*
+ * CAVEAT:
+ * 	1. Cover art MUST be retrived by this function, to avoid strdup in
+ * 	mp4ff_meta_find_by_name
```
Alexey-Yakovenko: The above comment is kind pointless.

plugins/libmp4ff/mp4ff.h
```diff
@@ -102,7 +102,17 @@ int mp4ff_meta_get_totaltracks(const mp4ff_t *f, char **value);
...
+ * 	2. This function now returns the size of the image data, instead of 0/-1
+ * 	values proposed by the original libmp4ff.
```
Alexey-Yakovenko: This comment should go as documentation to the new API, which you should add.

plugins/libmp4ff/mp4ff.h
```diff
@@ -102,7 +102,17 @@ int mp4ff_meta_get_totaltracks(const mp4ff_t *f, char **value);
...
+ *
+ * Note, user should not modify the data pointed by *value
```
Alexey-Yakovenko: This comment is unnecessary, because `const char **` implies this.

plugins/libmp4ff/mp4meta.c
```diff
@@ -36,36 +36,67 @@
 #include "mp4ffint.h"
 
 
-
-static int32_t mp4ff_tag_add_field(mp4ff_metadata_t *tags, const char *item, const char *value)
```
Alexey-Yakovenko: Why not keep this function, and add another one with a different name? This would have resulted in less code changes.

Alexey-Yakovenko: It's hard to say for sure, but it looks like you renamed this function to `mp4ff_tag_add_string_field`, and reimplemented it using the new function mp4ff_tag_add_blob_field_move.

Why did you need to rename this one, if its behavior didn't really change?

plugins/libmp4ff/mp4meta.c
```diff
@@ -36,36 +36,67 @@
...
+
+    if (!ret) {
+        free(itemcopy);
+        free(valuecopy);
+    }
+
+    return ret;
+}
+
+#if 0 /* Nobody are using it, but keep the code for safe */
```
Alexey-Yakovenko: Looks like you had to comment this out because of the function rename above?

This API is not used, I think, because writing mp4 tags is disabled (for now), but if I need to enable it, which I plan to do soon, it would appear broken?

plugins/libmp4ff/mp4meta.c
```diff
@@ -36,36 +36,67 @@
...
+/*
+ * Postcondition:
+ * 	+ tags->tags are set to NULL if realloc failed
+ * 	+ *pitem and *pvalue are set to NULL if assignment successes
+ * 		- otherwise they are untouched.
+ * 	+ tags->tags[tags->count].item and .value are set to *pitem and *pvalue if
+ * 		assignment succcess
+ * 		- otherwise they are set to NULL
+ */
+static int32_t mp4ff_tag_add_blob_field_move(mp4ff_metadata_t *tags, char **pitem, char **pvalue, uint64_t size)
```
Alexey-Yakovenko: The interface of this function is very confusing, and undocumented, as well as the name and behavior.

I don't see any real need for having to pass `char **` pointers,

plugins/libmp4ff/mp4meta.c
```diff
@@ -36,36 +36,67 @@
...
     }
+
+    *pvalue = NULL;
+    *pitem = NULL;
+
+    tags->count++;
+    return 1;
 }
 
+static int32_t mp4ff_tag_add_string_field(mp4ff_metadata_t *tags, const char *item, const char *value)
```
Alexey-Yakovenko: This should be renamed back to mp4ff_tag_add_field.

plugins/libmp4ff/mp4meta.c
```diff
@@ -283,19 +316,33 @@ static int32_t mp4ff_parse_tag(mp4ff_t *f, const uint8_t parent_atom_type, const
...
+                    datasize = subsize-(header_size+8);
+                    free(data);
+                    data = malloc(datasize);
+                    if ((uint32_t)mp4ff_read_data(f, data, (uint32_t)datasize) != (uint32_t)datasize) {
+                        free(data);
+                        data = NULL;
+                        datasize = 0;
+                    }
+                } else {
+                    if (data) {free(data);data = NULL;}
```
Alexey-Yakovenko: please put `free(data)` and `data=NULL` on separate lines.

plugins/libmp4ff/mp4meta.c
```diff
@@ -314,10 +361,21 @@ static int32_t mp4ff_parse_tag(mp4ff_t *f, const uint8_t parent_atom_type, const
 	{
 		if (!done)
 		{
-			if (name == NULL) mp4ff_set_metadata_name(f, parent_atom_type, &name);
-			if (name) mp4ff_tag_add_field(&(f->tags), name, data);
+            /* Some files tagged by iTunes 10.1 have a name atom with empty string. */
```
Alexey-Yakovenko: The block below seems to have its indentation messed-up

plugins/libmp4ff/mp4meta.c
```diff
@@ -314,10 +361,21 @@ static int32_t mp4ff_parse_tag(mp4ff_t *f, const uint8_t parent_atom_type, const
...
+             *
+             * However, data is not used elsewhere (freed immediately), thus
+             * copy can be done by move.
+              */
+            if (name) {
+                //fprintf(stderr, "[mp4ff] Assign %s (atom id: %d) to %s\n", name, (int)parent_atom_type, data);
+                mp4ff_tag_add_blob_field_move(&(f->tags), &name, &data, datasize);
+            }
 		}
-
```
Alexey-Yakovenko: Please don't re-format the existing code, when it's not changing otherwise.

plugins/libmp4ff/mp4meta.c
```diff
@@ -226,6 +258,7 @@ static int32_t mp4ff_parse_tag(mp4ff_t *f, const uint8_t parent_atom_type, const
     char * name = NULL;
 	char * data = NULL;
 	uint32_t done = 0;
+    uint64_t datasize = 0;
```
Alexey-Yakovenko: The `datasize` value gets calculated from uint32_t variables below, so there's no need to use 64 bit type.

plugins/libmp4ff/mp4meta.c
```diff
@@ -283,19 +316,33 @@ static int32_t mp4ff_parse_tag(mp4ff_t *f, const uint8_t parent_atom_type, const
...
-				}
+                } else if (parent_atom_type == ATOM_COVER) {
+                    /*
+                      * Cover data is basically a binary blob, which may include \0 in the middle,
+                     * and it also doesn't need \0 at the end
+                     */
+                    datasize = subsize-(header_size+8);
+                    free(data);
+                    data = malloc(datasize);
+                    if ((uint32_t)mp4ff_read_data(f, data, (uint32_t)datasize) != (uint32_t)datasize) {
```
Alexey-Yakovenko: You would not need the casts, if you used uint32_t for the datasize.

plugins/libmp4ff/mp4meta.c
```diff
@@ -459,9 +517,22 @@ int32_t mp4ff_meta_get_tempo(const mp4ff_t *f, char **value)
     return mp4ff_meta_find_by_name(f, "tempo", value);
 }
 
-int32_t mp4ff_meta_get_coverart(const mp4ff_t *f, char **value)
+uint64_t mp4ff_meta_get_coverart(const mp4ff_t *f, const char **value)
```
Alexey-Yakovenko: With the current code, uint64_t type is overkill.

plugins/libmp4ff/mp4ffint.h
```diff
@@ -154,10 +154,12 @@ typedef struct
 
 
 /* metadata tag structure */
+/* we need the size field because the value may not be a blob instead of a C string */
 typedef struct
 {
     char *item;
     char *value;
+    uint64_t size;
```
Alexey-Yakovenko: This needs to be uint32_t, because none of the code in mp4ff supports atoms larger than 32 bits allow, and I think mp4 format itself doesn't support that either.

Alexey-Yakovenko:
I wrote all of my concerns, but it's quite likely that I will have more questions after everything above gets fixed.

Alexey-Yakovenko:
I looked more at the code, and the whole thing about the cover atom. It doesn't really belong to metadata.

I think the right thing would be to store the cover data as a separate thing in the `mp4ff_t`, instead of `mp4ff_metadata_t`.

And keep the data in mp4ff_metadata_t string-only.

If we start adding binary blobs in there -- that would complicate working with the data in the input plugins. We'd need some way to know whether it's a string or not.

Then we need to see what happens when there are multiple cover atoms in the file.

Alexey-Yakovenko:
Another thing to remember: the cover blobs are often large, and take CPU and RAM to load. That's especially important when adding files to playlist, and playing them back -- we need to make sure that the covers DON'T load during insert/play.

Therefore, we should have a flag, which indicates whether we're loading cover, or not.

That might be tricky, but should be doable by changing how the `parse_atoms_int` function works, and/or making a new API that loads just the cover, but skips everything else.

Alexey-Yakovenko:
I have added the code necessary for loading multiple cover images, storing them in a linked list, and accessing them. Please see c40b5f832b7b87cf6abe7f6e0688c1e129e55344

The next step would be to add the code to use this from artwork plugin.

plugins/libmp4ff/mp4meta.c
xinyazhang: Yes you're right.
The point is, it looks like the original author assumes every tag field is a C string, which is not true esp. for cover arts, and this assumption is the reason why mp4ff can't handle embedded cover arts.

So here I renamed add_field to add_string_field to emphasize this assumption for tag fields.

plugins/libmp4ff/mp4meta.c
xinyazhang: I commented out this one simply for code clean up, because it's a static function and no code are using it.

plugins/libmp4ff/mp4meta.c
xinyazhang: It's not needed but good for semantics imo.
After a successful assignment the ownership of *pitem and *pvalue is transferred to tags, so it's better to reset the original point to avoid unintended sharing.

plugins/libmp4ff/mp4ff.h
xinyazhang: The problem is the existing public API has no point to work.
We need the length of the embedded image file, and getting the length will require the API returns the length in someway.

Therefore either we got a broken unnamed one, or a fixed one with the same ABI but different semantics, or a fixed one with ABI changed.

In my view the last one is better because it enforces mp4ff users to change the code accordingly instead of leaving an unfixed bug.

plugins/libmp4ff/mp4meta.c
xinyazhang: Err, I just left the original mp4ff code there (with tabs replaced by spaces)
However I'll fix this anyway.

plugins/libmp4ff/mp4meta.c
xinyazhang: Sure, mp4ff uses tabs and spaces in a mixed way.
Maybe I should send another commit to replace all tabs with spaces before?

plugins/libmp4ff/mp4meta.c
xinyazhang: No, datasize is calculated from subsize, which is uint64_t.

plugins/libmp4ff/mp4ffint.h
xinyazhang: MP4 does support it and the code can be found in mp4ff_atom_read_header.
Although most of mp4ff is implemented as "just works" without full compliance with MP4 specs.

Alexey-Yakovenko:
@xinyazhang you shouldn't be fixing anything related to mp4ff in your PR (please read the latest comments first). I wrote a different code, which doesn't require changing the mp4meta APIs or data structures, with some explanation why it's done this way.

Now we only need to add the necessary code to artwork plugin

Alexey-Yakovenko:
@xinyazhang I'm sorry, I was wrong about the 64 bit stuff - seems like mp4ff supports 64 bit offests correctly (needs to be tested though). However, I don't see any real need to use 64 bit for storing cover art file size.

xinyazhang:
Err, I don't think I can use 329c434 directly.
The problem is the tags are parsed and loaded when user calls mp4ff_open_read(_metaonly), when we have not chance to call mp4ff_set_load_covers.

How about recording the offsets and lengths during open and add an API to get the covers later?

Alexey-Yakovenko:
hmm. I'll have a look, and get back to you. missed this part somehow.

Alexey-Yakovenko:
How about 27ce7482667a3e9035220537c8682095bcdcd319?
Now you can call mp4ff_open_read_coveronly.

For now, it reads metadata + covers, but can be improved later.

xinyazhang:
Thanks, this should work.

xinyazhang:
Updates pushed.

By the way, do you have plan to remove the "cover extraction" step for local files?
I think we can let the UI read the embedded cover data directly, instead of relaying though a temporary cache file.

plugins/artwork/artwork.c
```diff
@@ -1507,6 +1508,56 @@ apev2_extract_art (const char *fname, const char *outname) {
...
+        .user_data = fp
+    };
+    mp4ff_t *mp4 = mp4ff_open_read_coveronly(&cb);
+    if (!mp4) {
+        deadbeef->fclose (fp);
+        return -1;
+    }
+
+    mp4ff_cover_art_t* art_list = mp4ff_cover_get (mp4);
+    // FIXME: What shall we do if we have multiple covers in the same file?
```
Alexey-Yakovenko: Right now we only support 1 cover per file. Please don't put comments like this in the code -- that's not the place for the questions.

Alexey-Yakovenko:
The code looks ok, I'll merge, but seems it's failing the build. I don't know why. Did you merge origin/master before applying your changes?

Alexey-Yakovenko:
> By the way, do you have plan to remove the "cover extraction" step for local files?

Yeah, I kinda have a "plan" to make a new cover art plugin with a better API, which could use the covers directly from local files/folders, without copying them. But I don't know when I'm gonna start working on this.

Alexey-Yakovenko:
we should possibly add an extra file type check (mp4, m4a, m4b, etc) to avoid attempting loading album art from any files. attempting to load every file can be VERY slow.

xinyazhang:
> Did you merge origin/master before applying your changes?
Yes I did rebase (otherwise I couldn't test locally).

> we should possibly add an extra file type check (mp4, m4a, m4b, etc) to avoid attempting loading album art from any files
It's not a big deal for performance, since mp4ff will failed to open instantly after the first unknown atom, but you're right, we should check this rather than offloading to mp4ff.

Alexey-Yakovenko:
> Yes I did rebase (otherwise I couldn't test locally).

the issue was with the -I order in AM_CFLAGS.. It was using the system libmp4ff, and failing (obviously)

> It's not a big deal for performance, since mp4ff will failed to open instantly after the first unknown atom

yeah, but it will still open the file, and try reading data from it, causing memory allocations, then clean up, etc -- that's a lot of operations, and they all add up.

PR #1558, opened 2016-04-27
gtkui: add copy/paste support

cboxdoerfer:
This is quite a big patch so I don't expect it to get merged right away.

What this PR does: It adds copy, cut and paste support to the ddblistview, the pltbrowser and the pltmenu. C&P works both from DeaDBeeF to file managers and the other way around, and of course within DeaDBeeF (copy track(s) in playlist1 and paste it/them in playlist2). Pasting tracks to the pltbrowser first creates a new playlist and then pastes the tracks there. C&P can be triggered either via menu items or via shortcuts (ctrl+c, ctrl+x, ctrl+v).

To make C&P available to the pltbrowser plugin I added an API to the gtkui plugin:
https://github.com/Alexey-Yakovenko/deadbeef/compare/master...cboxdoerfer:copy_paste?expand=1#diff-31721c7ad41dcfea0c8abdd8072c53e9R263

Hope I can get any feedback on that. I also thought about making a proper GObject, e.g. DdbClipboard, but found that to much overhead, with little benefit.

### DIFF
--- a/plugins/gtkui/clipboard.c
+++ b/plugins/gtkui/clipboard.c
@@ -73,6 +73,7 @@ clipboard_free_clipboard_data ()
         }
         if (current_clipboard_data->plt_title) {
             free (current_clipboard_data->plt_title);
+            current_clipboard_data->plt_title = NULL;
         }
         current_clipboard_data->num_tracks = 0;
         current_clipboard_data->cut = 0;
@@ -149,8 +150,6 @@ clipboard_clear_clipboard_data (GtkClipboard *clipboard,
             for (int i = 0; i < clip_ctx->num_tracks; i++) {
                 if (clip_ctx->tracks[i]) {
                     deadbeef->pl_item_unref (clip_ctx->tracks[i]);
-                    if (clip_ctx->tracks[i]) {
-                    }
                 }
             }
             free (clip_ctx->tracks);
--- a/plugins/gtkui/clipboard.c
+++ b/plugins/gtkui/clipboard.c
@@ -43,6 +43,7 @@ typedef struct {
 } clipboard_data_context_t;
 
 static clipboard_data_context_t *current_clipboard_data = NULL;
+static int current_clipboard_refcount = 0;
 
 enum {
     DDB_URI_LIST = 1,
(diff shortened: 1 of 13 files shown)


plugins/gtkui/clipboard.c
```diff
@@ -0,0 +1,548 @@
...
+   track gets deleted before a paste happend the data of the playlist item
+   might already be freed. So we have to copy that data in advance, which means we
+   have two copies of the same data in memory.
+
+   Of course we could fix that by storing the tracks' path instead and handle
+   a paste just like adding files or folders to a playlist. However this is quite slow
+   since we need to decode the tracks all over again.
+
+   As soon as we have a database we could improve that, because tracks would
+   be bound to the database and not the playlist and therefore be more persitent. */
```
Alexey-Yakovenko: I'm not sure if I understand it correctly, but the tracks (playItems) are reference counted, and if you do `pl_item_ref` before storing a pointer -- they are guaranteed to persist, then you can do `pl_item_unref` before quitting the player and clearing clipboard..

plugins/gtkui/clipboard.c
```diff
@@ -0,0 +1,548 @@
...
+        clipboard_write_uri_list (clip_ctx, uri_list);
+        buf = (guchar *)uri_list->str;
+        buf_len = uri_list->len + 1;
+    }
+    gtk_selection_data_set (sel, target, 8, buf, buf_len);
+    g_string_free (uri_list, TRUE);
+}
+
+static void
+clipboard_clear_clipboard_data (GtkClipboard *clipboard,
```
Alexey-Yakovenko: Is this the same as `clipboard_free_clipboard_data`?

Alexey-Yakovenko:
`clipboard_free_clipboard_data` - would look better if it's called `clipboard_free_data`, or even `clipboard_free` (depends).

plugins/gtkui/clipboard.c
cboxdoerfer: ... of course, you're right and I'm stupid. :D No idea why I haven't thought of that myself. Well at least it's easy to fix.

plugins/gtkui/clipboard.c
cboxdoerfer: Yes thx, they started out different but apparently became the same over time. Will get rid of one of them.

cboxdoerfer:
@Alexey-Yakovenko, what do you think of the gtkui plugin api? Do you think it's ok to useDDB_ACTION_CTX_* there?

Alexey-Yakovenko:
My understanding of the code is not yet good enough to tell whether it's good or bad. Let's say it's ok for now, we can change it any time.

cboxdoerfer:
Fixed the issues. I also added a ref counter for the clipboard data to make sure we do proper clean up at exit. From my understanding of the API GTK+ is supposed to do that, but it doesn't work reliably, so we do it manually if GTK+ fails.

PR #1573, opened 2016-05-10
gtkui: add support for dragging playlist items from deadbeef to other…

### LINKED ISSUE #1567: Possibility to drag&drop copy track
Hi, 
many players have the possibility to copy a track to a location by simply draging it to that location directly from the plalist (ui). That way one doesn't have to open the folder where music is stored, minimize the player, search for the song ,open the destination folder and copy the track.
 Any chance you implement this?

Thank you!

cboxdoerfer:
… applications

This should fix #1567

### DIFF
--- a/plugins/gtkui/ddblistview.c
+++ b/plugins/gtkui/ddblistview.c
@@ -1152,25 +1152,33 @@ ddb_listview_list_drag_motion                (GtkWidget       *widget,
 {
     DdbListview *pl = DDB_LISTVIEW (g_object_get_data (G_OBJECT (widget), "owner"));
     ddb_listview_list_track_dragdrop (pl, x, y);
+
+    // NOTE: we need to check whether the source supports TARGET_PLAYITEMS
+    // in order to know which GdkDragAction to use
     GList *targets = gdk_drag_context_list_targets (drag_context);
+    gboolean source_is_listview = FALSE;
     int cnt = g_list_length (targets);
-    int i;
-    for (i = 0; i < cnt; i++) {
+    for (int i = 0; i < cnt; i++) {
         GdkAtom a = GDK_POINTER_TO_ATOM (g_list_nth_data (targets, i));
         gchar *nm = gdk_atom_name (a);
-        if (!strcmp (nm, "text/uri-list")) {
+        if (!strcmp (nm, TARGET_PLAYITEMS)) {
+            source_is_listview = TRUE;
             g_free (nm);
             break;
         }
         g_free (nm);
     }
-    if (i != cnt) {
+
+    if (!source_is_listview) {
+        // source doesn't support TARGET_PLAYITEMS, so it's probably a file manager
+        // and GDK_ACTION_MOVE wouldn't make any sense: only support GDK_ACTION_COPY
         gdk_drag_status (drag_context, GDK_ACTION_COPY, time);
     }
     else {
+        // source is listview, so we can support both GDK_ACTION_MOVE and GDK_ACTION_COPY
         GdkModifierType mask;
 
-    GdkWindow *win = gtk_widget_get_window (widget);
+        GdkWindow *win = gtk_widget_get_window (widget);
 #if GTK_CHECK_VERSION(3,0,0)
         GdkDeviceManager *device_manager = gdk_display_get_device_manager (gdk_window_get_display (win));
         GdkDevice *pointer = gdk_device_manager_get_client_pointer (device_manager);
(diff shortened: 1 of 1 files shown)


plugins/gtkui/ddblistview.c
```diff
@@ -1200,6 +1208,47 @@ ddb_listview_list_drag_drop                  (GtkWidget       *widget,
...
+    }
+    gchar **uri_list = g_new0 (gchar *, num_selected + 1);
+    if (!uri_list) {
+        return NULL;
+    }
+    // NOTE: hash table is used to detect duplicates since we don't want to
+    // copy files more than once
+    GHashTable *table = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
+
+    deadbeef->pl_lock ();
```
Alexey-Yakovenko: This pl_lock should be moved up, to before the plt_get_sel_count call, otherwise selection could change before the playlist is locked.

plugins/gtkui/ddblistview.c
```diff
@@ -1200,6 +1208,47 @@ ddb_listview_list_drag_drop                  (GtkWidget       *widget,
...
+
+    DdbListviewIter it = deadbeef->plt_get_head (ps->drag_source_playlist);
+    int idx = 0;
+    while (it) {
+        if (ps->binding->is_selected (it)) {
+            const char *path = deadbeef->pl_find_meta (it, ":URI");
+            if (!g_hash_table_lookup (table, path)) {
+                // new track, add to hash table and uri list
+                gchar *key = g_strdup (path);
```
Alexey-Yakovenko: I can't see where this allocation gets freed.

plugins/gtkui/ddblistview.c
```diff
@@ -1200,6 +1208,47 @@ ddb_listview_list_drag_drop                  (GtkWidget       *widget,
...
+    // copy files more than once
+    GHashTable *table = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
+
+    deadbeef->pl_lock ();
+
+    DdbListviewIter it = deadbeef->plt_get_head (ps->drag_source_playlist);
+    int idx = 0;
+    while (it) {
+        if (ps->binding->is_selected (it)) {
+            const char *path = deadbeef->pl_find_meta (it, ":URI");
```
Alexey-Yakovenko: `deadbeef->is_local_file` should be used here to determine if the URI is pointing to a local file, that can actually be copied.

I think that GNOME/GIO/GVFS/... might be capable of copying remote files too, but I'm not sure that's what we want.

That check would also make sure you're not trying to copy archived files from e.g. a zip archive.

Alexey-Yakovenko: hmm.. I just read in the comment about the is_local_file, that it returns 1 for VFS files.. so another check would have to be added for that case, but I can't remember what's the simplest way to do it..

plugins/gtkui/ddblistview.c
cboxdoerfer: Line 1270: g_strfreev
https://github.com/Alexey-Yakovenko/deadbeef/pull/1573/files#diff-24cb6c48412b8580aa4ea844537ea634R1270

plugins/gtkui/ddblistview.c
cboxdoerfer: Sure, I'll have a look.

plugins/gtkui/ddblistview.c
cboxdoerfer: Oh, sorry wrong string. That's the key for the hash table and gets removed together with the table here:
https://github.com/Alexey-Yakovenko/deadbeef/pull/1573/files#diff-24cb6c48412b8580aa4ea844537ea634R1247

plugins/gtkui/ddblistview.c
cboxdoerfer: @Alexey-Yakovenko, do normal local files always have a simple path style string in the :URI tag? i.e. a string starting with '/'? Or is it possible that they also start with the URI scheme 'file://'? If that's the case we could simply check for that and ignore all tracks that start with a valid URI scheme.

Alexey-Yakovenko: @cboxdoerfer  I would not count on that.

But it's safe to assume the file is really local, if it starts with `/` or `file://`

plugins/gtkui/ddblistview.c
cboxdoerfer: Fixed

plugins/gtkui/ddblistview.c
cboxdoerfer: Fixed.

PR #1619, opened 2016-08-03
Fix build failures on armv6

Barthalion:
This is the patch used for Alpine Linux to make deadbeef build on armv6
with hard-float ABI. Narrowing issues come from the fact that char on
ARM is unsigned by default, and GCC6 treats this as an error.

(As I'm not portability expert, it would be nice if someone wiser could take a look at this change.)

### DIFF
--- a/plugins/adplug/adplug/s3m.cpp
+++ b/plugins/adplug/adplug/s3m.cpp
@@ -26,7 +26,7 @@
 #include <string.h>
 #include "s3m.h"
 
-const char Cs3mPlayer::chnresolv[] =	// S3M -> adlib channel conversion
+const signed char Cs3mPlayer::chnresolv[] =	// S3M -> adlib channel conversion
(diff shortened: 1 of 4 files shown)


oleksiy-unity3d:
Have you tested that adlib S3Ms and SIDs play with these changes, especially on non-arm builds?

oleksiy-unity3d:
It's weird that this fails on ARM GCC, cause I'm building the same code for android with both GCC and CLANG, and no failures there.

I'm not even sure that I had to tweak CFLAGS...

Seems like I had to use -fno-signed-zeros for libsidplay2 when building with -mfpu=neon, but I'm not sure that's your case.

Alexey-Yakovenko:
Have you tested that adlib S3Ms and SIDs play with these changes, especially on non-arm builds?

It's weird that this fails on ARM GCC, cause I'm building the same code for android with both GCC and CLANG, and no failures there.

I'm not even sure that I had to tweak CFLAGS...

Seems like I had to use -fno-signed-zeros for libsidplay2 when building with -mfpu=neon, but I'm not sure that's your case.

Alexey-Yakovenko:
@Barthalion ^^

Alexey-Yakovenko:
Rejecting pull request because no contact with the author.

PR #1625, opened 2016-08-15
Avoid mktemp and use mkstemp instead

Vascom:
As described in **man 3 mktemp**:
**Never  use mktemp()**.  Some implementations follow 4.3BSD and replace XXXXXX by the current process ID and a single letter, so that at most 26 different names can be returned.  Since on the one hand the names are easy to guess, and on  the  other  hand there  is a race between testing whether the name exists and opening the file, every use of mktemp() is a security risk.  The race is avoided by mkstemp(3) and mkdtemp(3).

### DIFF
--- a/plugins/converter/converter.c
+++ b/plugins/converter/converter.c
@@ -919,7 +919,7 @@ convert (DB_playItem_t *it, const char *out, int output_bps, int output_is_float
                     tmp = "/tmp";
                 }
                 snprintf (input_file_name, sizeof (input_file_name), "%s/ddbconvXXXXXX", tmp);
-                char *res = mktemp (input_file_name);
+                char *res = mkstemp (input_file_name);
                 strcat (input_file_name, ".wav");
             }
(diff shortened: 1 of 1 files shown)


Alexey-Yakovenko:
Your change has a syntax error.

Vascom:
You mean change **char *res** to **int res**?

Alexey-Yakovenko:
Yeah, you should maybe change the line to this:

```
(void)mkstemp (input_file_name);
```

Alexey-Yakovenko:
I looked at the manpage for mkstemp, and at the code that you suggested to change.

mkstemp works differently from mktemp. It returns an open file descriptor.

So it's not a direct replacement for the mktemp.

In the particular case of converter, the filename produced by mktemp is passed to another process, and we can't do the same with a file descriptor, so your change would simply break the code.

Therefore, I can't accept this change. Really sorry.

PR #1632, opened 2016-08-26
Title Formatting in the encoder command line

thesame:


Alexey-Yakovenko:
It's arguable whether `%playlist%` should be a `%value%` or a `$function`, but it should definitely be named differently (e.g. `playlist_title`).

We need to discuss this, and find the right thing.

Alexey-Yakovenko:
$shesc is undocumented and has a weird cryptic name.

Alexey-Yakovenko:
Before I could merge any of this, I need to write unit tests for each new TF feature, and make sure everything works, but that's just informational.

Alexey-Yakovenko:
Converter change: only briefly looked at this, but without extensive testing -- I don't believe that this will work correctly.

tf.c
```diff
@@ -1709,6 +1709,11 @@ tf_eval_int (ddb_tf_context_t *ctx, const char *code, int size, char *out, int o
                         }
                     }
                 }
+                else if (!strcmp (name, "playlist")) {
+                    char pltname[2000];
+                    deadbeef->plt_get_title (ctx->plt, pltname, sizeof (pltname));
+                    val = pltname;
```
Alexey-Yakovenko: This code assigns val to a stack local buffer pointer, which points to a memory which will get reused upon exiting the scope. It's strange that you didn't get a compiler warning on that.

a better way to write this:

```
val = ctx->plt->name;
```

because tf is part of the core, and can access internal stuff.

Alexey-Yakovenko:
There's no code in this PR, that can be accepted as-is, so closing it for now.

PR #1634, opened 2016-08-30
converter: build fix

thesame:


Alexey-Yakovenko:
converter fix seems to be correct, but the TF change seems to be wrong (won't compile).
I'll apply the fix manually.

PR #1638, opened 2016-09-04
add OpenBSD sndio backend

zmyrgel:
Add sndio audio backend from OpenBSD port.
Merging this would help reduce patching needed for deadbeef in OpenBSD.


Alexey-Yakovenko:
@zmyrgel 
The build fails on linux

```
sndio.c:23:19: fatal error: sndio.h: No such file or directory
compilation terminated.
```

You need to add the logic to configure.ac, to exclude the sndio plugin on systems which don't support it.

Another thing of importance.. While I can accept this plugin into the main line, I won't be able to provide any support for it, and I could not even validate that it works correctly.

Why is this plugin even needed? Doesn't OpenBSD have OSS?

zmyrgel:
Sndio is the audio framework used in OpenBSD, see http://www.sndio.org/ for details.
I've added check now and made sndio backend disabled by default.
It could be tweaked later on to be enabled automatically if detected but I think this would work for starters.

plugins/sndio/Makefile.am
```diff
@@ -0,0 +1,6 @@
+sndiodir = $(libdir)/$(PACKAGE)
+pkglib_LTLIBRARIES = sndio.la
+sndio_la_SOURCES = sndio.c
+sndio_la_LDFLAGS = -module -avoid-version
+sndio_la_LIBADD = $(LDADD) $(SNDIO_LIBS)
```
Alexey-Yakovenko: I think that this would not work unless you add the following line to configure.ac:

`AC_SUBST(SNDIO_LIBS)`

Alexey-Yakovenko:
I still don't like that it's disabled instead of being auto-configured.

Every other library in configure.ac is autodetected. All you need to add is something like this:

```
AC_CHECK_LIB([sndio], [main], [have_sndio=yes], [have_sndio=no])
AS_IF([test "${have_sndio}" = "yes"], [
    SNDIO_LIBS="-lsndio"
    AC_SUBST(SNDIO_LIBS)
    HAVE_SNDIO=yes
])
```

then you need to change the `AC_ARG_ENABLE` to say `default: auto`etc.

There's nothing complicated about that -- just copy what's already used by a similar module.

Alexey-Yakovenko:
merged

PR #1639, opened 2016-09-06
convgui: a bit more verbose status display

thesame:
http://en.zimagez.com/zimage/screenshot2016-09-0523-20-51.php

### DIFF
--- a/plugins/converter/converter.glade
+++ b/plugins/converter/converter.glade
@@ -1893,4 +1893,122 @@ Temporary file</property>
   </child>
 </widget>
 
+<widget class="GtkDialog" id="procdlg">
+  <property name="visible">True</property>
+  <property name="title" translatable="yes">Converting...</property>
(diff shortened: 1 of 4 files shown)


Alexey-Yakovenko:
Stuff that needs fixing IMO:

1. All dialogs should be following the same convention about border size and distance between widgets.
2. Tracks and Skipped can be placed on the same line (remember spacing between them!)
3. I think the better naming would be "Total" instead of "Track".
4. Why is progress bar so narrow? I suppose it's GTK theme, but then I need to get this tested, or see in Glade.. Which I need to do anyway.

Alexey-Yakovenko:
This is the layout to aim for

image

We don't need to have the "skipped" in the same dialog -- it's better to generate a report, and send it to the log (yeah, we'll have the log window in GTKUI too soon :)

The percentage is pretty useless -- it duplicates what progress bar is already showing.

Alexey-Yakovenko:
PR rejected (abandoned)

PR #1640, opened 2016-09-06
Issue 1562

mitchnull:
proposed patch for issue#1562: add option to change track with the scroll-wheel on the system tray icon
new config option:
tray.scroll_changes_track  (default: 0)

- first commit adds logic to gtkui plugin
- second commit adds logic to statusnotifier plugin


Alexey-Yakovenko:
merged

Issue #618, opened 2015-03-12
Remember minimization to tray after restart


Alexey-Yakovenko:
Original [issue 716](https://code.google.com/p/ddb/issues/detail?id=716) created by Alexey-Yakovenko on 2012-04-19T11:51:38.000Z:

submitted by: nobody
Date: 2011-05-17 08:28:25

DeaDBeeF minimized to tray, but if I close it (from tray menu, or system shutdown) and start again - DeaDBeeF opens as normal window. Will be good if DeaDBeeF remember own state.

kpcee:
Done with "[gtkui: added an option to start deadbeef hidden](https://github.com/DeaDBeeF-Player/deadbeef/commit/3ac4cc03fe52356784a50836cf361eed61dcc797)".

Issue #521, opened 2015-03-12
Add console option for hidding to tray


Oleksiy-Yakovenko:
Original [issue 619](https://code.google.com/p/ddb/issues/detail?id=619) created by Alexey-Yakovenko on 2012-04-19T11:49:56.000Z:

submitted by: nobody
Date: 2010-06-12 13:59:43
## Please add console option for hidding to tray

Пожалуйста, добавьте консольный параметр для скрытия в трей

jtechpro:
Is it fixed?

Alexey-Yakovenko:
If it's fixed -- it would be displayed as "Closed".

In recent builds, it is possible to use a configuration option `gtkui.starthidden 1`.
I don't remember if there's UI for it, but you can add the line above to the configuration file.

jtechpro:
Where is that configuration file located? I'm using portable deadbeef

Alexey-Yakovenko:
`$HOME/.config/deadbeef/config`

jtechpro:
That didn't work

Alexey-Yakovenko:
That probably means that you're not running a recent build.

jtechpro:
No, I'm running 0.7.2

Alexey-Yakovenko:
Yeah, 0.7.2 is not very recent.

jtechpro:
But from the website..?

Alexey-Yakovenko:
the latest build from master is here, as well as the source snapshot:

https://sourceforge.net/projects/deadbeef/files/travis/linux/master/

jtechpro:
No still, it didn't work

jtechpro:
Now, the tray icon is also not shown

Alexey-Yakovenko:
Regarding the tray not showing: if it still shows in 0.7.2, that means XEmbed is supported on your system, but the new status notifier protocol support is not compatible with deadbeef implementation.

To solve this, please remove the statusnotifier.so from deadbeef/plugins folder.

Regarding why the gtkui.start_hidden is not working for you -- I can't help. It works here.

jtechpro:
But previously, the icon is shown.

oleksiy-unity3d:
That's because previously (0.7.2) the statusnotifier.so didn't exist.

The build I linked above is not a stable release. It's a development build with many unfinished and/or untested work.

jtechpro:
Which Desktop and OS are you using?

Alexey-Yakovenko:
[...]

PR #1641, opened 2016-09-06
add option to start in tray: tray.start_minimized

mitchnull:
partial fix for issues #618 and #521

### DIFF
--- a/plugins/gtkui/gtkui.c
+++ b/plugins/gtkui/gtkui.c
@@ -1168,6 +1168,9 @@ gtkui_thread (void *ctx) {
     gtkui_is_retina = is_retina (mainwin);
 #endif
 
+    if (deadbeef->conf_get_int ("tray.start_minimized", 0)) {
+        gtk_widget_hide (mainwin);
+    }
     gtk_main ();
 
     deadbeef->unlisten_file_added (fileadded_listener_id);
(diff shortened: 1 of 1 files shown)


Alexey-Yakovenko:
I think that the configuration variable name is wrong, because:

1. It doesn't have anything to do with the tray / notification area.
2. It doesn't start the player minimized. In fact, it doesn't even start it hidden. Instead, it hides the window after it's been shown.

However, if we just keep the window hidden, i.e. not call gtk_widget_show (mainwin) -- this might cause all sorts of problems as well..

So at the very least, it needs to be called "gtkui.start_hidden".

Then, this needs a QA pass. We need to make sure that the following still works:

* the saved layout is correctly restored on startup
* columns have correct sizes in "column autoresize" mode

mitchnull:
Agreed... 

In fact, despite my initial tests working, now I get IA__gdk_window_get_state: assertion 'GDK_IS_WINDOW (window)' failed errors all over the place...

Please disregard this patch, I'll try to find a more reliable way...

mitchnull:
Changed the option name and used g_idle_add() to avoid the GTK assertions.  Seems to be working fine now.

oleksiy-unity3d:
Right, using g_idle_add should make sure that gtk_widget_hide will be called after all of the initial things taking place (like configure/realize), so there's a chance that layout and column sizes will be initialized correctly now.

Alexey-Yakovenko:
Right, using g_idle_add should make sure that gtk_widget_hide will be called after all of the initial things taking place (like configure/realize), so there's a chance that layout and column sizes will be initialized correctly now.

PR #1645, opened 2016-09-09
deadbeef.h: add logging api with va_list parameters

hyperblast:
This would allow more convenient integration with existing code that already uses variable arguments functions.

Compile checked on Linux, should I add some tests?

I've found this line in existing code suspicious a bit:

> (void) vsnprintf(text, 128, fmt, ap);

What is 128? Was it meant to be sizeof (text)?

### DIFF
--- a/logger.c
+++ b/logger.c
@@ -72,12 +72,32 @@ ddb_log_detailed (DB_plugin_t *plugin, uint32_t layers, const char *fmt, ...) {
 }
 
 void
+ddb_vlog_detailed (DB_plugin_t *plugin, uint32_t layers, const char *fmt, va_list ap) {
+    if (!_is_log_visible(plugin, layers)) {
+        return;
+    }
+
+    char text[2048];
+    (void) vsnprintf(text, sizeof (text), fmt, ap);
+
+    _log_internal (plugin, layers, text);
+}
+
+void
 ddb_log (const char *fmt, ...) {
     char text[2048];
     va_list ap;
     va_start(ap, fmt);
-    (void) vsnprintf(text, 128, fmt, ap);
+    (void) vsnprintf(text, sizeof (text), fmt, ap);
     va_end(ap);
+
+    _log_internal (NULL, 0, text);
+}
+
+void
+ddb_vlog (const char *fmt, va_list ap) {
+    char text[2048];
+    (void) vsnprintf(text, sizeof (text), fmt, ap);
 
     _log_internal (NULL, 0, text);
 }
--- a/deadbeef.h
+++ b/deadbeef.h
@@ -29,6 +29,7 @@
 #include <time.h>
 #include <stdio.h>
 #include <dirent.h>
+#include <stdarg.h>
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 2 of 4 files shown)


Alexey-Yakovenko:
Can you provide an example, which code this would integrate with?

logger.c
```diff
@@ -83,6 +95,14 @@ ddb_log (const char *fmt, ...) {
 }
 
 void
+ddb_vlog (const char *fmt, va_list ap) {
+    char text[2048];
+    (void) vsnprintf(text, 128, fmt, ap);
```
Alexey-Yakovenko: This doesn't seem right..

Alexey-Yakovenko:
> What is 128? Was it meant to be sizeof (text)?

This is a typo.. not critical, but yeah - it should be the size of the buffer.

hyperblast:
Hi!

I'm writing a relatively large plugin that is designed to support multiple players.

I have simple logging functions, i.e.

> void my_log_debug(const char* fmt, ...);
> void my_log_error(const char* fmt, ...);

These functions wrap logging APIs provided by the target player.
Current implementation for DeaDBeeF uses vfprintf to write to stderr:

> va_list va;
> va_start(va, fmt);
> vfprintf(stderr, fmt, va);
> va_end(va);

I wanted to use new logging API when it is released.
Since there is no dependency on deadbeef.h in common code,
I can't use macros to forward calls to log_detailed.

I could format to a temporary buffer first and then call log_detailed(..., "%s", temp);
But this would mean wasted work if actual message is not going to be written anywhere.
vlog_detailed solves this problem.

logger.c
hyperblast: As we've already found this is not a magic constant and should be replaced with sizeof(text);
I'll provide an update.

Alexey-Yakovenko:
ok, I will accept the patch after you fix the error with number 128.

(I'm out for the next 3 weeks though)

hyperblast:
I've fixed magic number issue, but somehow travis didn't like this change.
It seems that this failure is not related to my patch.
The fix is trivial and previous version was building fine.
Also it builds fine on my machine.
Anyway have a nice vacation.

logger.c
```diff
@@ -72,17 +72,37 @@ ddb_log_detailed (DB_plugin_t *plugin, uint32_t layers, const char *fmt, ...) {
...
+}
+
+void
 ddb_log (const char *fmt, ...) {
     char text[2048];
     va_list ap;
     va_start(ap, fmt);
-    (void) vsnprintf(text, 128, fmt, ap);
+    (void) vsnprintf(text, sizeof (text), fmt, ap);
     va_end(ap);
```
Alexey-Yakovenko: It seems that now this code should be refactored to simply call ddb_vlog_detailed, instead of duplicating the vsnprintf code etc.

I.e. it may become this:

```
void
ddb_log (const char *fmt, ...) {
    va_list ap;
    va_start(ap, fmt);
    ddb_vlog_detailed (NULL, 0, fmt, ap);
    va_end(ap);
}
```

what do you think?

Alexey-Yakovenko:
Yeah, it seems that travis failure is not related to your changes.

logger.c
hyperblast: I was thinking about moving `vsnprintf` call to `_log_internal` to have formatting logic in single place.

Also probably `_is_log_visible` should be available in the public API to cover heavy computation case e.g.:

> if (ddb->is_log_visible(...))
> {
>    // some heavy computation
>    ddb->log_detailed(...);
> }

Alexey-Yakovenko: There is no heavy computation in there

hyperblast: Sorry, I wasn't clear. I mean if plugin needs to compute something before writing to log, it might be useful to check if this message is going to be logged. So having `is_log_visible` in the API might be useful.

Alexey-Yakovenko: Ah, then yes. Anyway, this API is not complete yet, and it's to early to start exposing everything. The API may change too. It will become more clear when UI for the logger has more features.

hyperblast: Should I update my pull request with discussed changes or you want to rework logging yourself?

Alexey-Yakovenko: Let's keep it like this for now. I will look into it when I'm back in October.

PR #1647, opened 2016-09-17
Minor tweaks taken from OpenBSD port of deadbeef

zmyrgel:
Here's small tweaks taken from OpenBSD's port of deadbeef.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -46,11 +46,14 @@ deadbeef_SOURCES =\
 	playqueue.c playqueue.h\
 	sort.c sort.h\
 	logger.c logger.h
-	
+
 #	ConvertUTF/ConvertUTF.c ConvertUTF/ConvertUTF.h
 
 sdkdir = $(pkgincludedir)
 sdk_HEADERS = deadbeef.h
+
+# Fix 'undefined symbol' errors when loading plugins
+deadbeef_LDFLAGS = $(LDFLAGS) -Wl,--export-dynamic
 
 deadbeef_LDADD = $(LDADD) $(DEPS_LIBS) $(ICONV_LIB) $(DL_LIBS) -lm -lpthread $(INTL_LIBS) plugins/libparser/libparser.a
 
--- a/conf.c
+++ b/conf.c
@@ -73,6 +73,7 @@ conf_free (void) {
     }
     conf_items = NULL;
     changed = 0;
+    mutex_unlock (mutex);
     mutex_free (mutex);
     mutex = 0;
 }
--- a/plugins.c
+++ b/plugins.c
@@ -1327,7 +1327,7 @@ plug_select_output (void) {
 #else
     char outplugname[100];
 #if defined( HAVE_COCOAUI)
-    conf_get_str ("output_plugin", "CoreAudio", outplugname, sizeof (outplugname));
+    conf_get_str ("output_plugin", "CoreAudio output plugin", outplugname, sizeof (outplugname));
 #elif defined(HAVE_SNDIO)
     conf_get_str ("output_plugin", "sndio output plugin", outplugname, sizeof (outplugname));
 #else
--- a/plugins.c
+++ b/plugins.c
@@ -1326,8 +1326,10 @@ plug_select_output (void) {
     return 0;
 #else
     char outplugname[100];
-#ifdef HAVE_COCOAUI
+#if defined( HAVE_COCOAUI)
     conf_get_str ("output_plugin", "CoreAudio", outplugname, sizeof (outplugname));
+#elif defined(HAVE_SNDIO)
+    conf_get_str ("output_plugin", "sndio output plugin", outplugname, sizeof (outplugname));
 #else
     conf_get_str ("output_plugin", "ALSA output plugin", outplugname, sizeof (outplugname));
 #endif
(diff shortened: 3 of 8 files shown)


plugins.c
```diff
@@ -1327,7 +1327,7 @@ plug_select_output (void) {
 #else
     char outplugname[100];
 #if defined( HAVE_COCOAUI)
-    conf_get_str ("output_plugin", "CoreAudio", outplugname, sizeof (outplugname));
+    conf_get_str ("output_plugin", "CoreAudio output plugin", outplugname, sizeof (outplugname));
```
Alexey-Yakovenko: This is wrong -- the name of the plugin is `CoreAudio`, not `CoreAudio output plugin`

plugins/mp3/mp3.h
```diff
@@ -109,23 +109,16 @@ typedef struct {
     DB_fileinfo_t info;
     // input buffer, for MPEG data
     buffer_t buffer;
-    union {
```
Alexey-Yakovenko: There _is_ the need to use the union -- it saves some memory.

Perhaps the better idea would be to convert to named unions/structs, instead of this change.

Makefile.am
```diff
@@ -46,12 +46,15 @@ deadbeef_SOURCES =\
...
 	sort.c sort.h\
 	logger.c logger.h
-	
+
 #	ConvertUTF/ConvertUTF.c ConvertUTF/ConvertUTF.h
 
 sdkdir = $(pkgincludedir)
 sdk_HEADERS = deadbeef.h
 
+# Fix 'undefined symbol' errors when loading plugins
```
Alexey-Yakovenko: Neither the commit message, nor this comment, explain what this change really does.

Why are you getting the undefined symbol errors without this change in the first place?

plugins.c
```diff
@@ -1372,8 +1374,10 @@ plug_reinit_sound (void) {
 
     if (plug_select_output () < 0) {
         char outplugname[100];
-#ifdef HAVE_COCOAUI
-        conf_get_str ("output_plugin", "core audio output plugin", outplugname, sizeof (outplugname));
+#if defined(HAVE_COCOAUI)
+        conf_get_str ("output_plugin", "CoreAudio output plugin", outplugname, sizeof (outplugname));
```
Alexey-Yakovenko: The wrong coreaudio plugin name again, and the same below.

plugins/aac/aac.c
```diff
@@ -29,7 +29,7 @@
 #include "../../deadbeef.h"
 #include "aac_parser.h"
 
-#include "mp4ff.h"
+#include "../libmp4ff/mp4ff.h"
```
Alexey-Yakovenko: It seems that the same fix is needed in ALAC, converter, artwork and artwork-legacy plugins.

plugins/gtkui/prefwin.c
```diff
@@ -187,7 +187,13 @@ gtkui_run_preferences_dlg (void) {
     // output plugin selection
     combobox = GTK_COMBO_BOX (lookup_widget (w, "pref_output_plugin"));
 
+#if defined(HAVE_COCOAUI)
```
Alexey-Yakovenko: This change is in the GTKUI, this condition is always FALSE -- CocoaUI and CoreAudio are on OSX, we don't use GTK there.

We don't need to have coreaudio here.

plugins/gtkui/prefwin.c
```diff
@@ -452,7 +458,13 @@ on_pref_output_plugin_changed          (GtkComboBox     *combobox,
     DB_output_t *new = NULL;
 
     deadbeef->conf_lock ();
+#if defined(HAVE_COCOAUI)
```
Alexey-Yakovenko: same -- cocoaui and coreaudio don't co-exist with GTKUI.

conf.c
```diff
@@ -73,6 +73,7 @@ conf_free (void) {
     }
     conf_items = NULL;
     changed = 0;
+    mutex_unlock (mutex);
```
Alexey-Yakovenko: This change seems to be wrong. It may cause a race condition, if another thread locks the mutex between mutex_unlock and mutex_free calls. In this case, mutex_free will return `EBUSY` failure, and cause a resource leak.

Why is this change needed?

plugins/aac/aac.c
Alexey-Yakovenko: Cross this. I'm wrong. Other plugins already do the right thing: `-I@top_srcdir@/plugins/libmp4ff` in Makefile.am

AAC plugin should be doing the same.

Alexey-Yakovenko: Err.. but AAC plugin _is_ doing this as well.. Now, I need an explanation why this change is needed on OpenBSD. Everything should work without it.

plugins.c
zmyrgel: Ok, I'll provide more detailed commit messages, need to learn a bit of git-fu first.
I'll sync the CoreAudio naming, seems easy.

plugins/mp3/mp3.h
zmyrgel: Or perhaps this change can stay local in OpenBSD ports, OpenBSD is still using gcc 4.2.1 but recently imported llvm to base so the default compiler at least on some arches might change in the near future.

Makefile.am
zmyrgel: Hmm, I'm not actually sure. Its in the OpenBSD's patches and tried to dig old commit logs and mailing lists on why these were done but couldn't find it. 
My guess is that it has something to do with archs still using GCC3.
But I'll verify it.

plugins/gtkui/prefwin.c
zmyrgel: Ok, good to know.

plugins/aac/aac.c
zmyrgel: Ok, I'll work on this a bit more. These all might be workarounds for older GCC versions ( GCC2 was still used by some arches a while back).

Alexey-Yakovenko:
I took the commits aa86494e1f57339a53748dacbc38791dd25a4b9d and d93fd99b74966c138b4ddf582802fe2b91d6f72c from this PR. The rest is doing more harm than good, so I won't take them without explanation and significant rework.

PR #1655, opened 2016-10-08
add year to column dropdown options

alexcb:
This adds the `year` as an option of the column drop down for when users wants to add additional columns.

When I first started looking into this, I was unsure how I could add the Year to a column as it wasn't under the pre-populated list -- I now know that I could have just entered `%year%`. Nonetheless, maybe some other users have been confused about this -- now one can simply select `year` from the drop down.

Thanks,
Alex


alexcb:
I just realized that I haven't done anything with the translations -- i'll wait until I hear back from someone.

Alexey-Yakovenko:
You don't need to do anything about translation - it should get translated automatically via Transifex.

Issue #1630, opened 2016-08-24
Add an option to duplicate/clone the playlist


jabol240:
Very useful indeed.

Thanks for the great player, btw

PR #1656, opened 2016-10-10
Add ability to copy(duplicate) a playlist

alexcb:
Hello again,

I wanted to have a go at working on https://github.com/Alexey-Yakovenko/deadbeef/issues/1630
I have it working on linux, but can't test it on OSX (and noticed it likely needs to be added to https://github.com/Alexey-Yakovenko/deadbeef/blob/master/plugins/cocoaui/MainWindow.xib#L327 and https://github.com/Alexey-Yakovenko/deadbeef/blob/master/plugins/cocoaui/widgets/DdbTabStrip.m#L681)

### DIFF
--- a/plugins/gtkui/pltmenu.c
+++ b/plugins/gtkui/pltmenu.c
@@ -102,6 +102,19 @@ on_add_new_playlist1_activate          (GtkMenuItem     *menuitem,
 }
 
 static void
+on_copy_playlist1_activate        (GtkMenuItem     *menuitem,
+                                        gpointer         user_data)
+{
+    const ddb_playlist_t *plt = deadbeef->plt_get_for_idx (pltmenu_idx);
+    if (plt) {
+        int playlist = gtkui_copy_playlist (plt);
+        if (playlist != -1) {
+            gtkui_playlist_set_curr (playlist);
+        }
+    }
+}
+
+static void
 on_actionitem_activate (GtkMenuItem     *menuitem,
                            DB_plugin_action_t *action)
 {
@@ -338,6 +351,7 @@ gtkui_create_pltmenu (int plt_idx) {
     GtkWidget *rename_playlist1;
     GtkWidget *remove_playlist1;
     GtkWidget *add_new_playlist1;
+    GtkWidget *copy_playlist1;
     GtkWidget *separator11;
     GtkWidget *cut;
     GtkWidget *cut_image;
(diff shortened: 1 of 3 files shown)


plugins/gtkui/pltmenu.c
```diff
@@ -102,6 +102,19 @@ on_add_new_playlist1_activate          (GtkMenuItem     *menuitem,
...
 
 static void
+on_copy_playlist1_activate        (GtkMenuItem     *menuitem,
+                                        gpointer         user_data)
+{
+    const ddb_playlist_t *plt = deadbeef->plt_get_for_idx (pltmenu_idx);
+    if (plt) {
+		int playlist = gtkui_copy_playlist (plt);
+		if (playlist != -1) {
+    	    gtkui_playlist_set_curr (playlist);
```
Alexey-Yakovenko: bad indentation: mixed tabs and spaces.
deadbeef source code uses 4 spaces for indentation.

plugins/gtkui/pltmenu.c
```diff
@@ -374,6 +388,10 @@ gtkui_create_pltmenu (int plt_idx) {
     gtk_widget_show (add_new_playlist1);
     gtk_container_add (GTK_CONTAINER (plmenu), add_new_playlist1);
 
+    copy_playlist1 = gtk_menu_item_new_with_mnemonic (_("Copy Playlist"));
```
Alexey-Yakovenko: Instead of adding a custom UI menu item, this should be converted into a plugin action, and put into `plugins/hotkeys/actionhandlers.c`, with `DB_ACTION_PLAYLIST` flag, then registered in the `plugins/hotkeys/hotkeys.c` in the linked list of all plugin actions.

This way it will automatically work in all UI plugins, not just GTKUI.

plugins/gtkui/pltmenu.c
alexcb: whoops, easy fix.

plugins/gtkui/pltmenu.c
alexcb: That's some excellent feedback -- I'll have a look into this and rework the code. Thanks.

Alexey-Yakovenko: Another side-effect after converting this into a plugin action, is that you'll be able to bind the action to a hotkey. You should find your new action in the preferences -> hotkeys tree.

Alexey-Yakovenko: mmm.. wait, I just found that DB_ACTION_PLAYLIST is deprecated and ignored. Please wait with any changes for now. It might be that plugin actions don't support operations on the whole playlists. In which case it is correct to do what you did.

Alexey-Yakovenko: I figured, that you could use the DB_ACTION_COMMON, and call the action something like `Duplicate current playlist`. This action should then appear in main menu. I'm not sure though, where does it belong.

Perhaps doing it the way you did is the best way to go for now. So please just fix the code formatting, and I'll take the PR then (after some testing).

alexcb: It looks like I should add it to somewhere like https://github.com/Alexey-Yakovenko/deadbeef/blob/master/plugins/gtkui/gtkui.c#L1620

The code formatting has been fixed.

let me know if you want a hand with anything else here, or if there's other low hanging fruit I can try fixing.

PR #1676, opened 2016-11-27
Fix plugin actions internationalization

cspk:
The `artwork`, `hotkeys` and `lastfm` plugins lacked gettextization. Plus, if a plugin creates submenus (like `hotkeys` does with its "Skip to" actions), their titles aren't gettextized too. This PR fixes it.


Alexey-Yakovenko:
This is not the way it's supposed to work.

1. Localization is performed on [Transifex](https://www.transifex.com/deadbeef-player/deadbeef-player/dashboard/), I could not accept any changes in the PO files through git / pull requests.
2. Adding every string manually to a localization file is cumbersome and unreliable, it wouldn't pick up changes, etc.
3. The way it should work is by loading all plugins, scanning them for actions, and generating POT file.

Some time ago I started making a proof-of-concept utility to do that, but didn't have time to finish it.
You can find the file here to get the idea: https://github.com/Alexey-Yakovenko/deadbeef/tree/master/tools/action2po

Some stuff in this PR could be useful, basically, I think I can accept the commits nr. 2,3,4, but not the 1 and 5.

For now, I have to reject this PR though.

PR #1686, opened 2016-12-27
notify: Add support for non-legacy artwork plugin

vmednis:


Alexey-Yakovenko:
This is not the way it's supposed to work.

Plugins don't need to support both artwork plugins. We need to convert all plugins to use the new one, then we can remove the old one.

However, the new one is *not ready* for the general usage yet, so it doesn't make sense to convert GTKUI, notify & friends to use it *yet*.

vmednis:
Oh, ok.
Misunderstanding on my side.

PR #1704, opened 2017-02-05
Fixed serious compiler warnings

Mailaender:
From https://build.opensuse.org/package/view_file/multimedia:apps/deadbeef/deadbeef-compiler-warnings.patch?expand=1 which apparently wasn't submitted upstream.


Alexey-Yakovenko:
most of the changes don't make any sense, except the strncat bug in sc68 lib, which should be reported upstream.

Alexey-Yakovenko:
However, the bug in sc68 is in the code which is not used in deadbeef, so it's not a problem either.

Alexey-Yakovenko:
I'm sorry, but this pull request doesn't have a single useful change in it. Have to reject.

PR #1705, opened 2017-02-05
Added a translateable AppData file for Linux software galleries

Mailaender:
See https://en.opensuse.org/openSUSE:AppStore for details.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -67,6 +67,9 @@ docs_DATA = README help.txt about.txt translators.txt ChangeLog\
 desktopdir = $(datadir)/applications
 desktop_DATA = deadbeef.desktop
 
-EXTRA_DIST = $(docs_DATA) $(desktop_DATA) $(INTLTOOL_FILES) translation/extra.c sj_to_unicode.h examples/decoder_template.c examples/dsp_template.c yasmwrapper.sh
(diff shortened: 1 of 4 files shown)


Mailaender:
See https://build.opensuse.org/request/show/454765 for an integration test.

Alexey-Yakovenko:
Please answer all of the questions below:

* What is this file?
* What does it fix?
* How can I test that it works?
* Why do you think it's useful to have this file in Deadbeef source code repository?
* Where is it used?

Alexey-Yakovenko:
It looks like this file adds certain metadata for one of the linux distribution package repositories.

It's not something that I consider as a good thing to have in deadbeef source code.

Mailaender:
The idea of the AppData project is that the developer shall be in control. For Fedora this is mandatory. For @openSUSE it just adds a pretty entry with a screenshot to https://software.opensuse.org/appstore and tools which aggregate it.

Alexey-Yakovenko:
In control of what?

Mailaender:
@hughsie had a nice explanation at https://people.freedesktop.org/~hughsient/appdata/ that I used to link, but now it just redirects to the AppStream definition which isn't exactly what this is about.

PR #1710, opened 2017-02-13
fix: notifications

clixt:
- fix plugins -> notify bug:
   There are two displayed notifications (with and without cover) for new albums.
   This is now fixed, if album cover is present, only one notification with cover is shown.
 - plugins -> notify:
   added option to set custom notification timeout

### DIFF
--- a/plugins/notify/notify.c
+++ b/plugins/notify/notify.c
@@ -72,9 +72,7 @@ notify_thread (void *ctx) {
         if (dbus_message_iter_init(reply, &args)) {
             if (DBUS_TYPE_UINT32 == dbus_message_iter_get_arg_type(&args)) {
                 dbus_message_iter_get_basic(&args, &id);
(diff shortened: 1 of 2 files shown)


Alexey-Yakovenko:
Changing unrelated things in 1 commit is bad.
Commit message `Fixes for notify.c and shellexec.c`is not good enough -- it needs to explain what has been fixed.

Alexey-Yakovenko:
Would be nice to reference bug numbers which get fixed.

Alexey-Yakovenko:
This PR needs some serious rework to get merged. The changes in shellexec are not explained, and look wrong. The changes in notify need explanation, restoring the version number to 1.0, explaining and/or removing the customizable timeout, and maybe other things.

The changes that go to shellexec need to be in separate commits, and in separate PR.

Therefore the PR is rejected.

Issue #1575, opened 2016-05-19
Display total time of selected tracks in the statusbar


spvkgn:
It would be useful if Deadbeef could display selected tracks playtime in the statusbar like as displayed playlist total playtime. For instance foobar2k can demonstrate this feature ("show total time of selection" statusbar option).

mallniya:
Maybe same thing should be displayed in "Duration" field, "Properties" tab of "Track properties" menu when more than 1 track is selected.

PR #1720, opened 2017-03-06
gtkui: display total time of selected tracks in the statusbar (Closes: #1575)

thomassross:


Alexey-Yakovenko:
This won't work. It will run full loop over entire playlist multiple time per second, taking too much CPU. Try this with a playlist of 50K+ songs.

deadbeef.h
```diff
@@ -800,6 +800,7 @@ typedef struct {
     int (*plt_remove_item) (ddb_playlist_t *playlist, DB_playItem_t *it);
     int (*plt_getselcount) (ddb_playlist_t *playlist);
     float (*plt_get_totaltime) (ddb_playlist_t *plt);
+    float (*plt_get_seltime) (ddb_playlist_t *plt);
```
Alexey-Yakovenko: It's not allowed to add new API entry points in the middle of pre-existing APIs, as this would break all existing plugins

playlist.c
```diff
@@ -3365,6 +3365,27 @@ pl_get_totaltime (void) {
     return t;
 }
 
+float
+plt_get_seltime (playlist_t *playlist) {
+    float t = 0;
```
Alexey-Yakovenko: LOCK/UNLOCK missing

plugins/gtkui/gtkui.c
```diff
@@ -141,6 +141,24 @@ static float last_songpos = -1;
 static char sbitrate[20] = "";
 static struct timeval last_br_update;
 
+static void
+format_timestr (char *buf, float time) {
```
Alexey-Yakovenko: If this was done through title formatting, another time formatting function would not have been needed.

plugins/gtkui/gtkui.c
```diff
@@ -151,26 +169,22 @@ update_songinfo (gpointer ctx) {
...
+
+    float pl_seltime = deadbeef->pl_get_seltime ();
+    char seltime_str[512] = "";
+    format_timestr(seltime_str, pl_seltime);
 
     DB_playItem_t *track = deadbeef->streamer_get_playing_track ();
 
     if (!output || (output->state () == OUTPUT_STATE_STOPPED || !track)) {
-        snprintf (sbtext_new, sizeof (sbtext_new), _("Stopped | %d tracks | %s total playtime"), deadbeef->pl_getcount (PL_MAIN), totaltime_str);
+        snprintf (sbtext_new,
```
Alexey-Yakovenko: It would make much more sense to convert this code to use title formatting, and make it configurable.

plugins/gtkui/gtkui.c
```diff
@@ -181,7 +195,13 @@ update_songinfo (gpointer ctx) {
 
         char buffer[200];
         deadbeef->tf_eval (&ctx, statusbar_bc, buffer, sizeof (buffer));
-        snprintf (sbtext_new, sizeof (sbtext_new), "%s | %d tracks | %s total playtime", buffer, deadbeef->pl_getcount (PL_MAIN), totaltime_str);
+        snprintf (sbtext_new,
```
Alexey-Yakovenko: Same here. I already started conversion to title formatting, but some of the things are not yet accessible through TF.

Correct way to add this feature would be to add something like %selected_playtime%, and make this script customizable.

deadbeef.h
thomassross: Whoops. Is there a specific order to where the two API functions should go in the `#if (DDB_API_LEVEL >= 10)`/`#endif` block or just at the bottom?

Also, should there actually be both a `plt_get_seltime` and `pl_get_seltime` function or just one? I copied the `plt_get_totaltime`/`pl_get_totaltime` functions to create this one and it seemed like the `pl_get_totaltime` function was just the one that was for plugins to use, with a `LOCK`/`UNLOCK`, but you wrote that I should add `LOCK`/`UNLOCK` to `plt_get_seltime`. Just a little confused about this.

Thanks.

plugins/gtkui/gtkui.c
thomassross: By title formatting do you mean tf_eval?

deadbeef.h
thomassross: Oh. Forget this, I didn't realize you wanted me to implement this through title formatting, not adding another API

Alexey-Yakovenko: > Whoops. Is there a specific order to where the two API functions should go in the #if (DDB_API_LEVEL >= 10)/#endif block or just at the bottom?

Just at the bottom. But if you implement this stuff as part of title formatting - no new apis would be necessary.

> Also, should there actually be both a plt_get_seltime and pl_get_seltime function or just one?

Just one, plt_get_seltime.

> Just a little confused about this.

It seems to be a bug in the plt_get_totaltime, I'll fix it. Thanks for spotting.

plugins/gtkui/gtkui.c
Alexey-Yakovenko: > By title formatting do you mean tf_eval?

Yeah, you'd need to add a new variable handling to tf_eval_int.

deadbeef.h
thomassross: I think that adding the new apis would still be a good idea just in case a plugin wants to use them. What do you think?

playlist.c
```diff
@@ -541,6 +541,14 @@ plt_set_curr (playlist_t *plt) {
     if (plt != playlist) {
         playlist = plt;
         if (!plt_loading) {
+
+            for (playItem_t *it = playlist->head[PL_MAIN]; it; it = it->next[PL_MAIN]) {
```
thomassross: Thinking about replacing this with a `plt_recalculate_seltime` function -- what do you think? Should this also be open to plugins if it were to exist?

deadbeef.h
Alexey-Yakovenko: Should be fine.

playlist.c
Alexey-Yakovenko: I'm generally only exposing stuff "on demand". It doesn't make any sense to expose something if there's no use for it.

thomassross: Okay, cool. Should the function exist though? The exact same code is used for both `plt_set_curr` and `plt_set_curr_idx`.

Alexey-Yakovenko: Yes, this is fine.

thomassross:
I have made the changes requested. I will convert the rest of the statusbar to title formatting in a separate pull request.

deadbeef.h
```diff
@@ -1389,6 +1389,12 @@ typedef struct {
 
     void (*pl_item_set_endsample) (DB_playItem_t *it, int64_t sample);
 
+
+    // get total play time of selected tracks
+    float (*plt_get_seltime) (ddb_playlist_t *plt);
+
+    float (*pl_get_seltime) (void);
```
Alexey-Yakovenko: This function needs to be removed. The functions which operate on current playlist are considered harmful and usually marked as deprecatred.

playlist.c
```diff
@@ -536,11 +536,23 @@ plt_find (const char *name) {
 
 
 void
```
Alexey-Yakovenko: It seems that this is a local utility function, so it must be made `static`.

playlist.c
```diff
@@ -3365,9 +3388,44 @@ pl_get_totaltime (void) {
...
+    UNLOCK;
+    return t;
+}
+
 void
 pl_set_selected (playItem_t *it, int sel) {
     LOCK;
+    float dur = pl_get_item_duration (it);
+    if (dur > 0 && it->in_playlist){
+        for (playItem_t *playlist_it = playlist->head[PL_MAIN];
```
Alexey-Yakovenko: This loop will make selecting and deselecting items very slow, even when selected time functionality is not used.

Have you tried testing your changes with a large playlist of >50K items?

Alexey-Yakovenko: Every time this function is used, it will iterate over entire playlist.
Imagine that you have a playlist of 50K tracks, and you perform search in it.
Every time you type a character, this code would lead to 500M iterations JUST for updating the selected time.

This is unacceptable.

What I suggest doing instead:

Every time a selection changes, set a flag that selected time needs to be recalculated next time it's requested.

When plt_get_seltime is called, check the flag, recalculate as needed, and reset the flag to false.

Most of the time it should be clear which playlist is modified, so the flag should reside in `playlist_t`.

If it's not possible to determine the playlist -- perhaps it would make sense to set the flag for all playlists, or refactor the code to make playlist known.

playlist.c
```diff
@@ -2642,7 +2665,7 @@ void
 plt_select_all (playlist_t *playlist) {
     LOCK;
     for (playItem_t *it = playlist->head[PL_MAIN]; it; it = it->next[PL_MAIN]) {
-        it->selected = 1;
```
Alexey-Yakovenko: In many cases (like this), it was intentionally optimized this way. Your change deoptimizes certain bottlenecks, and makes them very slow.

Every time you make changes like that, keep in mind that there are large playlists.

I usually test with ~50-60K items. Some users reported that they have playlists with >200K tracks.

So, iterating over a few hundred thousands items is slow already. When you call mutex lock/unlock per item -- that's bad. That will make a significant slow down already.

But when you also iterate over entire playlist again for each item -- that's completely unacceptable.

E.g. in this case -- playlist is known, but you call pl_set_selected which will attempt to find which playlist the item belongs to.

How to fix that case:

Create a new utility function for selecting items, which takes playlist pointer, item pointer, and selection state.
Don't lock mutex, Don't iterate over entire playlist.

playlist.c
```diff
@@ -536,11 +536,23 @@ plt_find (const char *name) {
 
 
 void
+plt_recalculate_seltime (playlist_t *plt) {
```
Alexey-Yakovenko: This function doesn't seem to set seltime to 0 before recalculation.

playlist.c
```diff
@@ -536,11 +536,23 @@ plt_find (const char *name) {
...
+    }
+}
+
+void
 plt_set_curr (playlist_t *plt) {
     LOCK;
     if (plt != playlist) {
         playlist = plt;
         if (!plt_loading) {
+            plt_recalculate_seltime(playlist);
```
Alexey-Yakovenko: It doesn't seem useful to call that here. This would cause recalculation of seltime every time a playlist switch occurs.

That would not be a problem if you change the logic to recalculate seltime on demand.

playlist.c
```diff
@@ -560,6 +572,8 @@ plt_set_curr_idx (int plt) {
     if (p != playlist) {
         playlist = p;
         if (!plt_loading) {
+            plt_recalculate_seltime(playlist);
```
Alexey-Yakovenko: Same here.

playlist.c
thomassross: The reason I have this function called in this function and the other function is because when I was switching playlists the seltime would always be set to zero. Adding this fixed that. I'm not sure what you mean about changing the logic to recalculate seltime on demand here -- I get the other cases, but here if we set the flag and then next time it is requested, it would effectively be the same as calculating it here as far as I can tell. Please let me know if I'm wrong.

playlist.c
thomassross: Nice find. Thanks.

deadbeef.h
thomassross: Will do.

playlist.c
thomassross: Good idea. Thanks.

playlist.c
Alexey-Yakovenko: > but here if we set the flag and then next time it is requested, it would effectively be the same as calculating it here as far as I can tell. Please let me know if I'm wrong.

I'm not sure if I understand that. Why do you think it's useful to recalculate seltime on playlist switch? It's not supposed to change when playlists get switched...

playlist.c
thomassross: Okay. I'm a little confused on which option you want me to do. My idea so far is a combination of the two:

 - `pl_set_selected` sets the `recalc_seltime` flag on all playlists and sets the selected flag on the playitem that it is given as normal

 - `pl_select_all` calls the utility function you mentioned earlier: 

```pl_set_select_in_playlist(playlist_t *playlist, playItem_t *it, int sel)```

>Create a new utility function for selecting items, which takes playlist pointer, item pointer, and selection state.
>Don't lock mutex, Don't iterate over entire playlist.

 - `plt_set_curr` & `plt_set_curr_idx` -- I'm not sure about this one. As far as I can tell the seltime needs to be recalculated in this function (or set the `recalc_seltime` flag) but I might be wrong about this.

 - `plt_get_seltime`: If the `recalc_seltime` flag is set, recalculate the seltime, else return `playlist->seltime`

What do you think about this?

Alexey-Yakovenko: @thomassross Sorry, didn't have much time to look into this.

What I want you to do: make a playlist with ~50K tracks or more, and before sending next update -- please profile the modified code, and make sure there's no performance regressions. Check that search box (ctrl+f) is responsive. Check that selecting / unselecting playlist items is fast (especially close to the end of playlist). Check that playlist loading time is as fast at it was.

Now answering more specific questions.

> pl_set_selected sets the recalc_seltime flag on all playlists and sets the selected flag on the playitem that it is given as normal

First thing to try is to set recalc_time only for the current playlist.
I don't think that pl_set_selected is ever called to modify selection state of non-current playlist.

> pl_select_all calls the utility function you mentioned earlier:

Yes. And use the same function internally in playlist.c. It might be a good idea to make this function inline, or a macro. Make sure there's NO mutex locking in it. It should just do `selected=1; recalc_seltime=1`

> As far as I can tell the seltime needs to be recalculated in this function (or set the recalc_seltime flag) but I might be wrong about this.

It needs to be recalculated only in 1 situation: from plt_get_seltime when recalc_seltime==1.
Unless plt_get_seltime is called -- the recalculation should not ever happen.

thomassross: Okay, thanks. The only problem is that I only have about 2000 audio tracks on my computer... Lol

I am currently on vacation but next week when I get back I will look into this.

Alexey-Yakovenko: @thomassross I'm using HVSC for that: http://www.prg.dtu.dk/HVSC/HVSC_66-all-of-them.zip

it's only about 100MB, but has around 60K tracks in it.
To get more -- just select all, and duplicate via drag-n-drop.

thomassross:
I have made the changes requested.

Alexey-Yakovenko:
There are still a few small things that I don't like, but this can be merged anyway, and fixed later.
Thanks for fixing the issues!

PR #1748, opened 2017-04-16
Gtk log viewer

saivert:
Better get this PR started.

Currently missing:
* Sticky scrollbar for the log viewer widget.


plugins/gtkui/gtkui.c
```diff
@@ -1089,6 +1090,99 @@ add_window_init_hook (void (*callback) (void *userdata), void *userdata) {
...
+        wingeom_restore (logwindow, "logwindow", 40, 40, 400, 200, 0);
+    }
+    else {
+        wingeom_save(logwindow, "logwindow");
+    }
+    gtkui_show_log_window_internal(show);
+}
+
+gboolean
+on_gtkui_log_window_delete (GtkWidget *widget, GdkEventAny *event, GtkWidget **pwindow) {
```
Alexey-Yakovenko: ~~I think it's possible to set a checkbox on this window in glade, to not delete on close. Then this function won't be needed.~~

plugins/gtkui/gtkui.c
```diff
@@ -1089,6 +1090,99 @@ add_window_init_hook (void (*callback) (void *userdata), void *userdata) {
...
+    gtkui_show_log_window(FALSE);
+    return TRUE; // don't destroy window
+}
+
+void
+gtkui_create_log_window (GtkWidget **pwindow) {
```
Alexey-Yakovenko: The design of this function doesn't feel natural.. possibly it should just modify the global variable instead, since this window is always single-instance?

plugins/gtkui/actionhandlers.c
```diff
@@ -1140,3 +1140,15 @@ action_playback_loop_cycle_handler(DB_plugin_action_t *act, int ctx) {
...
+action_toggle_logwindow_handler_cb (void *data) {
+    gtkui_toggle_log_window();
+    return FALSE;
+}
+
+int
+action_toggle_logwindow_handler(DB_plugin_action_t *act, int ctx) {
+    g_idle_add (action_toggle_logwindow_handler_cb, NULL);
+    return 0;
+}
```
Alexey-Yakovenko: Missing EOL.

plugins/gtkui/gtkui.c
Alexey-Yakovenko: Hmm.. Now I can see that this function is necessary for saving window geometry. Nevermind the first comment please.

PR #1752, opened 2017-04-23
support SONGWRITER tag in cue sheet (closes #844)

### LINKED ISSUE #844: Composer tag in cue-sheet
Original [issue 942](https://code.google.com/p/ddb/issues/detail?id=942) created by Alexey-Yakovenko on 2013-02-26T22:11:46.000Z:

Could you add support please? In cue-sheets this field is called songwriter.
Or please add support for tag &quot;songwriter&quot;. Thanks.

wdlkmpx:
SONGWRITER – Per-disc or per-track songwriter name for CD-Text data.

http://wiki.hydrogenaud.io/index.php?title=Cue_sheet#Cue_sheet_commands

### DIFF
--- a/playlist.c
+++ b/playlist.c
@@ -997,10 +997,10 @@ plt_process_cue_track (playlist_t *playlist, const char *fname, const int64_t st
         pl_add_meta (it, "artist", albumperformer);
     }
     if (songwriter[0]) {
-        pl_add_meta (it, "composer", songwriter);
+        pl_add_meta (it, "SONGWRITER", songwriter);
     }
     else if (albumsongwriter[0]) {
-        pl_add_meta (it, "composer", albumsongwriter);
+        pl_add_meta (it, "SONGWRITER", albumsongwriter);
     }
     if (albumtitle[0]) {
         pl_add_meta (it, "album", albumtitle);
--- a/playlist.c
+++ b/playlist.c
@@ -920,7 +920,7 @@ pl_cue_parse_time (const char *p) {
 }
 
 static playItem_t *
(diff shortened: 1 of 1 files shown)


wdlkmpx:
hmm i also added support for REM COMMENT "..."
should i add it to this PR?

Alexey-Yakovenko:
@wdlkmpx please add to the same PR

Alexey-Yakovenko:
I disagree that SONGWRITER should go into COMPOSER field.
It should just go into a custom field names "SONGWRITER", which is what foobar2000 does too.
Perhaps instead of handling each of these fields individually, we need to make a generic solution for any unrecognised fields.

playlist.c
```diff
@@ -996,6 +996,12 @@ plt_process_cue_track (playlist_t *playlist, const char *fname, const int64_t st
     else if (albumperformer[0]) {
         pl_add_meta (it, "artist", albumperformer);
     }
+    if (songwriter[0]) {
+        pl_add_meta (it, "composer", songwriter);
```
Alexey-Yakovenko: `composer` needs to be changed to `SONGWRITER`, to keep the same behavior as foobar2000.

wdlkmpx:
Done, but I do expect to see SONGWRITER as the composer in the correspondent field...

I agree a more generic way to add the tags would better, specially for all the REM XXX tags.. maybe a hash table or something? or a small array of strings to store them all then write the fields + values..

I just tested foobar2000 and it refuses to load a cue file if there is an invalid tag (or cue command). ex COMMENT "aaa"... silly. It also doesn't show this stuff i've seen in several cue sheets:

REM DISCNUMBER 1
REM TOTALDISCS 5

So, I guess deadbeef can do better at adding tags from cue files... after all these years i thought foobar was complete, but it's not.

Last year a guy sent a few lossless LP rips (old anime soundtracks), telling me to not share them with anyone. A big track with no cue sheet. I would play the whole record and listen carefully  then i'd create a cue sheet specifying everything including the composer(s).. then split the flac file into tracks using the converter. Then I had to manually edit some fields to add the missing info and sent him the tracks and the cue file..

Alexey-Yakovenko:
> I just tested foobar2000 and it refuses to load a cue file if there is an invalid tag (or cue command). 

This behavior seems fine to me.. When someone creates a broken CUE file, it may appear as correct in the players, but then e.g. would fail when you attempt to burn to CD.

PR #1764, opened 2017-05-06
gtkui: added support for dim and bright text in playlist columns

saivert:


plugins/gtkui/gdkdrawing.c
```diff
@@ -220,6 +220,20 @@ draw_text_custom (drawctx_t *ctx, float x, float y, int width, int align, int ty
 }
 
 void
+draw_text_custom_with_markup (drawctx_t *ctx, float x, float y, int width, int align, int type, int bold, int italic, const char *markup) {
```
saivert: This isn't currently used but might come handy later.

PR #1765, opened 2017-05-07
basic cue sheet syntax checking

wdlkmpx:
Instead of adding like 7 extra if's (CATALOG, CDTEXTFILE, FILE, FLAGS, ISRC, POSTGAP, REM)
i decided to create an array of strings and check each line against that array..

Logs unknown cue commands as errors. Something like this:
Unknown cue sheet item: COMMENT "AAAA"

### DIFF
--- a/playlist.c
+++ b/playlist.c
@@ -1069,6 +1069,12 @@ plt_insert_cue_from_buffer_int (playlist_t *playlist, playItem_t *after, playIte
 
     const char *charset = junk_detect_charset_len (buffer, buffersize);
 
+    char *cue_commands[] = {
+        "CATALOG ", "CDTEXTFILE ", "FILE ", "FLAGS ", "INDEX ", "ISRC ",
+        "PERFORMER ", "POSTGAP ", "PREGAP ", "REM ", "SONGWRITER ",
+        "TITLE ", "TRACK ", NULL,
+    };
+
     LOCK;
     playItem_t *ins = after;
     char albumperformer[256] = "";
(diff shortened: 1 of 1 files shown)


playlist.c
```diff
@@ -1069,6 +1069,12 @@ plt_insert_cue_from_buffer_int (playlist_t *playlist, playItem_t *after, playIte
 
     const char *charset = junk_detect_charset_len (buffer, buffersize);
 
+    char *cue_commands[] = {
```
Alexey-Yakovenko: This needs to be declared as `const char *`.

playlist.c
```diff
@@ -1069,6 +1069,12 @@ plt_insert_cue_from_buffer_int (playlist_t *playlist, playItem_t *after, playIte
...
+        "CATALOG ", "CDTEXTFILE ", "FILE ", "FLAGS ", "INDEX ", "ISRC ",
```
Alexey-Yakovenko: There's no need to save space and combine multiple values on the same line.
Diff tools work better when the values each take a separate line, so please split them up.

playlist.c
```diff
@@ -1119,6 +1125,20 @@ plt_insert_cue_from_buffer_int (playlist_t *playlist, playItem_t *after, playIte
         buffersize -= p-buffer;
         buffer = p;
         p = pl_cue_skipspaces (str);
+
+        // validate cue command
+        int i = 0, valid_cue_command = 0;
```
Alexey-Yakovenko: it doesn't seem like the `i` variable is used anywhere outside of the for loop before, so it should be declared inline: `for (int i = 0; ...`, then this line should look this way: `int valid_cue_command = 0;`

playlist.c
```diff
@@ -1119,6 +1125,20 @@ plt_insert_cue_from_buffer_int (playlist_t *playlist, playItem_t *after, playIte
...
+        for (i = 0; cue_commands[i]; i++) {
+            if (!strncmp (p, cue_commands[i], strlen(cue_commands[i]))) {
+               valid_cue_command = 1;
+               break;
+            }
+        }
+        if (valid_cue_command == 0) {
+            trace_err ("Unknown cue sheet item: %s\n", p);
```
Alexey-Yakovenko: This error message needs to be removed, unless it points to the cuesheet file name, or the track filename for embedded cuesheet.

Alexey-Yakovenko:
A few more notes about this change...

* While I can see how you added checking for valid cuesheet commands, I can't see where or whether you added the actual handling for them. E.g. `CDTEXTFILE` is in the list of valid commands, but there's no code which parses and stores it in RAM.

* Throwing an error when invalid command is encountered is quite pointless, when there's no actual error condition. The invalid/unknown command is skipped, and the parser moves on.

wdlkmpx:
I think I better close this and open a new pull request that deals with some meaningful "extra tags" from cue sheets, a more generic method that includes REM DATE, REM GENRE and REM COMMENT

Alexey-Yakovenko:
Ok, feel free to do that.

PR #1767, opened 2017-05-10
better way to handle "extra tags" from cue sheets (and support more t…

wdlkmpx:
…ags)

Using arrays of strings, very easy to maintain..

Supported tags:

CATALOG (CATALOG)
ISRC (ISRC)
REM DATE (year)
REM GENRE (genre)
REM COMMENT (comment)
REM COMPOSER (composer)
REM DISCNUMBER (disc)
REM TOTALDISCS (numdiscs)
REM DISCID (DISCID)

plt_insert_cue_from_buffer_int() gets the extra tags before a TRACK command is detected
and adds them to the playlist tracks after processing the cue sheet.

This is very handy for those who have their music collection in audiofile+cue form
(properly organized) and want to see the info reflected in the player.
It's also useful to automatically add tags to transcoded files..

### DIFF
--- a/playlist.c
+++ b/playlist.c
@@ -161,7 +161,7 @@ const char *cue_field_map[] = {
     "REM DISCID ", "DISCID",
     NULL, NULL,
 };
-#define MAX_EXTRA_TAGS_FROM_CUE ((sizeof(cue_field_map) / sizeof(cue_field_map[0])) / 2)
+#define MAX_EXTRA_TAGS_FROM_CUE ((sizeof(cue_field_map) / sizeof(cue_field_map[0])))
 
 void
 pl_set_order (int order) {
@@ -1094,8 +1094,9 @@ plt_insert_cue_from_buffer_int (playlist_t *playlist, playItem_t *after, playIte
     const char *dec = pl_find_meta_raw (origin, ":DECODER");
     const char *filetype = pl_find_meta_raw (origin, ":FILETYPE");
 
-    char extra_tags_values[MAX_EXTRA_TAGS_FROM_CUE][255];
-    memset (extra_tags_values, 0, sizeof (extra_tags_values));
+    char extra_tags[MAX_EXTRA_TAGS_FROM_CUE][255];
+    memset (extra_tags, 0, sizeof (extra_tags));
+    int extra_tag_index = 0;
 
     int have_track = 0;
 
@@ -1199,7 +1200,9 @@ plt_insert_cue_from_buffer_int (playlist_t *playlist, playItem_t *after, playIte
             if (!have_track) {
                 for (int m = 0; cue_field_map[m]; m += 2) {
                     if (!strncmp (p, cue_field_map[m], strlen(cue_field_map[m]))) {
-                        pl_get_qvalue_from_cue (p + strlen(cue_field_map[m]), sizeof(extra_tags_values[m/2]), extra_tags_values[m/2], charset);
+                        strcpy(extra_tags[extra_tag_index],cue_field_map[m+1]);
+                        pl_get_qvalue_from_cue (p + strlen(cue_field_map[m]), sizeof(extra_tags[extra_tag_index+1]), extra_tags[extra_tag_index+1], charset);
+                        extra_tag_index += 2;
                         continue;
                     }
                 }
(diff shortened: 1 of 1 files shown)


playlist.c
```diff
@@ -1092,6 +1080,11 @@ plt_insert_cue_from_buffer_int (playlist_t *playlist, playItem_t *after, playIte
     const char *dec = pl_find_meta_raw (origin, ":DECODER");
     const char *filetype = pl_find_meta_raw (origin, ":FILETYPE");
 
+#define MAX_EXTRA_CUE_TAGS 9
```
Alexey-Yakovenko: It would make more sense to move all of this out of the function body (both the define, and the const arrays below).
It's also quite important to put each string in the arrays on separate line.

Regarding extra_tags_values: it would look more clean to just memset the whole array with 0s.

playlist.c
```diff
@@ -1092,6 +1080,11 @@ plt_insert_cue_from_buffer_int (playlist_t *playlist, playItem_t *after, playIte
...
+    const char extra_tags_cue[MAX_EXTRA_CUE_TAGS][30] =     { "CATALOG ", "ISRC ", "REM DATE ", "REM GENRE ", "REM COMMENT ", "REM COMPOSER ", "REM DISCNUMBER ", "REM TOTALDISCS ", "REM DISCID ", };
```
Alexey-Yakovenko: It's not quite correct to call cuesheet fields "tags".
Also it's quite unreliable to keep the 2 arrays in sync.

I suggest doing this in a similar way as many other places do this in deadbeef:
```
const char *field_map[] = {
         "CATALOG ", "CATALOG",
 ...............
         "REM GENRE ", "genre",
........... etc ..........
         NULL, NULL  // <-- end of list
};
```

playlist.c
```diff
@@ -1198,12 +1183,24 @@ plt_insert_cue_from_buffer_int (playlist_t *playlist, playItem_t *after, playIte
             pl_get_value_from_cue (p + 9, sizeof (index01), index01);
         }
         else {
+            // determine and get extra tags
+            int extra_tag = 0;
+            if (!have_track) {
+                for (int i = 0; i < MAX_EXTRA_CUE_TAGS; i++) {
```
Alexey-Yakovenko: This and similar loops will iterate over 30 items every time (MAX_EXTRA_CUE_TAGS), while you there are only ~10 items in the arrays. The suggestion I posted above would help to solve this too.

playlist.c
```diff
@@ -1198,12 +1183,24 @@ plt_insert_cue_from_buffer_int (playlist_t *playlist, playItem_t *after, playIte
...
+                    if (!strncmp (p, extra_tags_cue[i], strlen(extra_tags_cue[i]))) {
+                        pl_get_qvalue_from_cue (p + strlen(extra_tags_cue[i]), sizeof(extra_tags_values[i]), extra_tags_values[i], charset);
+                        extra_tag = 1;
+                    }
+                }
+            }
+            if (extra_tag == 0) {
```
Alexey-Yakovenko: This if block is redundant and needs to be removed, unless it does something useful.

playlist.c
```diff
@@ -1226,7 +1223,15 @@ plt_insert_cue_from_buffer_int (playlist_t *playlist, playItem_t *after, playIte
...
     for (int i = 0; i < ncuetracks; i++) {
+        // add extra tags to tracks
+        for (int y = 0; y < MAX_EXTRA_CUE_TAGS; y++) {
+           if (extra_tags_values[y]) {
+               pl_add_meta (cuetracks[i], extra_tags_fields[y], extra_tags_values[y]);
+           }
+        }
+        // generated "total tracks" field
         pl_add_meta(cuetracks[i], "numtracks", totaltracks);
+        // --
```
Alexey-Yakovenko: This comment line is meaningless.

wdlkmpx:
I might not be up to the task, I'm not familiar with the deadbeef codebase and i'll forever a noob in c programming.. however i pushed 2 commits with 2 different versions of the same fix.

The first thing that came to mind is something like this (but it looked out of place):

	struct cue_field_map {
		const char *cue_field;
		const char *tag_name;
		char tag_value[255];
	} cue_field_map[] = {
		{ "CATALOG "       , "CATALOG",  "", },
		{ "ISRC "          , "ISRC",     "", },
		{ "REM DATE "      , "year",     "", },
		{ "REM GENRE "     , "genre",    "", },
		{ "REM COMMENT "   , "comment",  "", },
		{ "REM COMPOSER "  , "composer", "", },
		{ "REM DISCNUMBER ", "disc",     "", },
		{ "REM TOTALDISCS ", "numdiscs", "", },
		{ "REM DISCID "    , "DISCID",   "", },
		{ NULL             , NULL,       "", },
	};

PR #1776, opened 2017-05-14
tf: $repeat should accept zero repeat count

saivert:


Alexey-Yakovenko:
The issue is valid, but I decided to write the fix slightly different

PR #1790, opened 2017-05-25
add linux task to vscode config

saivert:


.vscode/linux-build
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+make
+[ $? -ne 0 ] && exit
+. ./.install
```
Alexey-Yakovenko: this line is unnecessary -- quickinstall.sh does this on first line.

.vscode/linux-build
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+make
+[ $? -ne 0 ] && exit
```
Alexey-Yakovenko: I believe this should be rewritten as `make || exit 1`

PR #1799, opened 2017-06-01
gtkui: use GtkApplication model

saivert:
This changes the GTK3 implementation to use a application class based on GtkApplication.
First primary benefit of this is to get access to appmenu functionality which is also implemented with this change.
Second is progress dialog now indicates to desktop shell that the app is busy. e.g Gnome-shell will display a spinner next to the application name in the top bar.

This also cleans up the GTK2 parts of the code. Calling gdk_threads_enter/leave is no longer necessary since DeaDBeeF now runs GTK+ in the main thread.

### DIFF
--- a/plugins/gtkui/Makefile.am
+++ b/plugins/gtkui/Makefile.am
@@ -80,7 +80,7 @@ ddb_gui_GTK2_la_CFLAGS = -std=c99 $(GTK2_DEPS_CFLAGS) $(SM_CFLAGS) $(JANSSON_CFL
 endif
 
 if HAVE_GTK3
-GTKUI_SOURCES_GTK3 = $(GTKUI_SOURCES)
+GTKUI_SOURCES_GTK3 = deadbeefapp.c deadbeefapp.h $(GTKUI_SOURCES)
 
 ddb_gui_GTK3_la_LDFLAGS = -module -avoid-version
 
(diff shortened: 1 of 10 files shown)


Alexey-Yakovenko:
As you can see, the build fails on travis-ci.
And the reason for that is because it's using GTK3-3.0.0.
I will need to update GTK3 in static-deps package to fix that.
Last time we talked, we agreed that it should be 3.10, is that still the case?

plugins/gtkui/gtkui.c
```diff
@@ -78,6 +81,10 @@ GtkStatusIcon *trayicon;
 GtkWidget *traymenu;
 GtkWidget *logwindow;
 
+#if GTK_CHECK_VERSION(3,0,0)
```
Alexey-Yakovenko: It seems like this variable should be moved to deadbeefapp.c, then you won't need the #if.

plugins/gtkui/gtkui.c
```diff
@@ -1236,10 +1224,13 @@ gtkui_thread (void *ctx) {
 
     mainwin = create_mainwin ();
 
+#if GTK_CHECK_VERSION(3,0,0)
+    // This must be called before window is shown
+    gtk_application_add_window ( GTK_APPLICATION (gapp), GTK_WINDOW (mainwin));
```
Alexey-Yakovenko: Documentation is not very clear on why this call is needed. That's the only "useful" piece of information there:
> GTK+ will keep the application running as long as it has any windows.

However, I don't get, what would happen if this is not called?

plugins/gtkui/Makefile.am
```diff
@@ -89,5 +89,14 @@ ddb_gui_GTK3_la_LIBADD = $(LDADD) $(GTK3_DEPS_LIBS) $(SM_LIBADD) ../libparser/li
...
 ddb_gui_GTK3_la_OBJCFLAGS = $(GTK3_DEPS_CFLAGS) $(SM_CFLAGS) $(JANSSON_CFLAGS)
 
+nodist_ddb_gui_GTK3_la_SOURCES = gtkui-gresources.c
+
+resource_files = \
+	gtkui.gresources.xml	\
+	gtk/menus.ui
+
+gtkui-gresources.c: $(resource_files)
+	$(AM_V_GEN) $(GLIB_COMPILE_RESOURCES) --target=$@ --sourcedir=$(srcdir) --generate-source $<
```
Alexey-Yakovenko: ~What's the build artifact of this? It will need to be included in the portable build, that's the script that does the file copying: https://github.com/Alexey-Yakovenko/deadbeef/blob/master/scripts/portable_postbuild.sh~

~And I hope that GTK will be able to find it there...~

Alexey-Yakovenko: EDIT: nvm, I figured that this is a build-time thing, and produces gtkui-gresources.c. Please ignore the previous comment.

Alexey-Yakovenko: There's still a question though.. It seems that this fill will get generated in `plugins/gtkui` folder. That's a wrong location for generated files, and we need to look at how to avoid placing it there. But let's look into this after this PR lands.

plugins/gtkui/gtkui.c
saivert: In this case it is needed for the appmenu to work. If call is after gtk_widget_show(mainwin) then it wont work. You can have GtkApplication run without windows by calling g_application_hold() but that is irrelevant here.

saivert:
Gtk+ to 3.10 should be sufficient.

PR #1804, opened 2017-06-04
hotkeys: Add "Add/Remove From Playback Queue" option

### LINKED ISSUE #942: Add to toggle in playback queue hotkey
Original [issue 1040](https://code.google.com/p/ddb/issues/detail?id=1040) created by Alexey-Yakovenko on 2014-01-22T21:55:21.000Z:

In version 0.6 the 'Add to playback queue' and 'Remove from playback queue' actions were added so that hotkeys can be set for them. However, it would be nice if a single hotkey could be used to add or remove a track from the playback queue.

kuba160:
This option adds or removes song(s) from playback queue. If they are already present in queue, they will be removed. This will close #942.

### DIFF
--- a/plugins/hotkeys/actionhandlers.c
+++ b/plugins/hotkeys/actionhandlers.c
@@ -586,6 +586,36 @@ action_clear_playlist_handler (DB_plugin_action_t *act, int ctx) {
 }
 
 int
+action_toggle_playqueue_handler (DB_plugin_action_t *act, int ctx) {
+    ddb_playlist_t *plt = deadbeef->action_get_playlist ();
+
+    DB_playItem_t *it = deadbeef->plt_get_first (plt, PL_MAIN);
+    while (it) {
+        if (ctx == DDB_ACTION_CTX_PLAYLIST || (ctx == DDB_ACTION_CTX_SELECTION && deadbeef->pl_is_selected (it))) {
+            int i,on_queue=0;
+            for(i=0;i<deadbeef->playqueue_get_count();i++){
+                if(deadbeef->playqueue_get_item (i) == it ){
+                    on_queue=1;
+                    break;
+                }
+            }
+            if(on_queue)
+                deadbeef->playqueue_remove (it);
+            else
+                deadbeef->playqueue_push (it);
+
+        }
+        DB_playItem_t *next = deadbeef->pl_get_next (it, PL_MAIN);
+        deadbeef->pl_item_unref (it);
+        it = next;
+    }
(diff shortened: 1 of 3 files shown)


plugins/hotkeys/hotkeys.c
```diff
@@ -1070,12 +1070,20 @@ static DB_plugin_action_t action_add_to_playqueue = {
     .next = &action_remove_from_playqueue
 };
 
+static DB_plugin_action_t action_toggle_playqueue = {
+    .title = "Playback/Add\\/Remove From Playback Queue",
```
Alexey-Yakovenko: There was no opinions about this, so we'll go with what I suggested. Please use this text: `Toggle In Playback Queue`

plugins/hotkeys/hotkeys.c
```diff
@@ -1070,12 +1070,20 @@ static DB_plugin_action_t action_add_to_playqueue = {
...
+    .next = &action_add_to_playqueue
+};
+
 static DB_plugin_action_t action_toggle_mute = {
     .title = "Playback/Toggle Mute",
     .name = "toggle_mute",
     .flags = DB_ACTION_COMMON,
     .callback2 = action_toggle_mute_handler,
-    .next = &action_add_to_playqueue
+    .next = &action_toggle_playqueue
```
Alexey-Yakovenko: Every occurance of `toggle_playqueue` needs to be changed to `toggle_in_playqueue`.

plugins/hotkeys/actionhandlers.c
```diff
@@ -586,6 +586,36 @@ action_clear_playlist_handler (DB_plugin_action_t *act, int ctx) {
 }
 
 int
+action_toggle_playqueue_handler (DB_plugin_action_t *act, int ctx) {
+    ddb_playlist_t *plt = deadbeef->action_get_playlist ();
+
+    DB_playItem_t *it = deadbeef->plt_get_first (plt, PL_MAIN);
+    while (it) {
+        if (ctx == DDB_ACTION_CTX_PLAYLIST || (ctx == DDB_ACTION_CTX_SELECTION && deadbeef->pl_is_selected (it))) {
```
kuba160: So does both "Add to playqueue" and "Remove from playqueue". I think this does not have any good use and should be removed, but I think it would not be easy. And there is an issue with adding same track to playqueue multiple times. It will be added, but after 1st repeat whole queue of this track will be removed. I'm going to add this issue right after.

PR #1809, opened 2017-06-12
add cue playlist plugin (tracks/cdimage+cue support)

wdlkmpx:
I've had this code for a while and thought i should open a pull request before it gets lost..

This adds support for opnening *.cue files directly.

With this plugin cue support can be considered complete (in a basic level, because i remember seeing foobar2000 do advanced stuff even with tracks+cue stuff).

Features:
- tracks+cue  support (tags + replaygain, lacking everything else)
- cdimage+cue support (new api (hacks)* to get playlist.c do the required stuff..)

	plt_set_cue_file(const char *filename)
	plt_unset_cue_file(void)

There's also the ability to save playlists in cue format, but it's disabled.
It can be enabled if the auto totaltracks stuff is removed.. but i'm not sure.

cdimage+cue
------------
If only 1 file is detected then:
- set the internal deadbeef cue_file variable
- insert audio_file to playlist
- expect decoders to call deadbeef->plt_insert_cue_int(...), or plt_process_cue
- unset cue_file

tracks+cue
-----------
- 1 file can only have 1 track. 1 = 1
- to prevent deadbeef from loading external or internal cuesheets associated with the audio files referenced by the current cuesheet being processed by the plugin.. set a special value: "___null"
- and make playlist.c stop processing cue stuff while cue_file = "___null"
- unset cue_file after finishing processing the cuesheet.

### DIFF
--- a/deadbeef.h
+++ b/deadbeef.h
@@ -1306,6 +1306,9 @@ typedef struct {
     // search results, or not
     void (*plt_search_process2) (ddb_playlist_t *plt, const char *text, int select_results);
 
+    void (*plt_set_cue_file) (const char *filename);
+    void (*plt_unset_cue_file) (void);
+
     // try loading external and embedded cuesheet, using the configured order (cue.prefer_embedded, default=0)
     DB_playItem_t * (*plt_process_cue) (ddb_playlist_t *plt, DB_playItem_t *after, DB_playItem_t *it, uint64_t numsamples, int samplerate);
 
--- a/deadbeef.h
+++ b/deadbeef.h
@@ -1306,9 +1306,6 @@ typedef struct {
     // search results, or not
     void (*plt_search_process2) (ddb_playlist_t *plt, const char *text, int select_results);
 
-    void (*plt_set_cue_file) (const char *filename);
-    void (*plt_unset_cue_file) (void);
-
     // try loading external and embedded cuesheet, using the configured order (cue.prefer_embedded, default=0)
     DB_playItem_t * (*plt_process_cue) (ddb_playlist_t *plt, DB_playItem_t *after, DB_playItem_t *it, uint64_t numsamples, int samplerate);
 
@@ -1415,6 +1412,8 @@ typedef struct {
 
     // get total playback time of selected tracks
     float (*plt_get_selection_playback_time) (ddb_playlist_t *plt);
+
+    void (*plt_set_cue_file) (ddb_playlist_t *plt, const char *filename);
 #endif
 } DB_functions_t;
 
--- a/playlist.c
+++ b/playlist.c
@@ -121,6 +121,7 @@ static int no_remove_notify;
 static playlist_t *addfiles_playlist; // current playlist for adding files/folders; set in pl_add_files_begin
 
 static int conf_cue_prefer_embedded = 0;
+static char cue_file[255] = "";
 
 typedef struct ddb_fileadd_listener_s {
     int id;
@@ -1044,6 +1045,11 @@ plt_process_cue_track (playlist_t *playlist, const char *fname, const int64_t st
 
 playItem_t *
 plt_insert_cue_from_buffer_int (playlist_t *playlist, playItem_t *after, playItem_t *origin, const uint8_t *buffer, int buffersize, uint64_t numsamples64, int samplerate) {
+
+    if (cue_file[0] && strcmp(cue_file, "___null") == 0) {
+        return NULL;
+    }
+
     // FIXME: DB_playItem_t only supports 32bit sample count now;
     // the hack should be removed, when the 64 bit sample counts are implemented
     int numsamples = (int)numsamples64;
@@ -1278,31 +1284,41 @@ plt_insert_cue_from_buffer (playlist_t *playlist, playItem_t *after, playItem_t 
 
 playItem_t *
 plt_insert_cue_int (playlist_t *plt, playItem_t *after, playItem_t *origin, uint64_t numsamples, int samplerate) {
-    pl_lock ();
-    const char *fname = pl_find_meta_raw (origin, ":URI");
-    int len = strlen (fname);
-    char cuename[len+5];
-    strcpy (cuename, fname);
-    pl_unlock ();
-    strcpy (cuename+len, ".cue");
-    DB_FILE *fp = vfs_fopen (cuename);
-    if (!fp) {
-        strcpy (cuename+len, ".CUE");
-        fp = vfs_fopen (cuename);
+    if (cue_file[0] && strcmp(cue_file, "___null") == 0) {
+        return NULL;
     }
-    if (!fp) {
-        char *ptr = cuename + len-1;
-        while (ptr >= cuename && *ptr != '.') {
-            ptr--;
-        }
-        if (ptr < cuename) {
-            return NULL;
-        }
-        strcpy (ptr+1, "cue");
+
+    DB_FILE *fp;
+    if (cue_file[0]) {
+        fp = vfs_fopen (cue_file);
+    }
+    else {
+        pl_lock ();
+        const char *fname = pl_find_meta_raw (origin, ":URI");
+        int len = strlen (fname);
+        char cuename[len+5];
+        strcpy (cuename, fname);
+        pl_unlock ();
+        strcpy (cuename+len, ".cue");
         fp = vfs_fopen (cuename);
         if (!fp) {
-            strcpy (ptr+1, "CUE");
+            strcpy (cuename+len, ".CUE");
             fp = vfs_fopen (cuename);
+        }
+        if (!fp) {
+            char *ptr = cuename + len-1;
+            while (ptr >= cuename && *ptr != '.') {
+                ptr--;
+            }
+            if (ptr < cuename) {
+                return NULL;
+            }
+            strcpy (ptr+1, "cue");
+            fp = vfs_fopen (cuename);
+            if (!fp) {
+                strcpy (ptr+1, "CUE");
+                fp = vfs_fopen (cuename);
+            }
         }
     }
     if (!fp) {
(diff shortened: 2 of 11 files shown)


plugins/cue/COPYING
```diff
@@ -0,0 +1,21 @@
+CUE playlist plugin for DeaDBeeF Player
+Copyright (C) 2017 Alexey Yakovenko
```
Alexey-Yakovenko: wrong copyright -- the authorship is not my.

deadbeef.h
```diff
@@ -1306,6 +1306,9 @@ typedef struct {
     // search results, or not
     void (*plt_search_process2) (ddb_playlist_t *plt, const char *text, int select_results);
 
+    void (*plt_set_cue_file) (const char *filename);
```
Alexey-Yakovenko: ~~New APIs must be appended to the end of this structure, in the block with version check.
I.e. after `plt_get_selection_playback_time`.~~

Scratch this, and see another comment.. A new API for cue loading is required to handle this in a thread-safe / reentrant way.

playlist.c
```diff
@@ -4197,6 +4216,16 @@ plt_process_cue (playlist_t *plt, playItem_t *after, playItem_t *it, uint64_t to
 }
 
 void
+plt_set_cue_file(const char *filename) {
+    strncpy(cue_file, filename, sizeof(cue_file));
```
Alexey-Yakovenko: This is not thread safe, and will cause crashes in a number of known situations.
Instead, you need to make a new API function, which takes this new parameter as argument.

plugins/cue/cue.c
```diff
@@ -0,0 +1,616 @@
+/*
+    CUE playlist plugin for DeaDBeeF Player
+    Copyright (C) 2017 Alexey Yakovenko
```
Alexey-Yakovenko: same as above -- copyright is not mine.

Alexey-Yakovenko: However, if you based this on my code, a mention would be appreciated.

plugins/cue/cue.c
```diff
@@ -0,0 +1,616 @@
...
+    "REM COMPOSER ", "composer",
+    "REM DISCNUMBER ", "disc",
+    "REM TOTALDISCS ", "numdiscs",
+    "REM DISCID ", "DISCID",
+    NULL, NULL,
+};
+#define MAX_EXTRA_TAGS_FROM_CUE ((sizeof(cue_field_map) / sizeof(cue_field_map[0])))
+static int extra_tag_index = 0;
+
+static const uint8_t *
```
Alexey-Yakovenko: Looks like there is a lot of code copied from playlist.c
Would it make sense to extract the code into a library instead?

plugins/cue/cue.c
```diff
@@ -0,0 +1,616 @@
...
+        p = buffer;
+        while (p < end) {
+            p = skipspaces (p, end);
+            if (p >= end) {
+                break;
+            }
+            if (!strncasecmp (p, "FILE ", 5)) {
+                DB_playItem_t *it = NULL;
+                pl_get_qvalue_from_cue (p + 5, sizeof (audio_file), audio_file, charset);
+                deadbeef->plt_set_cue_file(fname);
```
Alexey-Yakovenko: This is wrong, because it's using a global variable, so it would break the first time when tracks are processed by multiple threads.

My previous mention about this was wrong (2 times), because I didn't realize how this is used.

This can be fixed pretty easily, by making  the plt_set_cue_file take a playlist object as argument, and using pl_lock/unlock inside, and of course store the cue filename in the playlist object.

This way, there won't be multithreading issues.

deadbeef.h
Alexey-Yakovenko: New APIs must be appended to the end of this structure, in the block with version check.
I.e. after `plt_get_selection_playback_time`.

Also, as mentioned below, plt_set_cue_file should take `ddb_playlist_t *plt` argument, and store the value there.

Also, it's enough to have only 1 function, and call `plt_set_cue_file (plt, NULL)` or `plt_set_cue_file (plt, "")`.

wdlkmpx:
Thanks for the review. Yes there is duplicated code, but the plugin do things a bit differently.

Creating cueutil.c/h does seem the right way unify both methods. I'll look into this.

I'll try to address all the issues. Maybe the copyright string should be "Alexey Yakovenko and other contributors"? most of the code and ideas originate from you..

Alexey-Yakovenko:
> I'll try to address all the issues. Maybe the copyright string should be "Alexey Yakovenko and other contributors"? most of the code and ideas originate from you..

Your code should go under your copyright.
You can also put multiple copyright lines, like here:
https://github.com/Alexey-Yakovenko/deadbeef/blob/master/plugins/artwork/artwork.c

depends what's there.

wdlkmpx:
Pushed new version. I see the macos build failed.

All reported issues have been addressed to some degree. But probably need some guidance to continue..
- thread safety
- cueutil.c/h
- copyright

Couldn't get rid of __null, but now it's __ignore

	if (playlist->cue_file && strcmp(playlist->cue_file, "__ignore") == 0) {
		return NULL;
	}

(tracks+cue hacks) Had to add it to:

- plt_insert_cue_int
to prevent this cue file from loading:
01. track name.flac
01. track name.cue
(any invalid file does the trick, but __ignore ensures that
vfs_open is not triggered anywhere in the function...)

- plt_insert_cue_from_buffer_int
i placed a file that contained an internal cuesheet
and the cuesheet got loaded, making cue playlist look wrong..
this fixed it...

I don't know know to fix these 2 issues without the "hacks". they're certainly rare and only people who do quirky things will hit those bugs (if the hacks are removed)..

Alexey-Yakovenko:
> Pushed new version. I see the macos build failed.

don't worry about macos for now. it's easy to fix.

playlist.h
```diff
@@ -79,6 +79,7 @@ typedef struct playlist_s {
     unsigned fast_mode : 1;
     unsigned files_adding : 1;
     unsigned recalc_seltime : 1;
+    char *cue_file;
```
Alexey-Yakovenko: this variable should go before the bit-fields, to avoid excessive padding.

Alexey-Yakovenko: (it's a private field, so it doesn't affect backwards compat)

plugins/cue/COPYING
```diff
@@ -0,0 +1,22 @@
+CUE playlist plugin for DeaDBeeF Player
+Copyright (C) 2017 wdlkmpx
```
Alexey-Yakovenko: I would very much prefer if a real name is used here, otherwise please state that this is a github account, like this: `github:wdlkmpx`

plugins/cue/cue.c
```diff
@@ -0,0 +1,540 @@
...
+#include "../../shared/cueutil.h"
+
+//#define trace(...) { fprintf(stderr, __VA_ARGS__); }
+#define trace(fmt,...)
+
+DB_functions_t *deadbeef;
+
+//===================================================================================
+
+const char *cue_field_map[] = {
```
Alexey-Yakovenko: That's still a duplicate with `playlist.c`, and can be moved to cueutil.c

plugins/cue/cue.c
```diff
@@ -0,0 +1,540 @@
...
+    "REM DATE ", "year",
+    "REM GENRE ", "genre",
+    "REM COMMENT ", "comment",
+    "REM COMPOSER ", "composer",
+    "REM DISCNUMBER ", "disc",
+    "REM TOTALDISCS ", "numdiscs",
+    "REM DISCID ", "DISCID",
+    NULL, NULL,
+};
+#define MAX_EXTRA_TAGS_FROM_CUE ((sizeof(cue_field_map) / sizeof(cue_field_map[0])))
```
Alexey-Yakovenko: same, should go to cueutil.h

deadbeef.h
```diff
@@ -1412,6 +1412,8 @@ typedef struct {
 
     // get total playback time of selected tracks
     float (*plt_get_selection_playback_time) (ddb_playlist_t *plt);
+
+    void (*plt_set_cue_file) (ddb_playlist_t *plt, const char *filename);
```
wdlkmpx: plt_set_cue_file is basically a way for the cue plugin to "communicate" with the internal cue sheet stuff. it can also be used by other (playlist) plugins to make deadbeef "__ignore" cuesheets until they finish processing stuff.

accepted values for filename:
- full path to cuesheet file: force cue_sheet overriding "internal decisions"
- __ignore: make deadbeef ignore cuesheets
- NULL: revert the above

wdlkmpx:
With V3 i think it's almost ready.

Alexey-Yakovenko:
I'm going to try it out before merging just to be sure it doesn't break anything in a big way.

Alexey-Yakovenko:
Good job! I merged what we have, and added mac build support. There will be some small followup fixes later.

wdlkmpx:
Cool but the cue plugin is missing in the travis-ci linux build, i always prefer to download the static binaries..

Alexey-Yakovenko:
good point, will fix that.

Alexey-Yakovenko:
@wdlkmpx So there are couple of things that would be very useful to have:

1. support loading CUE files directly, via open file(s) / add file(s) / add folder(s) interface.
2. being able to load files without cuesheet splitting, even if external CUE is present

Currently, cue files are treated either as playlists (loadable via `Open playlist`), or external file metadata (picked up automatically).

This is what I suggest.

1. When adding whole folders: find `*.cue` files, and try loading them first (priority over other normal music files). Add every successfully loaded-via-cue music file to a list. Then do the 2nd (normal) pass, and add the rest of the files which are not in the list loaded by cue, including the automatic CUE search (the way it works now).

2. When adding individual files or file list: do the same as 1, but only include those CUE files in the automatic search, that are specified in the file list.

Example: have the file.mp3 and file.cue in the folder

Use case 1:
* load list: file.mp3 + file.cue
* the cue is loaded first
* if cue references file.mp3 -- then this is registered and the file.mp3 is loaded with cue splitting, and removed from the input list (job done)
* it cue doesn't reference the file.mp3 -- file.cue is ignored
* next, file.mp3 is loaded, and file.cue is automatically found in the list
* automatic cue handling kicks in

Use case 2
* load just file.cue
* will load and split file.mp3 if it's correctly referenced
* nothing will be loaded if file.cue doesn't reference any existing file

Use case 3:
* load just file.mp3
* no cue splitting is performed!

What do you think? any opinions?
I think that would be close or the same as what foobar2000 does, it's more flexible, and would not require CUE playlist plugin.

wdlkmpx:
I agree with your reasoning and use case 1 is more advanced than what foobar does.

I just tested adding a folder to a foobar playlist, it added everything including flac "cdimages" which should be ignored as cue splitting already happened. I think it's better to first replicate this behavior, as it's the easy then do it better.

Adding files:
foobar opens cdimage.flac and cdimage.cue and cue splitting happens, so i have cdimage.flac + plus tracks... this means foobar doesn't check anything.

I think it's ok to get rid of the cue plugin and make all the cue stuff work internally.. that's the right way

PR #1816, opened 2017-06-20
main: call pl_init() after conf_init()

wdlkmpx:
The best way to initialize a (playlist.c) variable read from the config file is in pl_init(), but deadbeef segfaults because conf_get_* tries to lock a mutex that is created in conf_init().

This fixes that issue..

### DIFF
--- a/main.c
+++ b/main.c
@@ -1120,9 +1120,9 @@ main (int argc, char *argv[]) {
         return 0;
     }
 
-    pl_init ();
     conf_init ();
     conf_load (); // required by some plugins at startup
+    pl_init ();
 
     if (use_gui_plugin[0]) {
         conf_set_str ("gui_plugin", use_gui_plugin);


Alexey-Yakovenko:
`pl_init` doesn't access config variables (yet), so I suppose you need this for some new code.
Can you explain the use case?

wdlkmpx:
to save a few cpu cycles i was thinking about updating 'conf_cue_subindexes_as_tracks' only ```pl_init``` and ```pl_configchanged``` (triggered by the gtkui). but i'm not sure..

https://github.com/wdlkmpx/deadbeef/commit/ca9bf9efba8273165f1a90377e362b0115c0e5e2

Alexey-Yakovenko:
pl_configchanged is the right location for that. There's no need to perform this operation from pl_init.

wdlkmpx:
i might be wrong but pl_configchanged is not triggered at startup

when i was having crashes and manually edited
cue.subindexes_as_tracks... 'conf_cue_subindexes_as_tracks' was always
0

that's why i placed  conf_cue_subindexes_as_tracks = conf_get_int
("cue.subindexes_as_tracks", 0);
directly in plt_insert_cue_from_buffer_int()

Alexey-Yakovenko:
@wdlkmpx to clarify to the issue:

Playlist code, including cuesheet loading, is expected to work even if configuration system is not initialized at all.

That's important for unit tests. They expect specific behavior, which should not ever change from external factors.

wdlkmpx:
i tested that change

"cue.subindexes_as_tracks 1" in config file

i start deadbeef opening a flac file with internal cuesheet with many
subindexes and the subindexes are ignored. i open the preferences
dialog and see the CUE: subindexes as tracks checkbox is already
ticked.. i don't touch anything and close the preferences dialog.

i open the flac file again, and the subindexes now appear as tracks..

oleksiy-unity3d:
Do you open the file from command line, or from UI?
I think that maybe the problems happens when you open from command line, because the configchanged event is handled later than that.
If you open from UI -- it should not happen.

Alexey-Yakovenko:
(sorry, posted from a wrong gh account, reposting)
Do you open the file from command line, or from UI?
I think that maybe the problems happens when you open from command line, because (possibly) the configchanged event is handled later than that.
If you open from UI -- it should not happen.

wdlkmpx:
I think i opened the file from the command line. But now in virtual box i compiled deadbeef (version https://github.com/Alexey-Yakovenko/deadbeef/commit/4a8d76c859ad336019aa450f2f790f2186920f12)

Exactly the same behavior

by the way is there any plans to support an official/unofficial deadbeef build for windows?

Alexey-Yakovenko:
> I think i opened the file from the command line.

Then this is most likely the reason. I'll need to look into it.

> by the way is there any plans to support an official/unofficial deadbeef build for windows?

There was a user that contributed some significant amount of time and effort, and ported deadbeef to windows.

Obviously that was not a native UI port, but a Linux UI (GTK) compiled for windows, and I think he made a new output plugin using portaudio or SDL (not sure).

I was encouraging him to post his fork on github, but I'm not sure if he ever did, and I could not find it.

I think he didn't even fork the repo, but he basically took a snapshot and worked "offline".

Other than that, you should not expect windows port from me.

There's a way better music player on windows, which was and still is the main inspiration for deadbeef, called Foobar2000 -- if I was using windows, I'd not hesitate, and just use it.

Alexey-Yakovenko:
@wdlkmpx p.s. it's way better to chat on Slack. Come join us, there are more devs.

PR #1818, opened 2017-06-21
cue: ISRC only applies to the current track

wdlkmpx:
http://pastebin.com/raw/FrnVan7D

https://www.gnu.org/software/ccd2cue/manual/html_node/ISRC-_0028CUE-Command_0029.html#ISRC-_0028CUE-Command_0029

### DIFF
--- a/shared/cueutil.h
+++ b/shared/cueutil.h
@@ -36,7 +36,7 @@
 #define MAX_CUE_TRACKS 99
 
 extern const char *cue_field_map[];
-#define MAX_EXTRA_TAGS_FROM_CUE 18 //((sizeof(cue_field_map) / sizeof(cue_field_map[0])))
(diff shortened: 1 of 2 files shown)


shared/cueutil.h
```diff
@@ -36,7 +36,7 @@
 #define MAX_CUE_TRACKS 99
 
 extern const char *cue_field_map[];
-#define MAX_EXTRA_TAGS_FROM_CUE 18 //((sizeof(cue_field_map) / sizeof(cue_field_map[0])))
+#define MAX_EXTRA_TAGS_FROM_CUE 16 //((sizeof(cue_field_map) / sizeof(cue_field_map[0])))
```
Alexey-Yakovenko: I'm starting to dislike this even more :)

Issue #1533, opened 2016-04-04
Change design of the seekbar


TotalCaesar659:
The playback scale looks really great in STOP state...
![7](https://cloud.githubusercontent.com/assets/14265316/14235891/9d93e436-fa13-11e5-9fbe-9f0c1e3405e8.png)
...but it does worse in PLAY state.
![8](https://cloud.githubusercontent.com/assets/14265316/14235941/d3b99032-fa14-11e5-95bf-37cfc6f53baf.png)
As you see, the black colour doesn't look so good, it's uncomfortable for seeking and unesthetic.
It would be better to correct PLAY state for the scale to something like this:
![9](https://cloud.githubusercontent.com/assets/14265316/14236021/b0be5c64-fa16-11e5-9f25-fa76c034b61e.png)

Alexey-Yakovenko:
Changed the title to use preferred terminology

PR #1819, opened 2017-06-22
Changing design of the seekbar

TotalCaesar659:
Fixes https://github.com/Alexey-Yakovenko/deadbeef/issues/1533

### DIFF
--- a/plugins/gtkui/ddbseekbar.c
+++ b/plugins/gtkui/ddbseekbar.c
@@ -319,18 +319,20 @@ seekbar_draw (GtkWidget *widget, cairo_t *cr) {
     // left
     if (pos > 0) {
         cairo_set_source_rgb (cr, clr_selection.red/65535.f, clr_selection.green/65535.f, clr_selection.blue/65535.f );
-        cairo_rectangle (cr, ax, ah/2-4+ay, pos, 8);
+        cairo_rectangle (cr, ax, a.height/2-5+ay, pos, 10);
         cairo_clip (cr);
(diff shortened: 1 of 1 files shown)


Alexey-Yakovenko:
Screenshot?

kuba160:
I made a quick comparsion on my system (new vs. old)
deadbeef-seekbar
I'm using KDE with Breeze Dark theme.

Alexey-Yakovenko:
I will need to build & run this, before making any decision about this change.

Alexey-Yakovenko:
OK. I like the change to the thicker seekbar when active, but I don't like that it doesn't have a filler anymore. It looks incomplete and bland.

We can go with single-color scheme, but still need to come up with SOME filler, otherwise seekbar looks exactly the same when playing and not playing.

My suggestion for this:

draw blue background when active, with something like 30-50% alpha.

Another problem is that the blue filler overlaps with the border, and causes drawing artifacts because of antialiasing. They need to be drawn in reverse order, i.e. filler first, border second.

TotalCaesar659:
Well, the old seekbar is very bad by design, it looks heterogenous. So I've tried to make it by myself. But I don't have any programming skills, so you can change it as you want and merge then.
Of course, my version looks good with light themes, but not with dark ones.

Alexey-Yakovenko:
OK, I can finish what you started, but please give me some time -- I'm pretty busy atm.

TotalCaesar659:
Yeah, you don't need to hurry up.

Alexey-Yakovenko:
Please see the latest build on master, it should have what you wanted.

TotalCaesar659:
Thank you very much, the seekbar and volumebar look awesome!

PR #1843, opened 2017-08-08
Check for zlib when gme plugin is selected

kuba160:
as gme uses zlib

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -715,7 +715,9 @@ AS_IF([test "${enable_vtx}" != "no"], [
 ])
 
 AS_IF([test "${enable_gme}" != "no"], [
-    HAVE_GME=yes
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -715,7 +715,9 @@ AS_IF([test "${enable_vtx}" != "no"], [
 ])
 
 AS_IF([test "${enable_gme}" != "no"], [
-    HAVE_GME=yes
+    AS_IF([test "${HAVE_ZLIB}" = "yes"], [
```
Alexey-Yakovenko: It's better to write it like this:

`AS_IF([test "${enable_gme}" != "no" -a "${HAVE_ZLIB}" = "yes"], [`

only one if, instead of nested

Alexey-Yakovenko:
I applied the fix manually

Issue #1682, opened 2016-12-12
Indicate playback status in the tray icon


ypoluektovich:
It would be convenient if the tray icon changed according to playback status: it could show a right-facing triangle when playing, two vertical bars when paused, an opaque square (or the default icon?) when stopped. The icons can be made to follow the style of the current icon (green shapes on black rhombus-ish background).

This feature would allow a user to see the playback status at a glance, without the need to have the window opened, or to move the mouse to the icon (and then clicking or waiting). Sometimes it's useful to know whether the silence you hear is because the playback is stopped, or because there's a strip of silence in the audio track. With this information a glance away, the user will be able to fix the potential problem with just a single hotkey press.

PR #1847, opened 2017-08-11
Added option for showing playing status on tray icon (initial sketches, need help)

TotalCaesar659:
When I saw https://github.com/Alexey-Yakovenko/deadbeef/issues/1682 I've got an idea to create icons for playing, pause and stop for tray.
Examples:

This PR is initial, tray icons change only after starting DeadBeeF according to saved status. I don't know how to change them when playing a track, so it needs help.
And I've added only 32x32 icons for checking will they suite users and getting comments. If they do, I'm ready to create and add another sizes.


plugins/gtkui/gtkui.c
```diff
@@ -489,24 +489,47 @@ gtkui_update_status_icon (gpointer unused) {
...
-        icon_name = icon_is_builtin ? "deadbeef" : icon_name;
-    }
-
-    if (!gtk_icon_theme_has_icon(theme, icon_name)) {
-        char iconpath[1024];
-        snprintf (iconpath, sizeof (iconpath), "%s/deadbeef.png", deadbeef->get_system_dir(DDB_SYS_DIR_PREFIX));
-        trayicon = gtk_status_icon_new_from_file(iconpath);
+    int show_playing_status_on_tray_icon = deadbeef->conf_get_int ("gtkui.show_playing_status_on_tray_icon", 0);
+    
+    if (deadbeef->conf_get_int ("gtkui.show_playing_status_on_tray_icon", 1)) {
```
Alexey-Yakovenko: There are a few problems with this code. This function is called when configuration changes, but you need to change the icon when playback state changes.

Another problem is that previous icon object is never freed, so every time this code is called -- there would be a memory leak. However, this problem is present in the original code as well.

And last but not least, you don't want to read the files every time when playback state changes. Instead, you need to preload all icons at initialization time.

Alexey-Yakovenko:
The problem with the icons:
1. They don't look visually appealing enough to get accepted.
2. You provided only 32x32 icons, but all resolutions are necessary.
3. You need to add the new icons to icons/Makefile.am, so that they get installed.

TotalCaesar659:
>They don't look visually appealing enough to get accepted.

I've created the icons based on original DeaDBeeF logo. In fact, I've re-drawn original SVG file and converted it to PNG.

So they can't be less "visually appealing" than the original logo.

>You provided only 32x32 icons, but all resolutions are necessary.
>You need to add the new icons to icons/Makefile.am, so that they get installed.

I thought about it, but as I said, "I've added only 32x32 icons for checking will they suite users and getting comments. If they do, I'm ready to create and add another sizes."
So tell me, why they don't suit you. It would be good to get more concrete details (color, shadows etc.), not just "they don't look visually appealing enough". Many men, many minds.
Again, I already said about the original logo. Maybe, it should be re-drawn too with them.

plugins/gtkui/gtkui.c
TotalCaesar659: Well, I'm not software engineer, so I did it for myself first. So, if somebody checks it and finishes my job - it would be great.

Alexey-Yakovenko:
> Well, I'm not software engineer, so I did it for myself first. So, if somebody checks it and finishes my job - it would be great.

ok.. sorry, but there's nobody to finish your job.

Alexey-Yakovenko:
It's nothing personal, I just think that the icons need some extra work to make them look better. Right now the green cores look too blurry and oversized. Try making them more sharp and slightly smaller. I don't know how much it would help though -- not a designer myself.

Alexey-Yakovenko:
The reason why they're blurry is because of the glow that you carried from the original app icon.
You should try disabling this glow -- it doesn't work well with the kind of icons you're making, especially at this pixel size.

PR #1852, opened 2017-08-14
Add M_PI definition to files, where it can be missing

kuba160:
In these files `M_PI` is being used, but we cannot be sure that it has been defined. Deadbeef uses C99 standard by default (`-std=c99`) which does not include `M_PI` definition. Code compiles fine on most systems, because `M_PI` is included with GNU extensions.

To maintain compatibility with C99 check if `M_PI` is already defined, and if not, define it manually. This approach is already being used by some plugins in deadbeef.

### DIFF
--- a/fft.c
+++ b/fft.c
@@ -25,6 +25,9 @@
 #endif
 #include "deadbeef.h"
 #include <math.h>
+#ifndef M_PI
+#define M_PI 3.1415926535897932384626433832795029
+#endif
 #include <complex.h>
 
 #define N (DDB_FREQ_BANDS * 2)
--- a/plugins/rg_scanner/ebur128/ebur128.c
+++ b/plugins/rg_scanner/ebur128/ebur128.c
@@ -5,6 +5,9 @@
 #include <float.h>
 #include <limits.h>
 #include <math.h> /* You may have to define _USE_MATH_DEFINES if you use MSVC */
+#ifndef M_PI
+#define M_PI 3.1415926535897932384626433832795029
+#endif
 #include <stdio.h>
 #include <stdlib.h>
 


Alexey-Yakovenko:
It looks more correct to add `#define _USE_MATH_DEFINES` globally, no?

kuba160:
well, I'm gonna check if it works on compiler I used, and change it here if it works

kuba160:
oh, you mean globally? I don't know actually.

Alexey-Yakovenko:
> This approach is already being used by some plugins in deadbeef.

I can't find any existing deadbeef code which does that.
Some external libraries do that though. This is the list:

```
./osx/deps/faad2-2.7/libfaad/common.h:#define M_PI 3.14159265358979323846
./osx/deps/ffmpeg/include/libavutil/mathematics.h:#define M_PI           3.14159265358979323846  /* pi */
./plugins/gme/game-music-emu-0.6pre/vgmplay/VGMPlay/chips/mamedef.h:#define M_PI 3.14159265358979323846
./plugins/shn/shorten.h:#define M_PI	3.14159265358979323846
./plugins/wma/libwma/wmadec.h:#define M_PI    3.14159265358979323846
```

Alexey-Yakovenko:
> oh, you mean globally? I don't know actually.

not like system-global, add somewhere to configure.ac..

there's this line
```
CFLAGS="$CFLAGS $INSANE_CFLAGS -D_GNU_SOURCE $PREFIXFLAGS -DDDB_WARN_DEPRECATED=1"
```

you can add some mingw-specific flags there.

kuba160:
well,  `M_PI` still does not gets defined, even with `_USE_MATH_DEFINES`

Alexey-Yakovenko:
Ok.. then there should be another define. but defining M_PI in every file is not the right solution. I'll try to find the right way.

kuba160:
I have looked fft.c file from audacious. The newest version does not use M_PI anymore, it uses self-defined `TWO_PI` instead.

Alexey-Yakovenko:
Including math.h for M_PI was a norm since the 90s. All libc versions I ever encountered do have it, one way or another. If you can fix something by adding -std=gnu99 without code changes -- go for it. Adding #define M_PI to every file which needs it is not acceptable. If you don't like getting it from math.h, we can talk about adding something like pi.h, which can be used for everything. But I don't like this option. This PR is not going to be merged, closing.

PR #1865, opened 2017-08-26
fix invalid desktop file

xi:
In Debian 9 with LXDE, deadbeef would not start and show a warning about the invalid desktop instead.

"Shortcut Group" is a incompatible extension used by the unity desktop. The Desktop switched to the standard "Actions" syntax in Ubuntu 12.04. Source: https://bugs.launchpad.net/unity/+bug/789867

### DIFF
--- a/deadbeef.desktop.in
+++ b/deadbeef.desktop.in
@@ -20,43 +20,8 @@ Categories=Audio;AudioVideo;Player;GTK;
 Keywords=Sound;Music;Audio;Player;Musicplayer;MP3
 Keywords[zh_TW]=Sound;Music;Audio;Player;Musicplayer;MP3;音樂;音樂播放器;播放器;音訊
 
-X-Ayatana-Desktop-Shortcuts=Play;Pause;Stop;Next;Prev
+Actions=Play;Pause;Stop;Next;Prev
 X-PulseAudio-Properties=media.role=music
-
-[Play Shortcut Group]
-Name=Play
-Name[zh_CN]=播放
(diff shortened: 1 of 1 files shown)


Alexey-Yakovenko:
Is there a way to fix the problem without removing the Unity dock menu?

xi:
I do not have ubuntu, but according to the link I posted unity switched to the "Actions" syntax in Ubuntu 12.04 which has reached end of life in 2014.

Alexey-Yakovenko:
Yeah, but your change removes all of the actions. If you don't remove them - you still have problems? Perhaps you need to file a bug to Lxde, as it seems it's at fault of not being able to ignore the fields uknown to it?

Alexey-Yakovenko:
@xi 
> In Debian 9 with LXDE, deadbeef would not start and show a warning about the invalid desktop instead.

Can you link a screenshot of the warning, and possibly name the tool which produces it, so that I can learn more about what's going on? I also need to know when the problem is happening -- at install time, or when you're trying to run the app.

xi:
Somehow I can no longer reproduce the problem. Maybe it was actually unrelated to the invalid desktop file.

I still don't understand why you want to keep long deprecated and invalid "Shortcut" entries. But that is realy up to you.

Alexey-Yakovenko:
> I still don't understand why you want to keep long deprecated and invalid "Shortcut" entries. But that is realy up to you.

they are not invalid -- they're actually used, and removing them will break Unity dock icon menu.

PR #1888, opened 2017-09-24
cocoaui: dnd visual line indicator

carlosanunes:
Added visual line indicator for dnd drops.

A bit buggy when the list is grouped. Example: line is not viewed when group only has one element. line is not shown in first element of grouped list.

### DIFF
--- a/plugins/cocoaui/DdbListview.m
+++ b/plugins/cocoaui/DdbListview.m
@@ -422,6 +422,8 @@ int grouptitleheight = 22;
 
 @interface DdbListContentView : NSView {
     DdbListview *listview;
+    NSPoint _lastDragLocation;
+    BOOL    _dragging;
 }
 - (void)setListView:(DdbListview *)lv;
 @end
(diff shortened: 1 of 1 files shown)


plugins/cocoaui/DdbListview.m
```diff
@@ -523,6 +542,20 @@ - (void)renderAlbumArtForGroup:(DdbListviewGroup_t *)grp
...
 
 
+- (void) drawLineIndicator:(NSRect)dirtyRect yy:(int)yy  {
+
+    int yyline = yy;
+    float indicatorLineWith = 1.f;
+    if ( yyline > 0 ) {
+        yyline -= (indicatorLineWith / 2.f );
+    }
+    [[NSGraphicsContext currentContext] saveGraphicsState];
```
Alexey-Yakovenko: It's worth adding a check whether the line is inside of the dirty rect.

carlosanunes: Will add it when I fix the issues with groups.

PR #1901, opened 2017-10-01
fix #1900

### LINKED ISSUE #1900: DeaDBeeF stops music instead of switching to previous track when you're on another playlist
When DeaDBeeF plays music in playlist 1, and you switch to playlist 2 and press "Previous track" button, the player doesn't switch to previous track in playlist 1, but stops music at all. 
"Next track" button works normally.
Deadbeef version: devel
OS: Lubuntu 16.04.3

dafanasiev:
Fixes #1900, please review.


Alexey-Yakovenko:
merged

PR #1917, opened 2017-11-02
Fix signed char errors on ARM/PPC, fix GCC 7 warnings

StefanBruens:


Alexey-Yakovenko:
I'm not taking PRs with patches to 3rd party deps unless they fix actual problems.
GCC warnings are not.

Alexey-Yakovenko:
This PR may be accepted if the changes get accepted upstream, and you post commit links.

StefanBruens:
You obviously are unable to understand architectural differences, and do not understand what "undefined behaviour" in the C standard means. Irresponsible persons like you should not claim maintainership of any software.

Alexey-Yakovenko:
@StefanBruens 

I think you're misunderstanding something. The libraries which you patched are 3rd party -- I'm not maintaining them. You need to post the patches upstream first.

I'm not accepting such patches mainly because they make it hard to merge updates from upstream.

I did not write the code you changed, and there's even no guarantee it's used in this project.

Alexey-Yakovenko:
And btw, I can assure you that adplug and libsidplay2 both work fine when built for ARM on android. And they worked fine on PPC too when I last checked, but that was many years ago.

PR #1929, opened 2017-11-28
junklib: add more id3 field mappings

wdlkmpx:
data gathered from http://help.mp3tag.de/main_tags.html

also add a few comments and fix a couple entries..


Alexey-Yakovenko:
Most of this change is wrong, non-text id3 frames can't be mapped without conversion code for each individual case. Rejected.

PR #1939, opened 2017-12-10
Chinese Po File Translation completion

oupula:
RT


Alexey-Yakovenko:
I will merge this, because the number of changes is very small, but please use transifex next time.
https://www.transifex.com/projects/p/deadbeef-player/

oupula:
@Alexey-Yakovenko ok,thanks

PR #1946, opened 2017-12-23
Update zh_TW.po

JulianLai:


Alexey-Yakovenko:
Please update the translations via Transifex, as pointed out in the FAQ.

Issue #1927, opened 2017-11-22
100% CPU load by deadbeef-alsa since recent update


cbirchinger:
A very recent update on the 8. Nov. 2017 causes the ALSA plugin to use 100% CPU load
on a single thread (deadbeef-alsa). It does not matter what kind of device is being used.
I've tried onboard intel-hda, usb audio 2.0, xonar pci-e card and AMD HDMI, so i doubt
it's hardware dependent.

So started going back to older versions of "alsa.c". Everything else remains the latest
git release.

First version of alsa.c with problems:
Nov 8, 2017  "alsa: fixed error handling / locking" (8b8d5cd218aeec4b322b04111e851883c49ea670)

Last working version:
Nov 7 2017 "alsa: refactored code with better error handling and attempts to fix" (91124bf7e33816652793d43e4756b8add1c839f8)

Deadbeef version: latest git (22. Nov. 2017)
OS: Gentoo Linux (alsa-lib 1.1.2, alsa-plugins-1.1.1, Linux kernel 4.9.64)

If you suspect that newer alsa-libs are required, would the alsa.c plugin 
from the 7. Oct. 2017 be a good version for me to switch back as it
seems to be a good version before a larger rewrite?

Alexey-Yakovenko:
I suggest using the revision which works best for you. I'll look into this issue as soon as possible.

Lithopsian:
I see it too.  Uncontrolled looping at line 617.  Probably just needs a short sleep.

DaarkWel:
Same here, kinda annoying.
I tried with alsa.c version 91124bf and it works without high cpu load but presented some sound glitches every minute or so. Maybe because I'm using system-wide alsa-jack plugin, but with latest alsa.c version there's no glitches, yet there is 100% cpu load.

JulianLai:
Same here with deadbeef-git r7562.2f471821-1

wdlkmpx:
I also see this issue.. the problem seems to be gone if i insert ```usleep (10000);``` in line 625

Bfgeshka:
Problem bugs be a bit too. OSS output is ok as workaround, but OSS to ALSA bridge works not so good with mixing.

JulianLai:
@Alexey-Yakovenko wdlkmpx says that he has solved the problem.

Alexey-Yakovenko:
Sorry, I didn't have much time to fix this.

[...]

PR #1949, opened 2017-12-26
alsa: fix for very hires audio

wdlkmpx:
This is (apparently) an intelligent approach at fixing issue #1927  ..

Taking into account what JulianLai reported, it would seem the higher the samplerate the faster the looping.. so increase resource usage according to some key samplerate values..

It works for me with some test files, but more testing is needed to refine this or until a proper fix is found.


wdlkmpx:
Yeah youre right, it's not the best solution. 

Just in case... The most generic value seems to be 4000, there is no stutter while playing 192khz files.. and it still doesnt produce a high cpu load ..

PR #1993, opened 2018-03-29
Fixed GTK3 warning

TotalCaesar659:
There are many of them, but I've tried to make one first. I hope you like it.

### DIFF
--- a/plugins/gtkui/widgets.c
+++ b/plugins/gtkui/widgets.c
@@ -4142,11 +4142,7 @@ ddb_gtkui_widget_t *
 w_playtb_create (void) {
     w_playtb_t *w = malloc (sizeof (w_playtb_t));
     memset (w, 0, sizeof (w_playtb_t));
-#if GTK_CHECK_VERSION(3,0,0)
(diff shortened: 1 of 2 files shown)


plugins/gtkui/widgets.c
```diff
@@ -4142,7 +4142,11 @@ ddb_gtkui_widget_t *
 w_playtb_create (void) {
     w_playtb_t *w = malloc (sizeof (w_playtb_t));
     memset (w, 0, sizeof (w_playtb_t));
+#if GTK_CHECK_VERSION(3,0,0)
+    w->base.widget = gtk_box_new (GTK_ORIENTATION_HORIZONTAL, 0);
```
Alexey-Yakovenko: I think it would make sense to define GTK_ORIENTATION_HORIZONTAL as FALSE in support.h inside of `#if !GTK_CHECK_VERSION(3,0,0)` block, and then replace this code with just `w->base.widget = gtk_box_new (GTK_ORIENTATION_HORIZONTAL, 0);` without ifdefs.

PR #2009, opened 2018-05-08
Correct signed-unsigned char conversion.

Vascom:
This correction need to build deadbeef on ARM and PPC systems.

### DIFF
--- a/plugins/adplug/adplug/s3m.cpp
+++ b/plugins/adplug/adplug/s3m.cpp
@@ -26,7 +26,7 @@
 #include <string.h>
 #include "s3m.h"
 
-const char Cs3mPlayer::chnresolv[] =	// S3M -> adlib channel conversion
+const signed char Cs3mPlayer::chnresolv[] =	// S3M -> adlib channel conversion
   {-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,0,1,2,3,4,5,6,7,8,-1,-1,-1,-1,-1,-1,-1};
 
 const unsigned short Cs3mPlayer::notetable[12] =		// S3M adlib note table
--- a/plugins/adplug/adplug/s3m.h
+++ b/plugins/adplug/adplug/s3m.h
@@ -92,7 +92,7 @@ class Cs3mPlayer: public CPlayer
   char filetype[30];
 
  private:
-  static const char chnresolv[];
+  static const signed char chnresolv[];
   static const unsigned short notetable[12];
   static const unsigned char vibratotab[32];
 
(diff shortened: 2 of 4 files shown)


Vascom:
Can you do it as you see it?

Alexey-Yakovenko:
I thought it would be easy enough for you to do it, since you're already on it. But ok.

Alexey-Yakovenko:
Have you tested that s3m and sid formats still work correctly after your change?
This code actually compiles and works fine for ARM (tested on android), but looking closely at your change, it seems like it will get wrong values at runtime.

Vascom:
I am not tested it.
You can look at compile errors here http://koji.rpmfusion.org/kojifiles/work/tasks/4507/224507/build.log
and here http://koji.rpmfusion.org/kojifiles/work/tasks/4514/224514/build.log
It is for armv7hl but same errors for aarch64, ppc64 and ppc64le.

Alexey-Yakovenko:
```
#include <stdio.h>
#include <stdint.h>

int main () {
    int8_t signedValue = '\x80'; // <-- the original code
    int castFromSigned = signedValue;
    printf ("%d\n", castFromSigned); // <-- prints "-128"

    uint8_t unsignedValue = '\x80'; // <-- your modification (char is unsigned on ARM)
    int castFromUnsigned = unsignedValue;
    printf ("%d\n", castFromUnsigned); // <-- prints "128"
}```

Vascom:
OK. So I suggest you corect code. You better know this :)

Alexey-Yakovenko:
Ok so the errors you're getting is because of your compiler settings.

There are two problems with your pull request:

1. the code would not work correctly on ARM after you change
2. you would start getting the same compile warnings on x86 architectures

I'd happily fix the problem, but I don't have a possibility to test that the modified code works correctly on ARM atm.

Vascom:
What compiler settings causes this errors?

Alexey-Yakovenko:
`-Wnarrowing` is the one.

Vascom:
As I understand it from `-Wall` compiler directive. And I am don't think that after removing it an compile without errors program will work correctly. So it is better correct code and don't touch compiler.

Alexey-Yakovenko:
I'm not suggesting touching the compiler. I'm saying that this program code is valid and works, but obviously you can break it by using -Wall and treat warnings as errors.

It can be fixed, but it would need extensive testing before I can merge that. The fix would not be as simple as it looks.

Vascom:
May be make explicit type conversion `like static_cast<signed char>('\x80')`?
This should not break down the work.

Alexey-Yakovenko:
commits 394536bd135dd5f9fac211c391ce2e1a04bda6da and 8fdeb8102d7a8be6773dc4accc5fe87901488724 should fix your problems

Vascom:
Thanks.
Also need add #include <cstdint> to s3m.h.

Alexey-Yakovenko:
Are you getting a compile error without adding this include?
Cause it builds fine here on both macOS and Linux.

Alexey-Yakovenko:
Ah.. that's before I already fixed that with another commit 7163099d55c64bf11c3dac4dadfd83a0495ea207

Vascom:
OK

PR #2014, opened 2018-05-17
gtkui: move wingeom_restore call after window_init_hooks

saivert:
this is to account for the client side decoration added by my headerbar plugin so window is positioned correctly
it also makes more sense to have this right before gtk_widget_show


Alexey-Yakovenko:
This is a very risky change, that may break a lot of stuff.. but we still have time to revert it during the beta.

PR #2024, opened 2018-06-08
gtkui: another round of transient-for fixes

saivert:
This resolves popup menu positioning on Wayland.

### DIFF
--- a/plugins/gtkui/ddbtabstrip.c
+++ b/plugins/gtkui/ddbtabstrip.c
@@ -1012,7 +1012,8 @@ on_tabstrip_button_press_event(GtkWidget      *widget,
     }
     else if (TEST_RIGHT_CLICK(event)) {
         GtkWidget *menu = gtkui_create_pltmenu (tab_clicked);
-        gtk_menu_popup (GTK_MENU (menu), NULL, NULL, NULL, widget, 0, gtk_get_current_event_time());
+        gtk_menu_attach_to_widget (GTK_MENU (menu), GTK_WIDGET (widget), NULL);
+        gtk_menu_popup (GTK_MENU (menu), NULL, NULL, NULL, NULL, 0, gtk_get_current_event_time());
     }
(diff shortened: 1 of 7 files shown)


plugins/gtkui/plcommon.c
```diff
@@ -646,12 +646,12 @@ find_popup (GtkWidget *widget)
 
 static DdbListview *
 get_context_menu_listview (GtkMenuItem *menuitem) {
-    return DDB_LISTVIEW (g_object_get_data (G_OBJECT (find_popup (GTK_WIDGET (menuitem))), "ps"));
+    return DDB_LISTVIEW (g_object_get_data (G_OBJECT (gtk_widget_get_parent (GTK_WIDGET (menuitem))), "ps"));
```
saivert: Since we now call `gtk_menu_attach_to_widget` to fix transient-for issues with popup menus, this code is incorrect. We know we just want the parent here (the menu) so use `gtk_widget_get_parent` directly.

plugins/gtkui/ddbtabstrip.c
```diff
@@ -1012,7 +1012,8 @@ on_tabstrip_button_press_event(GtkWidget      *widget,
     }
     else if (TEST_RIGHT_CLICK(event)) {
         GtkWidget *menu = gtkui_create_pltmenu (tab_clicked);
-        gtk_menu_popup (GTK_MENU (menu), NULL, NULL, NULL, widget, 0, gtk_get_current_event_time());
+        gtk_menu_attach_to_widget (GTK_MENU (menu), GTK_WIDGET (widget), NULL);
+        gtk_menu_popup (GTK_MENU (menu), NULL, NULL, NULL, NULL, 0, gtk_get_current_event_time());
```
saivert: widget was passed as data param to gtk_menu_popup but is unused since we don't pass a GtkMenuPositionFunc. cleaned it up.

PR #2026, opened 2018-06-26
po: Fix pl.po.

orbea:
I noticed that with `msgfmt -c` the `po/pl.po` file has two errors. I think the correct way to fix them is with two percentage signs to make it literal. At least `msgfmt -c` no longer complains.

The current error:
```
$ msgfmt -c po/pl.po 
po/pl.po:2345: format specifications in 'msgid' and 'msgstr' for argument 1 are not the same
po/pl.po:2354: format specifications in 'msgid' and 'msgstr' for argument 1 are not the same
msgfmt: found 2 fatal errors
```
Reference: https://www.gnu.org/software/gettext/manual/html_node/gcc_002dinternal_002dformat.html#gcc_002dinternal_002dformat

### DIFF
--- a/po/pl.po
+++ b/po/pl.po
@@ -2341,8 +2341,8 @@ msgstr "Przewiń 5 sekund do tyłu"
 
 #: ../translation/plugins.c:176
 #, c-format
-msgid "Seek 1% Forward"
-msgstr "Przewiń o 1% do przodu"
+msgid "Seek 1%% Forward"
+msgstr "Przewiń o 1%% do przodu"
 
 #: ../translation/plugins.c:177
 msgid "Seek 1% Backward"
@@ -2350,8 +2350,8 @@ msgstr "Przewiń o 1% do tyłu"
 
 #: ../translation/plugins.c:178
 #, c-format
-msgid "Seek 5% Forward"
-msgstr "Przewiń o 5% do przodu"
+msgid "Seek 5%% Forward"
+msgstr "Przewiń o 5%% do przodu"
 
 #: ../translation/plugins.c:179
 msgid "Seek 5% Backward"


orbea:
I'm not sure what to do about that `build.sh` failure, I think the error message means the `wget` command failed...

Alexey-Yakovenko:
I think this would be a very wrong fix.
The reason for this problem is that the po file specifies "c-format" for this line, which is wrong.
The correct solution is to remove c-format line.

Alexey-Yakovenko:
(you can't solve this just by modifying a single o file.. it should be done somewhere else)

orbea:
Yes, removing the comment works, but at the same time the online documentation says those comments should not be modified by hand. See these references.
> c-format
> no-c-format

 > These flags should not be added by a human. Instead only the xgettext program adds them. In an automated PO file processing system as proposed here, the user’s changes would be thrown away again as soon as the xgettext program generates a new template file.
>
> The c-format flag indicates that the untranslated string and the translation are supposed to be C format strings. The no-c-format flag indicates that they are not C format strings, even though the untranslated string happens to look like a C format string (with ‘%’ directives).
>
> When the c-format flag is given for a string the msgfmt program does some more tests to check the validity of the translation. See msgfmt Invocation, c-format Flag and c-format.

https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html#PO-Files
https://www.gnu.org/software/gettext/manual/html_node/c_002dformat-Flag.html

Would you mind making a suggestion on how to get past this? I was trying to fix `autoreconf` and ended up exposing this issue, it should be fixed before I can do anything else. :)

Alexey-Yakovenko:
I don't know yet.. The _() macro thinks they are format strings..
`% F` seems to be a valid printf sequence.. so there's that.
Will need to dig, but don't have time this week.

Alexey-Yakovenko:
Made the fix. please let me know if it works. I don't have access to a linux machine atm.

orbea:
That worked here, thanks!

PR #2031, opened 2018-07-02
Move from deprecated function get_plugin_dir()

kuba160:
Switch from deprecated `get_plugin_dir` to `get_system_dir` in plugins.c. Additionally navigate to this function in `deadbeef.h`.

### DIFF
--- a/deadbeef.h
+++ b/deadbeef.h
@@ -694,6 +694,7 @@ typedef struct {
     // system folders
     // normally functions will return standard folders derived from --prefix
     // portable version will return pathes specified in comments below
+    // DEPRECATED IN API LEVEL 8, use get_system_dir instead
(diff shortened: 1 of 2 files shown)


deadbeef.h
```diff
@@ -694,6 +694,7 @@ typedef struct {
     // system folders
     // normally functions will return standard folders derived from --prefix
     // portable version will return pathes specified in comments below
+    // DEPRECATED IN API LEVEL 8, use get_system_dir instead
```
Alexey-Yakovenko: This comment is quite redundant, since all deprecated APIs are clearly marked as such.

kuba160: I think this comment is useful because replacement function `get_system_dir` is located 500 lines after it and finding it can be sometimes problematic (for example when somone searches for 'config').

PR #2035, opened 2018-07-18
Move fopen out of pl_lock in decoders

### LINKED ISSUE #2034: decoder->init function locks on fopen
Sometimes I get a little freeze just before playing next track, a half second pause about 3 seconds near the end of the current song.
I'm playing music from a very populated partition (using ntfs-3g) so I suspected the slow access was the cause, but why reading ahead could block the playing thread and GUI?
I tracked the issue to this code from the mp3 plugin in function cmp3_init:
```
deadbeef->pl_lock ();
info->buffer.file = deadbeef->fopen (deadbeef->pl_find_meta (it, ":URI"));
deadbeef->pl_unlock ();
```
I think it should only lock pl_find_meta not fopen. This piece of code is replicated for reading tags in each plugin too.
It also happens while I'm playing one song and I select another from the playlist and making it sure that it's from other folder (not cached or opened before), if I do a right click on it and press "Reload Metadata", here again the GUI and the sound stop while accessing the file.
No sure but this could become evident if you're playing files from a network share.

Moving the locks out of fopen solved both sound issues (#2035), although clicking "Reload Metadata" still freeze the GUI because the code runs on the same thread, and that's another story.

BorisCarvajal:
fix #2034 

I have a question, why WMA (and v2m) decoder doesn't use pl_lock when calling pl_find_meta function while the others do?

### DIFF
--- a/junklib.c
+++ b/junklib.c
@@ -4982,9 +4982,7 @@ error:
         free (buffer);
     }
     if (!err) {
-        pl_lock ();
         rename (tmppath, fname);
-        pl_unlock ();
     }
     else {
         unlink (tmppath);
--- a/plugins/aac/aac.c
+++ b/plugins/aac/aac.c
@@ -27,6 +27,7 @@
 #include <stdlib.h>
 #include <math.h>
 #include "../../deadbeef.h"
+#include "../../strdupa.h"
 #include "aac_parser.h"
 
 #include "mp4ff.h"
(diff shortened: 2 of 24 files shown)


Alexey-Yakovenko:
hmmm.. writing an explanation how I want to do it would take longer than doing it myself, so I'll just merge your PR, and do a followup fix.

BorisCarvajal:
strdupa was used in ffmpeg for storing the URI so I copied that approach.

Issue #1145, opened 2015-03-12
Subgrouping


Alexey-Yakovenko:
Original [issue 1243](https://code.google.com/p/ddb/issues/detail?id=1243) created by Alexey-Yakovenko on 2014-12-07T00:17:00.000Z:

The playlist could be easier readable with subgroups.

Like:

M ---------------------------
 Metallica ------------------
  Kill 'Em All --------------
   01.mp3
   02.mp3
  Ride The Lightning --------
   01.mp3
   02.mp3
  Load ----------------------
   01.mp3
   02.mp3

 Machine Head ---------------
  Burn My Eyes --------------
   01.mp3
   02.mp3

N ---------------------------
 Nero's Day at Disneyland ---
  From Rotting Fantasyland --
   01.mp3
   02.mp3

Having 3 bars (letter + artist + album) helps identifying change of artist in the playlist. 2 bars change of artist and 1 of album.
I indented it here since it's only text, it might not be necessary.

PR #2044, opened 2018-08-31
Implement subgroups for GTK

ToadKing:
This is a first pass at implementing #1145. It allows for an arbitrary number of nested groups and allows for formats that allow for groups with empty/invisible titles.

screenshot from 2018-08-30 22-10-12

Current Implementation List:
- [x] Parse groups to make sublists
- [x] Render subgroups correctly
- [x] Detect entries with mouse events correctly with subgroups
- [x] Implement pinning for subgroups - currently gets drawn/written over due to render changes
- [x] Draw artwork correctly - currently gets drawn/written over due to render changes
- [x] Configurable indentation for sublevels
- [x] Draw subgroups with artwork fields correctly - currently does not adjust its position to wrap around it
- [x] Add pinning to all subgroups, not just one

These two will be in a future PR:
- Change the way subgroups are stored in preferences to allow for unique TF sequences per group, not using a delimiting token.
- Make a new "Group By" dialog exposing the new functionality.

Questions:
- ~~How do we want to store subgroups in settings? Right now I pack it into the normal group setting and use a pipe `|` to separate them.~~ **This will be changed to store them as unique strings in settings in a future PR. For the time being, `|||` will be used instead.**
- ~~How do we want pinning to function? Does it pin every subgroup, or only the top-level one?~~ **Changed to allow you to configure which level you want pinned.**
- ~~Same with artwork. Should it only be displayed with the top level group? Or be configurable somehow?~~ **Changed it to display with the configured pinned group.**
- ~~How should we draw subgroups around artwork if the artwork column is not the easy case of being the first or last column?~~ **Added detection of which side of the playlist artwork is on.**

### DIFF
--- a/plugins/gtkui/ddblistview.c
+++ b/plugins/gtkui/ddblistview.c
@@ -125,7 +125,7 @@ ddb_listview_list_render_row_background (DdbListview *ps, cairo_t *cr, DdbListvi
 static void
 ddb_listview_list_render_row_foreground (DdbListview *ps, cairo_t *cr, DdbListviewIter it, int even, int idx, int y, int w, int h, int x1, int x2);
 static void
-ddb_listview_list_render_album_art (DdbListview *ps, cairo_t *cr, DdbListviewGroup *grp, int pinned, int grp_next_y, int y, GdkRectangle *clip);
+ddb_listview_list_render_album_art (DdbListview *ps, cairo_t *cr, DdbListviewGroup *grp, int min_y, int grp_next_y, int y, GdkRectangle *clip);
 static void
 ddb_listview_list_track_dragdrop (DdbListview *ps, int x, int y);
 int
@@ -884,11 +884,13 @@ fill_list_background (DdbListview *listview, cairo_t *cr, int x, int y, int w, i
     render_treeview_background(listview, cr, FALSE, TRUE, x, y, w, h, clip);
 }
 
-static void
-ddb_listview_list_render_subgroup (DdbListview *listview, cairo_t *cr, GdkRectangle *clip, DdbListviewGroup *grp, int idx, int grp_y, const int cursor_index, const int current_group_depth, int title_offset, const int subgroup_artwork_offset) {
+static int
+ddb_listview_list_render_subgroup (DdbListview *listview, cairo_t *cr, GdkRectangle *clip, DdbListviewGroup *grp, int idx, int grp_y, const int cursor_index, const int current_group_depth, int title_offset, const int subgroup_artwork_offset, const int pin_offset) {
     const int scrollx = -listview->hscrollpos;
     const int row_height = listview->rowheight;
     const int total_width = listview->totalwidth;
+
+    int subgroup_pinned_height = 0;
 
     // find 1st group
     while (grp && grp_y + grp->height < clip->y) {
@@ -896,10 +898,10 @@ ddb_listview_list_render_subgroup (DdbListview *listview, cairo_t *cr, GdkRectan
         idx += grp->num_items;
         grp = grp->next;
     }
-    DdbListviewGroup *pin_grp = gtkui_groups_pinned && grp && grp_y < 0 && grp_y + grp->height >= 0 ? grp : NULL;
 
     while (grp && grp_y < clip->y + clip->height) {
         const int title_height = grp->group_label_visible ? listview->grouptitle_height : 0;
+        const int is_pinned = gtkui_groups_pinned && grp_y < pin_offset && grp_y + grp->height >= 0;
 
         // only render list items when at the deepest group level
         if (!grp->subgroups) {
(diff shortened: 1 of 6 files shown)


ToadKing:
I finished implementing all the features I wanted.

* The subgroup label will appear on the opposite side of the playlist the artwork column is, to avoid being covered.
* You can configure which subgroup level artwork should be drawn for. This is also the group level that will be pinned when pinning is enabled.
* Right now there is no GUI to configure the artwork level or title padding. I coulnd't get the glide project to open. :(

cboxdoerfer:
> Right now there is no GUI to configure the artwork level or title padding. I coulnd't get the glide project to open. :(

DeaDBeeF uses a modified glade version, it's bundled with with the source code: https://github.com/DeaDBeeF-Player/deadbeef/tree/master/tools/glade

plugins/gtkui/plcommon.c
```diff
@@ -1275,20 +1275,39 @@ groups_changed (DdbListview *listview, const char *format)
...
+    char *saveptr;
+    char *token = strtok_r(mutable_format, "|", &saveptr);
+
+    // always have at least one
+    listview->group_formats = calloc(sizeof(DdbListviewGroupFormats), 1);
+    fmt = listview->group_formats;
+    fmt->group_format = strdup (token ? token : "");
+    fmt->group_title_bytecode = deadbeef->tf_compile (fmt->group_format);
+
+    while ((token = strtok_r(NULL, "|", &saveptr)) != NULL) {
```
cboxdoerfer: This will probably cause a lot of issues. Say the user enters a valid tf expression, e.g. `%album% $if($greater(%totaldiscs%,1),| Disc %discnumber%)', your code then breaks that valid expression apart into '%album% $if($greater(%totaldiscs%,1),' and ' Disc %discnumber%)' which will most likely fail to evaluate.

I guess there are three ways how this can be fixed:

1) splitting by '|' happens after calling 'tf_eval'
2) the title format engine itself learns about the '|' syntax
3) sub groups aren't parsed from a single group format string, but each group gets its own group format

cboxdoerfer:
There's also an issue with pinned sub groups. Currently only the sub group with the greatest depth is pinned at the top, when ideally you'd want to display each level. Like so: 
subgroups

plugins/gtkui/ddblistview.c
```diff
@@ -543,13 +546,13 @@ ddb_listview_destroy(GObject *object)
...
-        free (listview->group_format);
-        listview->group_format = NULL;
-    }
-    if (listview->group_title_bytecode) {
-        free (listview->group_title_bytecode);
-        listview->group_title_bytecode = NULL;
+    DdbListviewGroupFormats *fmt = listview->group_formats;
+    while (fmt) {
+        DdbListviewGroupFormats *next_fmt = fmt->next;
+        free (fmt->group_format);
```
Alexey-Yakovenko: it's worth to add NULL checks before free.

plugins/gtkui/ddblistview.c
```diff
@@ -687,61 +702,35 @@ ddb_listview_is_empty_region (DdbListviewPickContext *pick_ctx)
 // input: absolute y coord in list (not in window)
```
Alexey-Yakovenko: This comment belongs to the function `ddb_listview_list_pickpoint`, and should be moved.

plugins/gtkui/ddblistview.c
```diff
@@ -3244,17 +3342,52 @@ ddb_listview_min_group_height(DdbListviewColumn *columns) {
...
     listview->binding->ref(it);
     DdbListviewGroup *grp = calloc(1, sizeof(DdbListviewGroup));
     grp->head = it;
+    grp->group_label_visible = group_label_visible;
     return grp;
 }
 
+static int
+calc_subgroups_height(DdbListviewGroup *grp) {
+    DdbListviewGroup *next = grp;
```
Alexey-Yakovenko: This code looks awkward. You don't need to make a new `next` variable, just use grp in the loop:
`while (grp) { ...`

plugins/gtkui/ddblistview.c
```diff
@@ -3264,35 +3397,102 @@ calc_group_height(DdbListview *listview, DdbListviewGroup *grp, int min_height,
...
+        }
+        free(group_titles);
+    }
+    // no groups fast path
+    else {
+        for (DdbListviewGroup *grp = listview->groups; grp; grp = grp->next) {
+            do {
+                grp->num_items++;
+                it = next_playitem(listview, it);
+            } while (it && grp->num_items < BLANK_GROUP_SUBDIVISION);
```
Alexey-Yakovenko: A note to myself: verify that this optimisation is still functional.

plugins/gtkui/ddblistview.c
ToadKing: free(NULL) is a no-op so the extra check isn't necessary. Is there a platform where the check is required?

Alexey-Yakovenko: ok, good point. I still have that habit from the old days, when all runtimes have crashed from that.

plugins/gtkui/plcommon.c
```diff
@@ -2021,13 +2097,43 @@ pl_common_col_sort (int sort_order, int iter, void *user_data) {
...
     deadbeef->conf_unlock ();
     parser_unescape_quoted_string (format);
-    listview->group_format = format;
-    listview->group_title_bytecode = deadbeef->tf_compile (listview->group_format);
+    listview->group_formats = NULL;
+
+    DdbListviewGroupFormats *fmt = NULL;
+    char *saveptr = NULL;
+    char *token;
+    while ((token = strtok_stringdelim_r(format, SUBGROUP_DELIMITER, &saveptr)) != NULL) {
```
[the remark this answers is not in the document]
ToadKing: Yeah, that loop is kinda ugly since it has to deal with setting the head of the linked list and the first entry could be skipped because its empty.

PR #2055, opened 2018-10-15
Update README.md

laichiaheng:
libjansson-dev need to be installed.


Alexey-Yakovenko:
Dependencies are covered in the README (the one without .md extension). It already has jansson in the list.
libjansson-dev is not a distribution-agnostic name of the package, many distributions have it under different name, so it doesn't make sense to include it this way into the main README.md file.
Perhaps it makes sense to make a debian/ubuntu specific page with the specific package names.

PR #2057, opened 2018-10-21
alsa: fix a missing mutex unlock

huglovefan:
it's only reached when setting the format fails but at least the player doesn't freeze when that happens now


Alexey-Yakovenko:
Thanks for spotting this!

PR #2061, opened 2018-10-28
Fix scrolling with arrow keys with pinned nested subgroups

ToadKing:
A small bug that slipped through the cracks in #2044. When scrolling through a playlist with arrow keys, the playlist scrolling doesn't take the height of nested subgroup labels into account, sometimes hiding the selected track under a pinned label. This fixes the bug by making the whole height of all the subgroups into account for calculating the scroll position.


Alexey-Yakovenko:
@ToadKing for the future: please prefix commits with the name of the area you're changing (should have been prefixed with `gtkui:` in this case)

PR #2065, opened 2018-11-15
Fix use-after-free bug

kevinvandervlist:
I scanned the code base with the tool [cppcheck](http://cppcheck.sourceforge.net/), and I discovered one interesting finding:

```
[gtkui/plcommon.c:1311]: (error) Dereferencing 'fmt' after it is deallocated / released
```

Given that a local variable was already declared and pointing to the next element, I assume this dereference is a small mistake that is easily corrected.


Alexey-Yakovenko:
Thanks

PR #2075, opened 2019-01-11
lastfm: fix scrobbling when skipping tracks and playing the same track

spikeh:
Fixes #2075
I attempted to fix both of these problems. Everything works as expected but I am not sure if there are unintended consequences of not setting playing_track to NULL first in _play_track.


Alexey-Yakovenko:
I can't accept this change, since I know for sure it will break things.
I will have a look at the problem as soon as I have time.

PR #2078, opened 2019-01-19
*regression* - actually make the back button play the previous track

ghost:
Since commit a7d600a7, the back button actually acts as a clone of the forward button.

This commit resolves this for me locally.


tomjepp:
Thanks, the message was slightly sarcastic - I was very confused when I found it!

Alexey-Yakovenko:
Sorry for that.. it was an unfortunate typo when refactoring.

PR #2089, opened 2019-03-11
tf: Allow arbitrary strings in %tracknumber% and %track number%

### LINKED ISSUE #2062: Allow %tracknumber% with non-digit characters
I have a vinyl rip with two tracks on each side. The track numbers in FLAC metadata are prefixed with a side letter: A01, A02, B01, and B02.

While foobar2000 allows such "numbers" for %tracknumber%, DeaDBeeF gives an empty string in such cases: https://github.com/DeaDBeeF-Player/deadbeef/blob/3b8b2fae56e43749cf46c42463e282548a73c315/tf.c#L1959-L1962

This obviously gives issues when converting files with default "%tracknumber% - %title%" format. In my particular case, two tracks on the rip are named "Untitled", so when I try to convert FLACs to MP3s, two files have the same name (" - Untitled.mp3"), and depending on settings, the second track is either skipped or will overwrite the first one.

danpla:
Fixes #2062.

### DIFF
--- a/tf.c
+++ b/tf.c
@@ -1839,6 +1839,26 @@ format_playback_time (char *out, int outlen, float t) {
     return len;
 }
 
+// Get the bounds of %tracknumber% and %track number% from "track" metadata.
+// This is everything till the end of the string or till the first slash (to
+// extract from "current/total"), with stripped whitespace.
+static void tracknumber_bounds(const char *s, const char **start, const char **end) {
+    while (isspace (*s)) {
+        ++s;
+    }
+
+    *start = s;
+    *end = s;
+
+    while (*s && *s != '/') {
+        if (!isspace (*s)) {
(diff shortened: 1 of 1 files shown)


Alexey-Yakovenko:
This change breaks the existing functionality, see the [build log](https://api.travis-ci.org/v3/job/504811022/log.txt) for details.

Alexey-Yakovenko:
A simpler and more reliable approach would be to check if the first character in `track` field is not a digit, and in this case leave the text alone without any transformation.

danpla:
Digits may be followed by non-digit characters, in which case `atoi()` will silently drop them.

I'll fix the failures soon.

Alexey-Yakovenko:
I applied a simpler fix, added tests, and validated the behavior.

PR #2096, opened 2019-03-14
cocoaui: Shrink album art to fit if it is taller than the reserved space

### LINKED ISSUE #2095: macOS: album art gets cropped vertically if it is taller than it is wide
1. Add a track with a portrait-shaped piece of album art to the playlist
2. Observe that the album art is cropped - the bottom is missing

The size of album art isn't known when group heights are calculated.

As such, the minimum reserved space for album art is as wide as the art column, and as tall as the art column is wide.

This means the album art gets cut off.

However - if the list of tracks is taller than the minimum, the art may be displayed full size.

Deadbeef version: master 9b369519
OS: macOS 10.13.4

tomjepp:
Attempts to resolve #2095 macOS: album art gets cropped vertically if it is taller than it is wide.

I looked at resizing the groups as the art gets loaded but this required me to invalidate lots of the view each time and worked out *very* slow. If this would be possible in a performant way, I'd prefer to do it however!

As an alternative, try shrinking album art vertically. It's not perfect, but at least all of the art is visible.

### DIFF
--- a/plugins/cocoaui/DdbPlaylistViewController.m
+++ b/plugins/cocoaui/DdbPlaylistViewController.m
@@ -801,7 +801,16 @@ static void coverAvailCallback (NSImage *__strong img, void *user_data) {
 
     NSSize size = [image size];
 
-    [image drawInRect:NSMakeRect(art_x, ypos, art_width, size.height / (size.width / art_width))];
+    int render_width = art_width;
+    int render_height = size.height / (size.width / art_width);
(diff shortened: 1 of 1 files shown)


tomjepp:
Pushed a revised version - art_height was taller than the space available (and was cropping the bottom off of images still), whereas grp_next_y - min_y gives an accurate amount of space and images are no longer cropped.

PR #2103, opened 2019-03-17
tf: implement $stripprefix and $swapprefix

ToadKing:
The prefix comparison check is case-insensitive, matching the behavior of foobar2000.

I was going back and forth on whether to evaluate the prefix arguments as they're processed or evaluate them all up front first, but decided on the latter. It makes reusing code for the single parameter case easier plus it's easier to catch format errors in the prefix arguments if they exist if they're all processed every time.


Alexey-Yakovenko:
Merged manually with added tests

PR #2114, opened 2019-04-09
Estonian translation

Jalakas:


Alexey-Yakovenko:
There are a few problems with this PR:

1. I don't accept patches to external dependencies, unless they fix bugs in the code. Every change in the external dependencies increases complexity of merging upstream changes in the future. If you want to fix comments in all those various libraries -- you should submit patches upstream. Most of the changes you posted are like this.

2. interface.c is a generated file, doesn't matter if I merge this -- the change would get overwritten on the next UI change.

3. I **normally** don't accept translations directly via github. Translations are performed on Transifex [here](https://www.transifex.com/deadbeef-player/deadbeef-player/dashboard/). However, since this translation was empty before, I can make an exception, since this should not cause conflicts. Another problem with this is that normally translations are performed on release branch (1.8), while you did it on master, so some translations would be lost, and some would be missing. I can't do much about this.

I can't take it as-is, but when I have some time -- I can graft the po file from it.

Jalakas:
Uploaded my po translations to transifex. 
It would be nice to have some lines about translations and link to transifex project in README.

Can close this pull request now.

Alexey-Yakovenko:
Hm.. I realized that the only place this is mentioned is the [wiki FAQ](https://github.com/DeaDBeeF-Player/deadbeef/wiki/FAQ#i-want-to-help-with-translation-how-do-i-do-that)

I will definitely add this to the CONTRIBUTING.md

PR #2146, opened 2019-05-21
gtkui: Play next song when current removed from disc

kpcee:
If the currently played song is deleted from disk via the GUI, deadbeef will continue playing it. The patch will jump to the next song or stop playback when the end of the playlist is reached.

### DIFF
--- a/plugins/gtkui/actionhandlers.c
+++ b/plugins/gtkui/actionhandlers.c
@@ -389,6 +389,10 @@ action_delete_from_disk_handler_cb (void *data) {
     }
     deadbeef->pl_lock ();
 
+    // Saving the current song index
+    DB_playItem_t *it = deadbeef->streamer_get_playing_track ();
+    int index_old = deadbeef->pl_get_idx_of (it);
+
     if (ctx == DDB_ACTION_CTX_SELECTION) {
         DB_playItem_t *it = deadbeef->plt_get_first (plt, PL_MAIN);
         while (it) {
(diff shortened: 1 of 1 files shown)


Alexey-Yakovenko:
Can you please describe in more detail what are you trying to do with this patch? I couldn't figure it out.

Alexey-Yakovenko:
I think that your intention is to interrupt playback as soon as the currently playing track is deleted, and jump to whatever the next track is going to be. However the way you did it won't be able to handle most normal conditions. For example, I think this approach would ignore the playback queue, shuffle and repeat modes, and so on.

Another concern is the change of the existing behavior, which some people may be relying upon.

Perhaps a better and more flexible approach would be to add a new action for this, called "delete from disk and skip to next track", which would still need to skip the track correctly.

I can't accept this PR as it is, and I'm not sure how hard or easy is it to implement properly.

kpcee:
Thanks for watching the patch. Most of the problems mentioned above could be solved by adjusting the patch and replacing `DB_EV_PLAY_CURRENT` with
`deadbeef->sendmessage (DB_EV_NEXT, 0, 0, 0);`

I can't speak for other users, but when I delete the current song from the playlist and hard drive, I don't necessarily want to hear it to the end.
Anyway, thanks for your feedback.

Alexey-Yakovenko:
yeah, I DB_EV_NEXT should work, but should be performed before deleting the track.

Alexey-Yakovenko:
however, if deleting fails - the behavior would still look weird.
there will be many issues that would have to be sorted out.

kpcee:
If the currently played song has been removed from the playlist `deadbeef->pl_get_idx_of (it)` returns the value -1, but okay this only works after deletion.

kpcee:
I've thought about the problem again and I have to agree that it is a complicated matter. But I came up with a very simple solution that doesn't change the previous procedures. Instead of selecting the next song myself and making mistakes, I wind the currently played song to the end and deadbeef jumps to the next song all by itself.

The behaviour only happens if the user sets the variable "gtkui.play_deleted_tacks_completely" to zero, default is the previous behaviour. If the currently played song is deleted, it has the index -1 and the current song is rewound to the end.
I'm happy with the solution, it works fine and maybe someone else will like it.

```
    if (deadbeef->get_output ()->state () == OUTPUT_STATE_PLAYING) {
        DB_playItem_t *it = deadbeef->streamer_get_playing_track ();
        int idx = deadbeef->pl_get_idx_of (it);
        int play_complete = deadbeef->conf_get_int ("gtkui.play_deleted_tacks_completely", 1);

        if (idx == -1 && play_complete == 0) {
            float pos = deadbeef->pl_get_item_duration (it);
            deadbeef->playback_set_pos (pos);
        }
    }
```

Alexey-Yakovenko:
Thanks for looking into it. I will reopen the issue and reinvestigate when I have some time.

Alexey-Yakovenko:
oh.. that was not an issue. perhaps one should be made.

PR #2147, opened 2019-05-22
gtkui: fix missing bits since switching from name to id for output_pl…

saivert:
…ugin config item


Alexey-Yakovenko:
This issue definitely needs fixing to ship 1.8.1, but the fix you made is incomplete.
The correct implementation needs to compare with both the id and the name, for compatibility reasons.

Please see how it's done in plugins.c.
```
        if (!strcmp (p->plugin.id, outplugname)
            || !strcmp (p->plugin.name, outplugname)) {
```

saivert:
Isn't it better to just rewrite the config item once during startup? Maybe this can be done in plugins.c. Once output_plugin has been rewritten with the ID then all the rest should work.

Alexey-Yakovenko:
> Isn't it better to just rewrite the config item once during startup? Maybe this can be done in plugins.c. Once output_plugin has been rewritten with the ID then all the rest should work.

Yes, this would also work.

saivert:
Seems like this is already done during startup. plug_reinit_sound is called on start at least once which leads to the output_plugin config item being set.

PR #2149, opened 2019-05-27
gtkui: refactor out gtkui_thread function as the UI never ran in its …

saivert:
…own thread and don't change process name

### DIFF
--- a/plugins/gtkui/gtkui.c
+++ b/plugins/gtkui/gtkui.c
@@ -801,9 +801,6 @@ add_mainmenu_actions_cb (void *data) {
 }
 
 int
-gtkui_thread (void *ctx);
-
-int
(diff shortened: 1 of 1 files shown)


Alexey-Yakovenko:
Why not just rename the function into something more meaningful? I don't like that the already large function gtkui_start becomes even larger.

saivert:
Is it really that large? It contains 6 lines from before. It should contain the init of gtk and then run the event loop. Rest of the startup code is in `gtkui_mainwin_init()`. There is at least very slim chance that is is going to grow further. Wouldn't this function contain code similar to this already if it wasn't for the fact that you very early in development tried out putting it in a thread?

Alexey-Yakovenko:
ok, right. I didn't see from the diff, that gtkui_start is 4 loc.

PR #2150, opened 2019-05-27
Replaygain dialogs enlarged

kpcee:
The Replaygain scanner dialogs are a bit small and could be enlarged. On the screenshot you can see the change, left as it is current and right new size.

Unbenannt

### DIFF
--- a/plugins/gtkui/deadbeef.glade
+++ b/plugins/gtkui/deadbeef.glade
@@ -10542,7 +10542,7 @@ Descending</property>
 </widget>
 
 <widget class="GtkWindow" id="rg_scan_progress">
-  <property name="width_request">440</property>
+  <property name="width_request">550</property>
   <property name="visible">True</property>
   <property name="title" translatable="yes">ReplayGain Scan Progress</property>
   <property name="type">GTK_WINDOW_TOPLEVEL</property>
@@ -10695,7 +10695,7 @@ Descending</property>
 </widget>
 
 <widget class="GtkWindow" id="rg_scan_results">
-  <property name="width_request">550</property>
+  <property name="width_request">900</property>
   <property name="height_request">350</property>
   <property name="visible">True</property>
   <property name="title" translatable="yes">ReplayGain Scan Results</property>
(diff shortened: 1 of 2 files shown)


Alexey-Yakovenko:
are you sure that simply changing the glade file does what you show on the screenshots?

kpcee:
The size of the dialogs can be found in the file deadbeef.glade. To make the changes visible, they have to be changed in the interface.c file. However, according to the header this is a dynamically generated file (DO NOT EDIT THIS FILE - it is generated by Glade). The normal build process doesn't care about changes in the deadbeef.glade file and I don't know exactly what to do that glade generates it. I should have mentioned that.

To see the changes live you have to change the file interface.c in the two lines.
```
5314: gtk_widget_set_size_request (rg_scan_progress, 550, -1);
5382: gtk_widget_set_size_request (rg_scan_results, 900, 350);
```

saivert:
You have to build the version of glade that is bundled in the deadbeef source under tools subdirectory. Then open deadbeef.glade using this verison of glade. Make the changes in glade, then use it to regenerate the source files.

Alexey-Yakovenko:
I can generate the file, that's not a problem, although I need to decide / think why this change is necessary. Perhaps there's a better solution, like saving user-adjusted window size.

kpcee:
[Here](https://github.com/DeaDBeeF-Player/deadbeef/compare/master...kpcee:replaygain_windows) is a working patch for all three dialogs, save / restore the sizes.

However, could someone please tell me how to compile glade? The instructions in the file "INSTALL" do not work. I'm sorry about the stupid question but I have absolutely no idea about C / GTK+ / automake, I use only C++ / Qt / qmake otherwise.

saivert:
To compile glade:

```
cd deadbeef-source/tools/glade
autoreconf -fi
./configure
make
```

Make sure you have dependencies installed like libxml2 development package.

kpcee:
That worked fine, thank you very much.

kpcee:
What is now preferred, the automatic saving of the dialog size or a change with Glade?

Alexey-Yakovenko:
Technically, I don't care about either. I think the size of 900px is fine. However, interface.c needs to be generated. Do you need me to do it for you?

kpcee:
If you know how to do it, it's no problem.

kpcee:
This pull request is now obsolete.

PR #2152, opened 2019-06-03
gtkui: use newer GTK popup menu apis for correct behavior under wayla…

saivert:
…nd when no widget is associated with the menu


saivert:
Due to the convoluted way menus are managed I think this is the best option for now to fix this nagging issue under Wayland. It is the least intrusive code change.

PR #2156, opened 2019-06-11
gtkui: fixes crash when removing playlist widget from layout

saivert:


Alexey-Yakovenko:
Let's merge this and see if anything bad happens. Better some fix than no fix.

Alexey-Yakovenko:
The reason why I don't like the fix is this:

when `queue_cover_callback(coverart_release, info);` is called, it would enqueue cover fetch operation on another thread, and eventually call coverart_release, from the same thread.

Then it would call `g_object_unref(info->listview->list);` from non-UI thread, which is basically a race condition.

The reason why it crashes without this fix is probably because GTK's mainloop was already gone when this was executed. However, I think now it might start crashing randomly, or leak references, or whatever else.

Alexey-Yakovenko:
Now that I read more carefully when it *was* crashing, I think the reason is something else.
Please mind that I don't run deadbeef on linux daily, and quite unlikely that I would notice any new issue.
Let me know if fixing / investigating is needed.

Issue #2005, opened 2018-04-23
Play, Pause, Change Volume, Change Playback Order from Command Line


davestechshop:
Hello. Does DeaDBeeF have commands that allow controlling play, pause, volume level and playback order from the Linux command line? If yes, where is the documentation? Thank you.

Deadbeef version: 0.7.2
OS: Arch Linux KDE

Bfgeshka:
It does, check `deadbeef --help`.

davestechshop:
Thanks. What about changing volume? I don't see that option listed. Is there a custom way to change volume from the command line?

deadbeef 0.7.2
Usage: deadbeef [options] [--] [file(s)]
Options:
   --help  or  -h     Print help (this message) and exit
   --quit             Quit player
   --version          Print version info and exit
   --play             Start playback
   --stop             Stop playback
   --pause            Pause playback
   --toggle-pause     Toggle pause
   --play-pause       Start playback if stopped, toggle pause otherwise
   --next             Next song in playlist
   --prev             Previous song in playlist
   --random           Random song in playlist
   --queue            Append file(s) to existing playlist
   --gui PLUGIN       Tells which GUI plugin to use, default is "GTK2"
   --nowplaying FMT   Print formatted track name to stdout
                      FMT %-syntax: [a]rtist, [t]itle, al[b]um,
                      [l]ength, track[n]umber, [y]ear, [c]omment,
                      copy[r]ight, [e]lapsed
                      example: --nowplaying "%a - %t" should print "artist - title"
                      for more info, see http://github.com/Alexey-Yakovenko/deadbeef/wiki/Title-formatting
                      NOTE: --nowplaying is deprecated.
   --nowplaying-tf FMT  Print formatted track name to stdout, using the new title formatting
                      FMT syntax: http://github.com/Alexey-Yakovenko/deadbeef/wiki/Title-formatting-2.0
                      example: --nowplaying-tf "%artist% - %title%" should print "artist - title"

Bfgeshka:
AFAIK, there is no way to change sound volume for deadbeef. Yet, you can manipulate overall system sound via cli.

Alexey-Yakovenko:
everything answered it seems? closing

kuba160:
option to change volume via command line came with commit ef8396e5f6fa9a53ed4c59e7a13737f6d63bbcc3, use master branch or wait for deadbeef 1.8 release.

PR #2161, opened 2019-07-04
volume command line argument supports increment format #2005

eauno1re:


main.c
```diff
@@ -145,8 +145,9 @@ print_help (void) {
     fprintf (stdout, _("                      FMT syntax: http://github.com/DeaDBeeF-Player/deadbeef/wiki/Title-formatting-2.0\n"));
     fprintf (stdout, _("                      example: --nowplaying-tf \"%%artist%% - %%title%%\" should print \"artist - title\"\n"));
     fprintf (stdout, _("   --volume [NUM]     Print or set deadbeef volume level.\n"));
-    fprintf (stdout, _("                      The NUM parameter can be specified in percents (if no suffix) or dB [-50, 0].\n"));
-    fprintf (stdout, _("                      Examples: --volume 80 or --volume -20dB\n"));
+    fprintf (stdout, _("                      The NUM parameter can be specified in percents (absolute value or increment/decrement)\n"));
+    fprintf (stdout, _("                      or in dB [-50, 0] (if with suffix).\n"));
+    fprintf (stdout, _("                      Examples: --volume 80, --volume +10, --volume -5 or --volume -20dB\n"));
```
Alexey-Yakovenko: The text is not accurate here. The code only handles percentage increments, but not dB.
I suggest cleaning up the code a little bit to handle both cases correctly.

eauno1re: The idea was add inc/dec option only for percentage. Percentage delta cat be either positive or negative, while it's impossible to distinguish negative dB delta and negative absolute dB value (e.g "-5dB" - does it mean we set volume to -5dB or we reduce volume by 5db?).

Alexey-Yakovenko: ah. thanks for the explanation. then it looks fine.

PR #2178, opened 2019-08-02
Automatic playlist sorting

kpcee:
This patch automatically sorts the current playlist when new files are added. It is then no longer necessary to manually sort the playlist because the playlist is messed up.

How does it work?
The feature must be enabled first using `gtkui.autosort_enabled 1`
After that you choose the way how the playlist should be sorted via Sort by -> Title, Track ...
The selection is stored in the config under the key `gtkui.last_used_sortby_format`.

Each time new files are added, the current playlist is sorted by the stored value. There is no need for manual sorting any more.

### DIFF
--- a/playlist.c
+++ b/playlist.c
@@ -3812,6 +3812,30 @@ plt_add_files_end (playlist_t *plt, int visibility) {
         l->callback_end (&d, l->user_data);
     }
     background_job_decrement ();
+    
+    // Autosort playlist
+    int autosort_enabled = deadbeef->conf_get_int ("gtkui.autosort_enabled", 0);
+    if (autosort_enabled == 1){
+
+        ddb_playlist_t *plt = deadbeef->plt_get_curr ();
+        const char *last_format = deadbeef->conf_get_str_fast ("gtkui.last_used_sortby_format", "None");
+        const char *formats = "%album% %artist% %title% %tracknumber% %year%";
+
+        if (strstr (formats, last_format) != NULL){
+            deadbeef->plt_sort_v2 (plt, PL_MAIN, -1, last_format, DDB_SORT_ASCENDING);
+        } else if (strcmp(last_format, "%random%") == 0){
+            deadbeef->plt_sort_v2 (plt, PL_MAIN, -1, NULL, DDB_SORT_RANDOM);
+        } else if (strcmp(last_format, "%custom%") == 0) {
+            int order = deadbeef->conf_get_int ("gtkui.sortby_order", 0);
+            const char *fmt = deadbeef->conf_get_str_fast ("gtkui.sortby_fmt_v2", "");
+            deadbeef->plt_sort_v2 (plt, PL_MAIN, -1, fmt, order == 0 ? DDB_SORT_ASCENDING : DDB_SORT_DESCENDING);
+        }
+
+        deadbeef->plt_save_config (plt);
+        deadbeef->plt_unref (plt);
+
+        deadbeef->sendmessage (DB_EV_PLAYLISTCHANGED, 0, DDB_PLAYLIST_CHANGE_CONTENT, 0);
+    }
 }
 
 void
--- a/playlist.c
+++ b/playlist.c
@@ -3813,29 +3813,40 @@ plt_add_files_end (playlist_t *plt, int visibility) {
     }
     background_job_decrement ();
     
-    // Autosort playlist
-    int autosort_enabled = deadbeef->conf_get_int ("gtkui.autosort_enabled", 0);
-    if (autosort_enabled == 1){
+    plt_autosort (plt);
+}
 
-        ddb_playlist_t *plt = deadbeef->plt_get_curr ();
-        const char *last_format = deadbeef->conf_get_str_fast ("gtkui.last_used_sortby_format", "None");
-        const char *formats = "%album% %artist% %title% %tracknumber% %year%";
-
-        if (strstr (formats, last_format) != NULL){
-            deadbeef->plt_sort_v2 (plt, PL_MAIN, -1, last_format, DDB_SORT_ASCENDING);
-        } else if (strcmp(last_format, "%random%") == 0){
-            deadbeef->plt_sort_v2 (plt, PL_MAIN, -1, NULL, DDB_SORT_RANDOM);
-        } else if (strcmp(last_format, "%custom%") == 0) {
-            int order = deadbeef->conf_get_int ("gtkui.sortby_order", 0);
-            const char *fmt = deadbeef->conf_get_str_fast ("gtkui.sortby_fmt_v2", "");
-            deadbeef->plt_sort_v2 (plt, PL_MAIN, -1, fmt, order == 0 ? DDB_SORT_ASCENDING : DDB_SORT_DESCENDING);
-        }
-
-        deadbeef->plt_save_config (plt);
-        deadbeef->plt_unref (plt);
-
-        deadbeef->sendmessage (DB_EV_PLAYLISTCHANGED, 0, DDB_PLAYLIST_CHANGE_CONTENT, 0);
+void
+plt_autosort (playlist_t *plt) {
+    int autosort_enabled = deadbeef->plt_find_meta_int (plt, "autosort_enabled", 0);
+    if (autosort_enabled != 1) {
+        return;
     }
+
+    const char *last_format = deadbeef->plt_find_meta (plt, "last_used_sortby_format");
+    if (last_format == NULL) {
+        return;
+    }
+    const char *formats = "%album% %artist% %title% %tracknumber% %year%";
+    if (strstr (formats, last_format) != NULL) {
+        deadbeef->plt_sort_v2 (plt, PL_MAIN, -1, last_format, DDB_SORT_ASCENDING);
+    }
+    else if (strcmp (last_format, "%random%") == 0) {
+         deadbeef->plt_sort_v2 (plt, PL_MAIN, -1, NULL, DDB_SORT_RANDOM);
+    }
+    else if (strcmp (last_format, "%custom%") == 0) {
+        int order = deadbeef->plt_find_meta_int (plt, "sortby_order", 0);
+        const char *fmt = deadbeef->plt_find_meta (plt, "sortby_fmt_v2");
+        if (fmt == NULL) {
+            return;
+        }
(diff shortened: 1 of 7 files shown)


Alexey-Yakovenko:
Some additional changes are needed before this can be accepted:

* UI checkbox needs to be added to the "Sort playlist" popup menu
* The autosorting enabled flag needs to be set per playlist
* The sort mode needs to be set per playlist
* Use `plt_get/set_meta_*` to set the per-playlist settings, instead of `conf_get/set_*`

playlist.c
```diff
@@ -3812,6 +3812,30 @@ plt_add_files_end (playlist_t *plt, int visibility) {
         l->callback_end (&d, l->user_data);
     }
     background_job_decrement ();
+    
+    // Autosort playlist
+    int autosort_enabled = deadbeef->conf_get_int ("gtkui.autosort_enabled", 0);
+    if (autosort_enabled == 1){
+
+        ddb_playlist_t *plt = deadbeef->plt_get_curr ();
```
Alexey-Yakovenko: It would make a lot of sense to move this code into a separate function

Alexey-Yakovenko: Also, you already have the playlist pointer passed as an argument, you should use it instead of getting a current playlist.

kpcee:
Thank you for the detailed review. 
I've updated the PR, please check it out.

playlist.c
```diff
@@ -3812,6 +3812,41 @@ plt_add_files_end (playlist_t *plt, int visibility) {
         l->callback_end (&d, l->user_data);
     }
     background_job_decrement ();
+    
+    plt_autosort (plt);
+}
+
+void
+plt_autosort (playlist_t *plt) {
```
Alexey-Yakovenko: You should declare this function static, and add a forward declaration - otherwise it fails a build on mac.

playlist.c
```diff
@@ -3812,6 +3812,41 @@ plt_add_files_end (playlist_t *plt, int visibility) {
...
+    int autosort_enabled = deadbeef->plt_find_meta_int (plt, "autosort_enabled", 0);
```
Alexey-Yakovenko: The `plt` variable has type of `playlist_t *`, but `deadbeef->plt_find_meta_int` expects argument of type `ddb_playlist_t *`.
It's preferred to call `plt_find_meta_int` directly, instead of via `deadbeef->`, then you would not need to type-cast.

Alexey-Yakovenko: (same thing for the other similar calls)

playlist.c
```diff
@@ -3812,6 +3812,41 @@ plt_add_files_end (playlist_t *plt, int visibility) {
...
+    if (autosort_enabled != 1) {
+        return;
+    }
+
+    const char *last_format = deadbeef->plt_find_meta (plt, "last_used_sortby_format");
+    if (last_format == NULL) {
+        return;
+    }
+    const char *formats = "%album% %artist% %title% %tracknumber% %year%";
+    if (strstr (formats, last_format) != NULL) {
```
Alexey-Yakovenko: This is an unacceptable hack, which will more than likely cause problems in the future.
If you want to check if the last_format is one of these values -- make an array of strings, and compare in a loop.

playlist.c
```diff
@@ -3812,6 +3812,41 @@ plt_add_files_end (playlist_t *plt, int visibility) {
...
+        deadbeef->plt_sort_v2 (plt, PL_MAIN, -1, last_format, DDB_SORT_ASCENDING);
+    }
+    else if (strcmp (last_format, "%random%") == 0) {
+         deadbeef->plt_sort_v2 (plt, PL_MAIN, -1, NULL, DDB_SORT_RANDOM);
+    }
+    else if (strcmp (last_format, "%custom%") == 0) {
```
Alexey-Yakovenko: `%random%` and `%custom%` is not valid title formatting, and it doesn't make any sense to use it here.
You should reimagine how you want to store this configuration.
For example: `autosort_mode=random|custom`, and `autosort_custom=whatever-title-formatting`.
Then you wouldn't need to do the `strstr` hack either.

kpcee:
Okay, next try.

plugins/gtkui/callbacks.c
```diff
@@ -667,6 +667,7 @@ on_sort_by_title_activate              (GtkMenuItem     *menuitem,
     deadbeef->plt_unref (plt);
 
     deadbeef->sendmessage (DB_EV_PLAYLISTCHANGED, 0, DDB_PLAYLIST_CHANGE_CONTENT, 0);
+    deadbeef->plt_set_meta_int (plt, "autosort_mode", 0);
```
Alexey-Yakovenko: Ok.. I dislike this even more than the previous version.

I would just have 2 variables:

1. `autosort_mode` which is either `random`, `tf` (title formatting), or anything else (which means "off").
2. `autosort_tf` set to the title formatting script, if not random.

Then all you have to do is check whether autosort_mode is random, or tf, and do the corresponding logic.
You don't need the hardcoded tf strings at all.

You don't need to use `sortby_fmt_v2` name -- in the context of autosort it's not "v2". It is "v2" in GTKUI because previously there was sortby_fmt "v1" which was before proper title formatting was implemented. You don't need to handle backwards compatibility, so just call it "autosort_tf".

It would also make sense to use "autosort_order" to store the ascending/descending flag.
I would even call it "autosort_descending" 1 or 0, which would be the most readable outcome.

Alexey-Yakovenko: BTW, mac build is still failing. You can see the errors by clicking "Details" link next to the failure.

kpcee:
Is everything okay now?

Alexey-Yakovenko:
@kpcee I didn't have time to look at the latest iteration yet, but I will do as soon as I can.
Right now I can only say that the builds are passing.

playlist.c
```diff
@@ -3813,31 +3813,27 @@ plt_add_files_end (playlist_t *plt, int visibility) {
     }
     background_job_decrement ();
     
-    plt_autosort (plt);
+    plt_autosort (plt, d.plt);
 }
 
 static void
-plt_autosort (playlist_t *plt) {
+plt_autosort (playlist_t *plt, ddb_playlist_t *ddbplt) {
```
Alexey-Yakovenko: You don't need to pass the same thing cast to another pointer type.
It's the same pointer.

playlist.c
```diff
@@ -3813,31 +3813,27 @@ plt_add_files_end (playlist_t *plt, int visibility) {
...
-        const char *fmt = plt_find_meta (plt, "sortby_fmt_v2");
+    const char *autosort_mode = plt_find_meta (plt, "autosort_mode");
+    if (strcmp (autosort_mode, "tf") == 0) {
+        int order = plt_find_meta_int (plt, "autosort_ascending", 0);
+        const char *fmt = plt_find_meta (plt, "autosort_tf");
         if (fmt == NULL) {
             return;
         }
-        plt_sort_v2 (plt, PL_MAIN, -1, fmt, order == 0 ? DDB_SORT_ASCENDING : DDB_SORT_DESCENDING);
+        deadbeef->plt_sort_v2 (ddbplt, PL_MAIN, -1, fmt, order == 0 ? DDB_SORT_ASCENDING : DDB_SORT_DESCENDING);
```
Alexey-Yakovenko: You can simply call `plt_sort_v2` without `deadbeef`, and use `plt` as argument.
Same below.

plugins/gtkui/actionhandlers.c
```diff
@@ -678,9 +678,9 @@ action_sort_custom_handler_cb (void *data) {
         deadbeef->plt_unref (plt);
 
         deadbeef->sendmessage (DB_EV_PLAYLISTCHANGED, 0, DDB_PLAYLIST_CHANGE_CONTENT, 0);
-        deadbeef->plt_set_meta_int (plt, "autosort_mode", 6);
-        deadbeef->plt_replace_meta (plt, "sortby_fmt_v2", fmt);
-        deadbeef->plt_set_meta_int (plt, "sortby_order", order);
+        deadbeef->plt_replace_meta (plt, "autosort_mode", "tf");
```
Alexey-Yakovenko: You need to use `plt_replace_meta ` without `deadbeef->` and so on.

kpcee:
All tests have been successfully completed, I hope this is okay now.

plugins/gtkui/actionhandlers.c
Alexey-Yakovenko: (This comment above was my mistake, I didn't notice it was part of GTKUI)

kpcee:
I have now restored `deadbeef->` everywhere to files located in the plugin folder.
But we still have one problem with the function `static void plt_autosort (playlist_t *plt)` in the playlist.c file.

`deadbeef->plt_sort_v2 (plt,` ... can't be removed, otherwise GitHub will return a compile error, but we have a wrong pointer as long as we use `deadbeef->`.
How should this be fixed? A cast of the pointer in the function or pass the casted pointer over the function or anything else?

Alexey-Yakovenko:
> How should this be fixed?

The best solution to this is to move `plt_autosort` function to `sort.c`.
Remember to remove `static` from it, and move the declaration from `playlist.h` to `sort.h`

kpcee:
Well, that doesn't work either. It fixes the problem with plt_sort_v2, but now plt_find_meta and plt_find_meta_int return error messages and for deadbeef->sendmessage() you have to add an extra #include, without deadbeef-> doesn't work.
Since the source code is okay so far, it might be possible that you insert the relevant parts where you think it is correct. This would be faster than writing back and forth for days.

Alexey-Yakovenko:
You just need to include the relevant headers. In your case it should be "plmeta.h" and "messagepump.h", You will also need to call `messagepump_push` instead of `sendmessage`.
Obviously, I could just implement the feature myself, but I don't have time.

kpcee:
Done but the The Travis CI build fails with "error: unexpected service error: The Xcode build system has crashed. Please close and reopen your workspace."

Alexey-Yakovenko:
@kpcee it's probably an out of memory or something similar. I reran the build.

kpcee:
Okay, now it works without errors.

Alexey-Yakovenko:
The code looks ok, next step - I need to test it. I will contact you as soon as I have any news.

Alexey-Yakovenko:
I applied the changes that I thought were necessary, and did some basic testing. Merged manually.
Thanks for your work.

PR #2190, opened 2019-08-20
 Implement remaining string title formatting funcs

esainane:
This implements the previously unimplemented functions listed in http://wiki.hydrogenaud.io/index.php?title=Foobar2000:Title_Formatting_Reference#String_functions

Caveats listed in commit ace3f7f. The main points are:
 - Conforming to DeaDBeeF's existing error semantics over fb2k's error semantics
 - Conforming to DeaDBeef's existing boolean semantics over fb2k's boolean semantics
 - Implementing $len2 as documented, rather than matching the behavior I observe from foobar2000 under wine (where it appears to behave identically to $len). If foobar2000's $len2 does behave in this manner under a native environment, I can match and/or document this behavior.
 - $strstr possibly finding false positives in current implementation due to using byte rather than character boundaries. The end result is always in bytes.
 - $tab does add tabs, but these render as underscores in my (GTK2) interface.
 - I am unable/unsure how to run the automated tests in `osx/Tests/TitleFormattingTests.m`. I have done my best to test all functions manually, but for added confidence it might be useful to run these tests.

Four sanity commits are provided before the functional changes in the last two commits. The four commits provide nonfunctional changes, bounds checking, buffer length corrections, and documentation for existing code as I understand it, which might help identify any invalid assumptions I've made early. If it would help, I can split these four out into a separate pull request.


esainane:
`wcwidth.c` and `wcwidth.h` need to be added to the build for the osx version. `osx/deadbeef.xcodeproj/project.pbxproj` appears to be a generated file, but I'm afraid I'm unable/unsure of how to update it.

Alexey-Yakovenko:
> osx/deadbeef.xcodeproj/project.pbxproj appears to be a generated file, but I'm afraid I'm unable/unsure of how to update it.

don't worry about this. I'll update the xcode project.

esainane:
> I disagree with all of the added "-1"s. Adding them presumes that the underlying code doesn't work correctly. I don't believe this is the case, so this needs to be undone. Unless of course you can provide a test which would demonstrate an actual problem in the code.

It's not that the called code is not correct, if that's what you mean by underlying code. As far as I'm aware, `strncat` works exactly as documented. However, this called code is called in a way that does not match the documentation. See `man 3 strncat`, where it declares that `it will use at most n bytes from src`, and `If  src  contains  n  or  more  bytes,  strncat()  writes  n+1  bytes  to  dest  (n  from  src plus the terminating null byte).  Therefore, the size of dest must be at least strlen(dest)+n+1.` With a buffer that presently contains the empty string, this means that it have size at least n + 1, which is not the case when n is `sizeof(buf)`.

This is true for every function that does not expect the terminating null character to be part of the buffer length in its arguments, such as all of the functions of type `tf_func_ptr_t` with their `out` and `outlen` parameters, or `tf_eval_int` (and so indirectly, `TF_EVAL_CHECK`), as currently documented at https://github.com/DeaDBeeF-Player/deadbeef/blob/master/tf.c#L176

For a simple test case that demostrates actual problems in the code:

```bash
$ cat test.c
```
```c
#include <stdlib.h>
#include <stdio.h>

#include <string.h>

int main(void) {
    const char very_long_source[] = "According to all known laws of aviation,  there is no way a bee should be able to fly.  Its wings are too small to get its fat little body off the ground.  The bee, of course, flies anyway  because bees don't care what humans think is impossible.";
    char buf[10];
    *buf = 0;

    strncat(buf, very_long_source, sizeof(buf));
    return 0;
}
```
```bash
$ gcc test.c
$ valgrind ./a.out
==1583== Memcheck, a memory error detector
==1583== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==1583== Using Valgrind-3.14.0 and LibVEX; rerun with -h for copyright info
==1583== Command: ./a.out
==1583== 
==1583== Source and destination overlap in strncat(0x1ffefffc56, 0x1ffefffc60, 10)
==1583==    at 0x4838BEC: strncat (vg_replace_strmem.c:350)
==1583==    by 0x1091AD: main (in /home/user/a.out)
==1583== 
==1583== 
==1583== HEAP SUMMARY:
==1583==     in use at exit: 0 bytes in 0 blocks
==1583==   total heap usage: 0 allocs, 0 frees, 0 bytes allocated
==1583== 
==1583== All heap blocks were freed -- no leaks are possible
==1583== 
==1583== For counts of detected and suppressed errors, rerun with: -v
==1583== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)
$
```
`strncat` behaves as documented, copying sizeof(buf) characters from the source. It then adds a null terminator byte, which is a string overflow, and was dutifully detected by valgrind.

Correcting the parameter passed into `strncat` to `sizeof(buf) - 1`, leaving room for the null terminator it will add, fixes this issue.

This is also helpfully detected in recent versions of clang, for a few specific library string functions. For the above test case, it will provide both the explanation and suggest a fix:

```bash
$ clang -Wall test.c
test.c:11:36: warning: the value of the size argument in 'strncat' is too large, might lead to a buffer overflow [-Wstrncat-size]
    strncat(buf, very_long_source, sizeof(buf));
                                   ^~~~~~~~~~~
test.c:11:36: note: change the argument to be the free space in the destination buffer minus the terminating null byte
    strncat(buf, very_long_source, sizeof(buf));
                                   ^~~~~~~~~~~
                                   sizeof(buf) - strlen(buf) - 1
1 warning generated.
```
...though when we know `strlen(buf)` is 0, as in the above test case, this simplifies to `sizeof(buf) - 1`.

It was actually these warnings I received while building DeaDBeeF which led me to inspect usage of string functions in the areas I was looking with a little more scrutiny than usual. For example:

```
plcommon.c: In function ‘on_group_by_custom_activate’:
plcommon.c:1556:13: warning: ‘strncat’ specified bound 1024 equals destination size [-Wstringop-overflow=]
             strncat(format, SUBGROUP_DELIMITER, sizeof(format));
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
plcommon.c:1558:9: warning: ‘strncat’ specified bound 1024 equals destination size [-Wstringop-overflow=]
         strncat(format, fmt->format, sizeof(format));
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

libsc68/conf68.c: In function ‘load_from_registry’:
libsc68/conf68.c:258:7: warning: ‘strncat’ specified bound 128 equals destination size [-Wstringop-overflow=]
       strncat(path, opt->name, sizeof(path));
       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
libsc68/conf68.c: In function ‘config68_save’:
libsc68/conf68.c:198:5: warning: ‘strncat’ specified bound 128 equals destination size [-Wstringop-overflow=]
     strncat(tmp, confname, sizeof(tmp));
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
There are some warnings for src/wildmidi_lib.c on lines 637 and 620 as well, but only that the bounds depends on the length of the source argument, which after inspection I'm fairly sure is just a false positive.

I should also give credit to clang for raising an alert about the extra iteration in `nes_apu.c`:
```
game-music-emu-0.6pre/vgmplay/VGMPlay/chips/nes_apu.c: In function ‘device_reset_nesapu’:
game-music-emu-0.6pre/vgmplay/VGMPlay/chips/nes_apu.c:724:25: warning: iteration 23 invokes undefined behavior [-Waggressive-loop-optimizations]
  info->APU.regs[address]=value;
  ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~
game-music-emu-0.6pre/vgmplay/VGMPlay/chips/nes_apu.c:884:2: note: within this loop
  for (CurReg = 0x00; CurReg < 0x18; CurReg ++)
  ^~~
```
Aren't modern compilers amazing? :)

esainane:
I'll force push amended commits with the preferred comment style shortly.

Alexey-Yakovenko:
Ok, regarding the "-1" I missed a few. The ones about strncat are probably valid, although I would need to take a closer look. I'm surprised no static analyzers spotted that in such long time.
However adding -1 to TF_EVAL_CHECK invocations looks very weird to me.
Looks like "treating symptoms not cause"

esainane:
I'm not really sure what I can suggest? `tf_eval_int`, and so its wrapper `TF_EVAL_CHECK`, do seem to behave as documented in [`// tf_eval_int expects outlen to not include the terminating zero`](https://github.com/DeaDBeeF-Player/deadbeef/blob/master/tf.c#L176), and all of the functions in the function table behave the same way?

If you really want, I can look into making sure every function in tf.c uses `outlen` like the `n` in `snprintf` rather than the `n` in `strncat`. That's not going to be a small change, and we'd still then need to -1 when invoking `u8_strnbcpy` or the like, unless those should be changed too?

(I just wanted to port over the `$substr($directory(%path%,1),1,$sub($strstr($directory(%path%,1),' ['),1))` column I had from foobar2000 ^^; )

Alexey-Yakovenko:
> I'm not really sure what I can suggest?

Nothing.. Let's keep this for later. I will have to have another look at how all this stuff work, and see why it's written the way it is.

esainane:
Okay. Thank you for your time, and for making such a great audio player. :)

Alexey-Yakovenko:
> tf_eval_int, and so its wrapper TF_EVAL_CHECK, do seem to behave as documented in // tf_eval_int expects outlen to not include the terminating zero, and all of the functions in the function table behave the same way?

I verified where and why you added -1 to the calls of TF_EVAL_CHECK, and your fix is correct.
I'm still puzzled though, as to why static analyzer didn't catch this. Probably some misconfiguration.

Alexey-Yakovenko:
I created a new branch `tf` for this work, with macos build fixed. If you want to continue any work on this, please make PRs into tf branch instead. I will merge it to master when everything is ready.

Alexey-Yakovenko:
@esainane FYI: your changes cause a few titleformatting tests to fail
```
	-[TitleFormatting test_FixEofTwoArgsWithSmallBuffer_DoesntOverflowOffByOne]
	-[TitleFormatting test_HexPadded_ReturnsHexConvertedNumberWithPadding]
```

esainane:
Taking a look, it appears both test cases are expecting there to be an empty string when tf_eval is given a small outlen.

The culprit is likely the last commit, https://github.com/DeaDBeeF-Player/deadbeef/pull/2190/commits/d0db6de1ad18c7d76d83d53a0d83f3d07b82f69e - intended to ensure recursive calls always have enough working memory. This changes behavior to allocate a larger temporary buffer if outlen is small, and copies as much as possible from this temporary buffer into the destination buffer when done.

Do the tests still fail if you go to https://github.com/DeaDBeeF-Player/deadbeef/blob/tf/tf.c#L2571 and replace `if (res > 0) {` with

```c
                    if (res > outlen) {
                        res = -1;
                    } else if (res > 0) {
```
?

That could be one way to restore the old behavior if desired, though it might break other functions. I notice $fix_eol and $hex are both explicit about returning -1 when outlen is insufficient, while other functions seem to go as far as they can? Is there consistent behavior you'd want one way or the other?

esainane:
(For example, during testing, $directory once tried to process a path string that was written as much as available - "/media/Ic" - and dutifully returned "media" which, while helpful in working out what was going on, wasn't the intended result)

Alexey-Yakovenko:
> The culprit is likely the last commit, d0db6de

Yes, it was this commit. I reverted the change.
It hard to tell what will fail or not without unit tests, so when the tests are added it will be more clear.

Alexey-Yakovenko:
`$progress(0,0,10,x,=)` returns `0`, I think that this is not the expected output.

Alexey-Yakovenko:
Other failing tests
`$roman(100500)` -> `100500`, expected `(C)D`
`$roman(51880)` -> `MMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMDCCCLXXX`, expected `(L)MDCCCLXXX`

Alexey-Yakovenko:
`$strchr(DeaDBeeF,R)` returns `0` instead of ""
I'm not sure if this is expected.

esainane:
> `$progress(0,0,10,x,=)` returns `0`, I think that this is not the expected output.

You are correct:

progress-0-0

I'll allow and add a case to handle the 0/0 parameters shortly.

---

> Other failing tests
> `$roman(100500)` -> `100500`, expected `(C)D`
> `$roman(51880)` -> `MMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMDCCCLXXX`, expected `(L)MDCCCLXXX`



In foobar2000, `$roman(100500)` produces the empty string:
roman-100500
In foobar2000, `$roman(51880)` produces `MMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMDCCCLXXX`, as is implemented currently:
roman-51880

foobar2000 appears not to allow any input above 100000, and is limited to using `M` as the highest component. Changing this behavior would be a compatibility break. If you confirm, I can implement higher limits and components if desired?

Either way, passing through the input unaltered once the limit is reached is definitely a bug. It's an unusual one too, as `$roman` returns -1 when the input is out of range, and does return the empty string when a nested call rather than the top level call. Perhaps `tf_eval` should be calling the `TF_EVAL_CHECK` wrapper rather than `tf_eval_int` directly?

---

> `$strchr(DeaDBeeF,R)` returns `0` instead of ""
> I'm not sure if this is expected.

This is the foobar2000 behavior:

strchr-0

Alexey-Yakovenko:
Limiting the input to 100000 is fine. Thanks. I'll update the tests.

PR #2194, opened 2019-08-23
Title formatting, Part 2

esainane:
This follows on from #2190.

- When the top level function fails, produce an empty string, like it would if it was a nested call.
- Adjust `$progress` to support and match fb2k's handling of zero-range calls, such as `$progress(0,0,10,x,=)`
- Adds a test to track support of complicated function chains.

### DIFF
--- a/osx/Tests/TitleFormattingTests.m
+++ b/osx/Tests/TitleFormattingTests.m
@@ -2218,5 +2218,13 @@ static DB_output_t fake_out = {
     XCTAssert(!strcmp (buffer, ""), @"The actual output is: %s", buffer);
 }
 
+- (void)test_Complex_Sane {
+    pl_replace_meta (it, "path", "/media/Icy/Music/Long/Folder/Structure/2019.01.01 [Meta1] Meta2 [Meta3] Meta4 [Meta5] Meta6 [Meta7] Meta8 [Meta9]/some_reasonably_long_path.flac");
+    char *bc = tf_compile("$substr($directory(%path%,1),1,$sub($strstr($directory(%path%,1),' ['),1))");
+    tf_eval (&ctx, bc, buffer, 1000);
+    tf_free(bc);
+    XCTAssert(!strcmp (buffer, "2019.01.01"), @"The actual output is: %s", buffer);
+}
+
 
 @end
--- a/tf.c
+++ b/tf.c
@@ -197,7 +197,7 @@ tf_eval (ddb_tf_context_t *ctx, const char *code, char *out, int outlen) {
         break;
     default:
         // tf_eval_int expects outlen to not include the terminating zero
-        l = tf_eval_int (ctx, code, codelen, out, outlen - 1, &bool_out, 0);
+        TF_EVAL_CHECK(l, ctx, code, codelen, out, outlen - 1, 0);
         break;
     }
 
--- a/tf.c
+++ b/tf.c
@@ -1181,8 +1181,11 @@ tf_func_progress_impl(ddb_tf_context_t *ctx, int argc, const uint16_t *arglens, 
 
     TF_EVAL_CHECK(len, ctx, argpos, arglens[1], out, outlen - 1, fail_on_undef);
     range = atoi(out);
-    if (range <= 0) {
+    if (range < 0) {
         return -1;
+    }
+    if (range == 0) {
+        range = 1;
     }
     argpos += arglens[1];
 


esainane:
I am still unable/unsure of how to run the automated tests from a linux environment, but I have tried to match the format of the other tests.

osx/Tests/TitleFormattingTests.m
```diff
@@ -2218,5 +2218,13 @@ - (void)test_StrrchrDeaDBeeF_R_ReturnsEmpty {
     XCTAssert(!strcmp (buffer, ""), @"The actual output is: %s", buffer);
 }
 
+- (void)test_Complex_Sane {
```
Alexey-Yakovenko: Test naming convention: test_INPUTCONDITION_EXPECTEDOUTPUT
This test is also bad, because it's testing multiple things at once, so if it starts failing it would be hard to tell why and debug.
Anyway.. you don't have to make tests, especially if you don't have access to XCode to run them.
I'll cover that.

Alexey-Yakovenko:
> I am still unable/unsure of how to run the automated tests from a linux environment, but I have tried to match the format of the other tests.

Unfortunately this only works on mac/xcode. I chose convenience for myself over cross-platform support.

osx/Tests/TitleFormattingTests.m
esainane: Would the following test be more acceptable?
```c
- (void)test_NestingDirectoryInSubstr_HasNoIntermediateTruncation {
    pl_replace_meta (it, "path", "/Wrong1/Wrong2/CorrectSlice/track.flac");
    char *bc = tf_compile("$substr(%path%, $strstr(%path%, $directory(%path%,1)), 10000)");
    tf_eval (&ctx, bc, buffer, 1000);
    tf_free (bc);
    XCTAssert(!strcmp (buffer, "CorrectSlice/track.flac"), @"The actual output is: %s", buffer);
} 
```
Assuming I've implemented the test correctly, it would also fail, producing `"/Wrong1/Wrong2/CorrectSlice/track.flac"` as the intermediate buffer can't hold the full length of `%path%`.

Alexey-Yakovenko: If you expect this test to fail, you should invert the assert condition.
If the intermediate/internal buffer can't hold the value - well, we should evaluate whether the intermediate buffer size should be increased, or it should limit the length and return nothing.

esainane: The test fails with the current HEAD, but it's my belief that it shouldn't. This was the sort of scenario that https://github.com/DeaDBeeF-Player/deadbeef/commit/d0db6de1ad18c7d76d83d53a0d83f3d07b82f69e prevented which was reverted in https://github.com/DeaDBeeF-Player/deadbeef/commit/3a3ed30d42e7841ac130b92c5d881fafd213a81e in order to allow tests to continue to check behavior with artificially small buffers. This test was added to make sure the scenario isn't forgotten about.

esainane: To elaborate, the problem is that whenever a very small value is expected - such as the 1 in `$directory(%path%,1)` or the 0 or 5 in `$substr(%path%,0,5)` - only a small buffer is allocated to retrieve the value, ten to twenty bytes long. This is fine and efficient for constant values, but isn't anywhere near enough if, rather than a constant value, a function call is invoked instead. 

By automatically increasing the buffer size, we stopped these failure modes, but we also stopped failure modes that `test_FixEofTwoArgsWithSmallBuffer_DoesntOverflowOffByOne` and `test_HexPadded_ReturnsHexConvertedNumberWithPadding` were expecting.

I'm not sure what to suggest. Smart buffer growth could be restored, and another parameter added to optionally inhibit it so that the tests can continue to test for behavior under artificially small buffer scenarios, or all temporary buffers could be increased to 1kB rather than the 10-20 bytes they have now?

Leaving nested function calls with only 10-20 byes of working memory would mean quite a lot of custom tf scripts from fb2k would fail in unexpected and hard to diagnose ways.

Alexey-Yakovenko: > in order to allow tests to continue to check behavior with artificially small buffers

Well this IS the intended behavior, that the underlying code would not try to be smart.

esainane: So what would be the best way to preserve this behavior, while not having intermediate results be truncated?

Alexey-Yakovenko: In what case do you need to have this behavior? Any realistic case which may cause problems with the current implementation?

esainane: In the specific case, with foobar2000 I had a custom column using the script `$substr($directory(%path%,1),1,$sub($strstr($directory(%path%,1),' ['),1))`. My library is arranged so that each folder containing audio files has the format `date [id] name [release]`, and I make columns with each component extracted, as this is more reliable and consistent than the information from cuesheets.

With foobar2000 this does work as expected. With the current DeaDBeeF implementation, because only a small buffer is allocated in `$substr` for the indices, `%path%` gets truncated to `/media/Ic`, `$directory` returns `media`, `$strstr` doesn't have the right directory name to extract information from, and so the end result is the empty string.

In the general case, I'd expect the results of any call to `$strchr` or `$strstr` to be used as an index to `$substr` or postprocessed to select the slice in `$directory`, and only being able to search within a 10-20 byte buffer would greatly diminish their usefulness.

Alexey-Yakovenko: @esainane Thanks for the explanation, and sorry for delays with replying.
I will see what can be done with this as soon as I have time.

esainane: Not a problem at all! Please do let me know if there is anything I can do to help. :)

Alexey-Yakovenko: > The test fails with the current HEAD, but it's my belief that it shouldn't. This was the sort of scenario that [d0db6de](https://github.com/DeaDBeeF-Player/deadbeef/commit/d0db6de1ad18c7d76d83d53a0d83f3d07b82f69e) prevented which was reverted in [3a3ed30](https://github.com/DeaDBeeF-Player/deadbeef/commit/3a3ed30d42e7841ac130b92c5d881fafd213a81e) in order to allow tests to continue to check behavior with artificially small buffers. This test was added to make sure the scenario isn't forgotten about.

Reverting the 3a3ed30 doesn't help the test to pass. Breakpoint is not hit on the condition that the buffer is too small.

Alexey-Yakovenko: The test had a mistake, I corrected it, and now I can reproduce the issue.
Trying to figure out what's going on.

Alexey-Yakovenko: So the reason was not caused by the general output buffer management.
It was incorrect temp buffer size in the strstr function implementation.
I corrected it, and the test is passing now.

Alexey-Yakovenko: FYI: `$trim` treats `\t` chars as non-spaces.
I'm assuming that this is intended, and adding a test for that.
Please let me know if this is wrong.

Alexey-Yakovenko:
I'm about to merge this branch to master.
Please use this sheet if you want to continue tf work.
And thanks for the great job with this PR!

PR #2210, opened 2019-09-10
Start minimized to tray

kpcee:
This is a pull request for a better start minimized to tray.
The following things have been improved:
* the main window is no longer displayed and then hidden
* no more high CPU load if the main window was not switched to visible before (caused by init_widget_layout ();)
* no endless error messages GDK_IS_WINDOW (window)' failed errors (caused by update_songinfo (gpointer unused))
* shutdown of the application has also been reviewed and modified
* start minimized to tray can now be configured in the settings
* a check has been added that "minimized to tray" and "trayicon invisible" cannot be enabled both

pic

### DIFF
--- a/plugins/converter/callbacks.h
+++ b/plugins/converter/callbacks.h
@@ -134,3 +134,7 @@ on_bypass_same_format_toggled          (GtkToggleButton *togglebutton,
 void
 on_retag_after_copy_toggled            (GtkToggleButton *togglebutton,
                                         gpointer         user_data);
+
+void
+on_minimize_on_startup_clicked         (GtkButton       *button,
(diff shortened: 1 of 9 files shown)


Alexey-Yakovenko:
The code looks ok and safe. I think it will be fine to merge after I do some simple testing.
The main concern is titled WMs, but from what I see in the code, the logic should not be affected much.

kpcee:
Sorry, I hadn't considered that scenario.

Alexey-Yakovenko:
> Sorry, I hadn't considered that scenario.

Thanks, I'll try to test again today, and get back to you.

PR #2219, opened 2019-09-18
Add playback_time_ms title format keyword

nonoo:
The playback_time_ms keyword can be used with the ddb_playback_status plugin to display playback position in a higher resolution.

### DIFF
--- a/tf.c
+++ b/tf.c
@@ -2024,7 +2024,7 @@ tf_eval_int (ddb_tf_context_t *ctx, const char *code, int size, char *out, int o
                 int skip_out = 0;
 
                 // temp vars used for strcmp optimizations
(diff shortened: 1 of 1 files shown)


Alexey-Yakovenko:
It looks quite meaningless to me, since update interval for %playback_position% is 1 sec. How are you planning to use this?

nonoo:
I use this with the Playback Status Widget (https://github.com/cboxdoerfer/ddb_playback_status) and it supports higher refresh rates.

Alexey-Yakovenko:
@nonoo I have added 100ms refresh rate for that case, + added a test, and implemented timer support in cocoaui, so that it's usable on Mac as well.

Issue #2197, opened 2019-08-24
%_path_raw% field provides path in non-expected format


kravich:
### Steps to reproduce the problem
Add custom column to playlist with %_path_raw% in "Format" filed

### What's going on? Describe the problem in as much detail as possible.
Foobar2000's title formatting reference states that %_path_raw% field [should produce URL](http://wiki.hydrogenaud.io/index.php?title=Foobar2000:Title_Formatting_Reference#.25_path_raw.25). However, the actual result differs - for network sources produced string is excatly a URL, but for local files it is actually an absolute file path:
![screen](https://user-images.githubusercontent.com/774509/63637853-dc189900-c6c4-11e9-84ba-73659e5680d5.png)

This happens because titleformatting engine uses ":URI" metadata record [as is](https://github.com/DeaDBeeF-Player/deadbeef/blob/master/tf.c#L2346) to produce %_path_raw% output.

Is it a bug or an expected result because of historical reasons?
If it is a bug, I am going to provide a fix for it.

### Information about the software:
Deadbeef version: 1.8.2
OS: Ubuntu Linux 19.10

Alexey-Yakovenko:
Have you verified whether foobar 2000 actually produces URLs for file paths?

Alexey-Yakovenko:
Nvm, I checked, it does, it's a bug, please feel free to submit a fix

kravich:
Do I understand correctly that fix should affect only titleformatting API and should not change the semantic of ":URI" metadata record?

Alexey-Yakovenko:
Yes, only title formatting

PR #2222, opened 2019-09-23
Path raw fix

kravich:
This is a fix for #2197.

It's pretty trivial, however, there is some difficulty with relative paths that also may be stored in :URI metadata record. I decided to format them as non-standard file URI extension, because the alternative is to call realpath() to convert relative paths to absolute, and it is bad idea to do so in tf_eval() because it should be fast.

Maybe, it is a good idea not to store relative paths inside :URI at all by explicitly resolving them inside plt_insert_file_int(). It seems that foobar2000 is doing similar thing because for relative path provided through File->Add location it outputs absolute %_path_raw%. However, this is more intrusive change and it should be made per separate PR, if at all.

Please let me know what do you think on it.

I also wanted to add new tests for this use case, but unfortunately I have no Mac to verify them.

Testing:
screen

### DIFF
--- a/osx/Tests/TitleFormattingTests.m
+++ b/osx/Tests/TitleFormattingTests.m
@@ -1341,6 +1341,30 @@ static DB_output_t fake_out = {
     XCTAssert(!strcmp (buffer, "file:///home/user/filename.mp3"), @"The actual output is: %s", buffer);
 }
 
+- (void)test_RawPathWithHttpUriScheme_ReturnUnStripped {
+    pl_replace_meta (it, ":URI", "http://example.com/filename.mp3");
+    char *bc = tf_compile("%_path_raw%");
+    tf_eval (&ctx, bc, buffer, 1000);
+    tf_free (bc);
+    XCTAssert(!strcmp (buffer, "http://example.com/filename.mp3"), @"The actual output is: %s", buffer);
+}
+
+- (void)test_RawPathAbsolutePath_ReturnsExpected {
+    pl_replace_meta (it, ":URI", "/path/to/filename.mp3");
+    char *bc = tf_compile("%_path_raw%");
+    tf_eval (&ctx, bc, buffer, 1000);
+    tf_free (bc);
+    XCTAssert(!strcmp (buffer, "file:///path/to/filename.mp3"), @"The actual output is: %s", buffer);
+}
+
+- (void)test_RawPathRelativePath_ReturnsEmpty {
+    pl_replace_meta (it, ":URI", "relative/path/to/filename.mp3");
+    char *bc = tf_compile("%_path_raw%");
+    tf_eval (&ctx, bc, buffer, 1000);
+    tf_free (bc);
+    XCTAssert(!strcmp (buffer, ""), @"The actual output is: %s", buffer);
+}
+
 - (void)test_PlaylistName_ReturnsPlaylistName {
     char *bc = tf_compile("%_playlist_name%");
     playlist_t plt = {
@@ -1461,6 +1485,14 @@ static DB_output_t fake_out = {
 - (void)test_PathWithNullUri_ReturnsEmpty {
     pl_delete_meta (it, ":URI");
     char *bc = tf_compile("%path%");
+    tf_eval (&ctx, bc, buffer, 1000);
+    tf_free (bc);
+    XCTAssert(!strcmp (buffer, ""), @"The actual output is: %s", buffer);
+}
+
+- (void)test_RawPathWithNullUri_ReturnsEmpty {
+    pl_delete_meta (it, ":URI");
+    char *bc = tf_compile("%_path_raw%");
     tf_eval (&ctx, bc, buffer, 1000);
     tf_free (bc);
     XCTAssert(!strcmp (buffer, ""), @"The actual output is: %s", buffer);
--- a/tf.c
+++ b/tf.c
@@ -2343,7 +2343,40 @@ tf_eval_int (ddb_tf_context_t *ctx, const char *code, int size, char *out, int o
                     }
                 }
                 else if (!strcmp (name, "_path_raw")) {
-                    val = pl_find_meta_raw (it, ":URI");
+                    const char *v = pl_find_meta_raw (it, ":URI");
+
+                    if (v) {
+                        if (v[0] == '/') {
+                            // This is an absolute path, just prepend proper prefix
+                            const char prefix[] = "file://";
+                            tf_append_out (&out, &outlen, prefix, sizeof (prefix) - 1);
+                            tf_append_out (&out, &outlen, v, strlen (v));
+                        }
+                        else {
+                            int is_uri = 1;
+
+                            for (const char *p = v; p; p++) {
+                                if (!strncmp (p, "://", 3)) {
+                                    break;
+                                }
+
+                                if (!isalpha (*p)) {
+                                    is_uri = 0;
+                                    break;
+                                }
+                            }
+
+                            if (is_uri) {
+                                // This is already a URI, just copy as is
+                                tf_append_out (&out, &outlen, v, strlen (v));
+                            }
+                            else {
+                                // Relative paths are considered invalid
+                            }
+                        }
+
+                        skip_out = 1;
+                    }
                 }
                 else if (!strcmp (name, "path")) {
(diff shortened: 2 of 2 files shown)


kravich:
Turns out you have a CI bot to run XCode tests. I'll add tests and verify them using Travis.

kravich:
Nah, this bot seems to just build player sources, and it does not run tests. However, I've added new tests. I didn't verify them but they should be ok since they are just a copy-paste of exisiting test_PathStripFileUriScheme_ReturnStripped with required modifications.

Alexey-Yakovenko:
@kravich can you provide an example when files have relative paths in the URI metadata field?
This should not happen, and I'm assuming it to be a bug, but maybe there's a case I forgot about.

osx/Tests/TitleFormattingTests.m
```diff
@@ -1341,6 +1341,30 @@ - (void)test_RawPathWithFileUriScheme_ReturnUnStripped {
...
+    tf_free (bc);
+    XCTAssert(!strcmp (buffer, "file:///path/to/filename.mp3"), @"The actual output is: %s", buffer);
+}
+
+- (void)test_RawPathRelativePath_ReturnsExpected {
+    pl_replace_meta (it, ":URI", "relative/path/to/filename.mp3");
+    char *bc = tf_compile("%_path_raw%");
+    tf_eval (&ctx, bc, buffer, 1000);
+    tf_free (bc);
+    XCTAssert(!strcmp (buffer, "file:relative/path/to/filename.mp3"), @"The actual output is: %s", buffer);
```
Alexey-Yakovenko: I'm not sure this makes any sense... I think it should just return a normal URI-path: `file://relative/path/...`
Although even that is fairly useless.

tf.c
```diff
@@ -2343,7 +2343,55 @@ tf_eval_int (ddb_tf_context_t *ctx, const char *code, int size, char *out, int o
...
+                        int is_absolute = (isalpha (v[0]) && v[1] == ':' && v[2] == '/');
+                        #else
+                        int is_absolute = (v[0] == '/');
+                        #endif
+
+                        if (is_absolute) {
+                            // This is an absolute path, just prepend proper scheme
+                            tf_append_out (&out, &outlen, "file://", 7);
+                            #ifdef __MINGW32__
+                            tf_append_out (&out, &outlen, "/", 1);
```
Alexey-Yakovenko: This looks like windows compatibility, and I don't see why `__MINGW32__` macro should be used for that.
Why not `_WIN32`?

Alexey-Yakovenko: Also there's no real need to append them as 2 separate string chunks..
I'd do it like this:
```
#ifdef _WIN32
const char prefix[] = "file:///";
#else
const char prefix[] = "file://";
#endif

tf_append_out (&out, &outlen, prefix, sizeof (prefix) - 1);
```

tf.c
```diff
@@ -2343,7 +2343,55 @@ tf_eval_int (ddb_tf_context_t *ctx, const char *code, int size, char *out, int o
...
+                                    break;
+                                }
+                            }
+
+                            if (is_uri) {
+                                // This is already a URI, just copy as is
+                                tf_append_out (&out, &outlen, v, strlen (v));
+                            }
+                            else {
+                                // This is a relative path, format it as file:path/to/track.mp3.
```
Alexey-Yakovenko: I propose to return an empty string in this case.

tf.c
kuba160: `__MINGW32__` macro is used throughout the deadbeef source code (there are currently no `_WIN32` macros). If you'd like I can make PR and replace `_MINGW32__` occurrences.

kravich:
> @kravich can you provide an example when files have relative paths in the URI metadata field?
> This should not happen, and I'm assuming it to be a bug, but maybe there's a case I forgot about.

@Alexey-Yakovenko To create playitem with relative path in :URI metadata field, one must pass relative path to plt_insert_file_int(). I found only one real-world use-case how to achieve that: to pass relative path in File->Add location window.

In all other cases I observed the fname passed to plt_insert_file_int() is always an absolute path or a URI, so it looks like relative paths in :URI work just by chance, because underlying APIs are happy to accept both kinds of paths.

Since relative path in :URI gives no benefits and just complicates things, it seems that the best solution will be to resolve relative paths to absolute inside plt_insert_file_int() and similar functions prior to allocating corresponding playitem. If you like this plan I may perform an investigation to make sure that :URI with relative path is not used for something useful and introduce proposed change in separate PR.

osx/Tests/TitleFormattingTests.m
kravich: Since you proposed to return empty string for relative path, in this test now I just check for empty output.

Alexey-Yakovenko:
@kravich It might be better to just check whether the path is relative, and ignore it.
The caller is supposed to perform the conversion.

tf.c
kravich: > Also there's no real need to append them as 2 separate string chunks..
Fixed.

As for the __MINGW32__ macro - it is already used everywhere across the source code for windows compatibility. Indeed, _WIN32 is a better candidate, but I originally decided to use __MINGW32__ to be consistent with the rest of the code.

I propose to leave __MINGW32__ for now, and to perform massive replacement of __MINGW32__ to _WIN32 in a separate PR as @kuba160 suggested.

However, I'll switch to _WIN32 if you confirm that it should be done in this PR.

tf.c
kravich: Done

kravich:
> @kravich It might be better to just check whether the path is relative, and ignore it.
> The caller is supposed to perform the conversion.

Yeah, this will also fix the issue. Do you want me to introduce this change?

Alexey-Yakovenko:
@kravich sure, if you have time. otherwise I can do it some time this weekend.

kravich:
> sure, if you have time

To be honest, not too much. I'd be happy to provide a fix, but it won't be fast. So if you have a time this weekend, it's better that you will do a fix. Sorry.

Alexey-Yakovenko:
ok I'll do it.

tf.c
```diff
@@ -2343,7 +2343,54 @@ tf_eval_int (ddb_tf_context_t *ctx, const char *code, int size, char *out, int o
                     }
                 }
                 else if (!strcmp (name, "_path_raw")) {
-                    val = pl_find_meta_raw (it, ":URI");
+                    const char *v = pl_find_meta_raw (it, ":URI");
+
+                    if (v) {
+                        #ifdef __MINGW32__
+                        int is_absolute = (isalpha (v[0]) && v[1] == ':' && v[2] == '/');
```
Alexey-Yakovenko: This would cause buffer overrun if the path is shorter than 3 bytes.

tf.c
```diff
@@ -2343,7 +2343,54 @@ tf_eval_int (ddb_tf_context_t *ctx, const char *code, int size, char *out, int o
...
+                                    break;
+                                }
+                            }
+
+                            if (is_uri) {
+                                // This is already a URI, just copy as is
+                                tf_append_out (&out, &outlen, v, strlen (v));
+                            }
+                            else {
+                                // This is a relative path. Since file:// URI with relative path is useless and
```
Alexey-Yakovenko: This comment should be updated to say that relative paths are considered invalid.
One line is enough.

kravich: Fixed

tf.c
kravich: Actually, because of [short-circuit](https://en.wikipedia.org/wiki/Short-circuit_evaluation) evaluation nature of logical expressions this expression will properly handle paths shorter than 3 bytes:
* if path len is 0, isalpha (v[0]) will evaluate to false and (v[1] == ':' && v[2] == '/') won't be evaluated at all
* if path len is 1 and contains alpha, (v[1] == ':') will be evaluated to false and (v[2] == '/') won't be evaluated at all
* etc

tf.c
kravich: I've also replaced __MINGW32__ macro with _WIN32. Since you already use _WIN32 in your commits, no reason to stick with __MINGW32__ anymore =)

tf.c
Alexey-Yakovenko: Ah yes, your'e right - I didn't think about that.
👍

PR #2230, opened 2019-10-03
Two new play modes

kpcee:
This pull request adds two new play modes:

**Keep Artist**
Retains the artist of the current song and automatically searches the playlist sequentially for the next one with identical artist. When the end is reached, the search begins from top of the playlist.

**Top Rated Tracks**
Randomly selects a song from the playlist, from this starting point the next song will be selected with a rating of at least 3 stars. Songs can be rated with the Song Rating plugin.

pic

### DIFF
--- a/deadbeef.h
+++ b/deadbeef.h
@@ -328,6 +328,8 @@ enum playback_order_t {
     PLAYBACK_ORDER_SHUFFLE_TRACKS = 1,
     PLAYBACK_ORDER_RANDOM = 2,
     PLAYBACK_ORDER_SHUFFLE_ALBUMS = 3,
+    PLAYBACK_ORDER_ARTIST = 4,
+    PLAYBACK_ORDER_TOP_RATED = 5,
 };
 
 // playback modes
--- a/plugins/gtkui/actionhandlers.c
+++ b/plugins/gtkui/actionhandlers.c
@@ -783,6 +783,30 @@ action_playback_loop_all_handler(DB_plugin_action_t *act, int ctx) {
 }
 
 gboolean
+action_playback_order_top_rated_handler_cb (void *data) {
+    gtk_check_menu_item_set_active (GTK_CHECK_MENU_ITEM (lookup_widget (mainwin, "order_top_rated")), 1);
+    return FALSE;
+}
+
+int
+action_playback_order_top_rated_handler(DB_plugin_action_t *act, int ctx) {
+    g_idle_add (action_playback_order_top_rated_handler_cb, NULL);
+    return 0;
+}
+
+gboolean
+action_playback_order_artist_handler_cb (void *data) {
+    gtk_check_menu_item_set_active (GTK_CHECK_MENU_ITEM (lookup_widget (mainwin, "order_artist")), 1);
+    return FALSE;
+}
+
+int
+action_playback_order_artist_handler(DB_plugin_action_t *act, int ctx) {
+    g_idle_add (action_playback_order_artist_handler_cb, NULL);
+    return 0;
+}
+
+gboolean
 action_playback_order_random_handler_cb (void *data) {
     gtk_check_menu_item_set_active (GTK_CHECK_MENU_ITEM (lookup_widget (mainwin, "order_random")), 1);
     return FALSE;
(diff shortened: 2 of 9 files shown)


Alexey-Yakovenko:
I'm not sure it's a good idea to introduce a new built-in playback mode which relies on additional plugin to be installed. It would make more sense to integrate such functionality in the rating plugin.
I don't know if that is possible using the current plugin API though. But we can look what additional API entry points are necessary.

kpcee:
The plugin is only needed to change the rating and saves it in the tag of the file. If Deadbeef can read the [Popularimeter](http://id3.org/id3v2.3.0#sec4.18) tag at some point, you could use the official tag.

Alexey-Yakovenko:
I'm sorry, but I can't accept a feature that can't be used without installing an additional plugin.
It also doesn't make any sense to me to write ratings into audio files.
This information is user-local, and should be stored in some kind of local media library / database.

Alexey-Yakovenko:
As I explained before, it is still possible to discuss how to get to the menu from plugins, and extend them.
It won't be easy or cross-platform, but it will do what you want.

kpcee:
Is not further bad, which is not taken over I just use in my local fork. However, it makes a lot of sense to save the rating in the audio file, especially as the official ID3 standard provides it.
1. If the path change, a renaming of a single directory is enough, all ratings of the affected files are lost in the database.

2. I come from a good but abandoned music player (gmusicbrowser) with a database and could save my ratings only with effort. Thanks to the Popularimeter frame, you can now take your ratings with you when you change the player without data loss, as long as the player supports it. That's great, that's really great.

Alexey-Yakovenko:
I fully understand your point, and you being ok with and even wanting to write user-specific data into audio files. This is not what I want to do in deadbeef by default though.

kpcee:
Another idea:
gmusicbrowser has the useful playback modes "Keep Artist" and "Keep Album". If this meets more interest, I could change "Top Rated Songs" accordingly.

Alexey-Yakovenko:
I don't mind those modes, but some more thought is needed. I don't like placing them into Shuffle menu, they don't belong there logically.

kpcee:
Okay, I'm open for ideas. But if we don't place them under Shuffle we would have to consider the setting made there, which would greatly increase code complexity and redundancy.

Alexey-Yakovenko:
What about gmusicbrowser? Where does it place all the modes? Do they interfere with shuffle?
Can you enable shuffle + keep artist at the same time?

kpcee:
It's been quite a while since I used gmusicbrowser the last time, I have to check it out first.

kpcee:
gmusicbrowser is a very complex program and offers many possibilities. That's why I took a screenshot. You can set whether the playback should be linear, random or weighted, you can also create your own rules. At the top of the screenshot at the lock and artist you can set "Keep Album" and "Keep Artist". The playback then follows the settings linear or random even if you have activated "Keep Album" or "Keep Artist" and respect them.
In other words, if you wanted to do it like gmusicbrowser, "Keep Album" or "Keep Artist" would have to respect the settings under Shuffle. At least so far that Off, Tracks and Random would be considered.

screenshot

Alexey-Yakovenko:
Well this is exactly what I meant. This filtering needs to be applied on top of shuffle mode, not replace it.

kpcee:
If you see it strictly then Shuffle -> Album would also have to be swapped out. It plays the tracks linearly and is like a brother of Keep Artist and Keep Album. It's just a design question how far you want to divide it up.

Alexey-Yakovenko:
I have to disagree. Shuffle albums mode only changes the shuffle algorithm, it doesn't filter tracks.
Keep artist and keep album -- well, I'm still not 100% sure what they do, but assuming they skip any non-matching tracks, that's filtering, and doesn't have anything to do with shuffling.

kpcee:
When I activate Shuffle Albums, whole albums are played from the first to the last song in sequence. Then deadbeef jumps to the first song of another album. The procedure repeats itself. And Shuffle Albums filter tracks, tracks that have already been played before are skipped.

**Keep Album:**
An album is played just like Shuffle Albums, but instead of playing another album it jumps back to the beginning and no skipping of played tracks (that's how I wanted to do it).

**Keep Artist:**
All songs of the current artist are played one after the other, when the end of the playlist is reached the playback starts again from the beginning of the playlist (that's how it's done in PR).

So for me it's all very similar in behavior. If you don't see it that way, it's okay.

Alexey-Yakovenko:
> An album is played just like Shuffle Albums, but instead of playing another album it jumps back to the beginning and no skipping of played tracks (that's how I wanted to do it).

In a sense, yeah, but it would still prevent using shuffle tracks mode on a single album, so it limits the potential functionality.

I think this should be a separate function, that can be used in combination with the current shuffle mode.

PR #2238, opened 2019-10-09
gettext/intl fixup

rindeal:
deadbeef currently doesn't build with newer gettext versions, spitting this message:

```
Use of AM_GNU_GETTEXT without [external] argument is no longer supported.
```

So this is my attempt at fixing this situation. I don't use NLS myself so please try if everything works with this patch applied.

I also noticed there was an unfinished attempt here - https://github.com/DeaDBeeF-Player/deadbeef/pull/2027.

First issues with the bundled gettext were reported in 2016 - https://github.com/DeaDBeeF-Player/deadbeef/issues/1530.


Alexey-Yakovenko:
Yeah.. the previous attempt didn't work. It compiled, but localization was gone.
Thanks for giving a try to fix this. I'll be away from computer for around a week, but I'll check it out as soon as I can.

Alexey-Yakovenko:
Can I ask you to provide a hint how intl folder was updated? I would like to verify integrity of the files before merging.

Alexey-Yakovenko:
nvm, found the --intl parameter of gettextize. Will get back to you shortly.

configure.ac
```diff
@@ -17,16 +17,12 @@ AC_PROG_INSTALL
 dnl AC_PROG_LIBTOOL
 AC_CONFIG_MACRO_DIR([m4])
 AC_C_BIGENDIAN
-AM_GNU_GETTEXT
 PKG_PROG_PKG_CONFIG
 AM_PROG_CC_C_O
+AM_ICONV
```
Alexey-Yakovenko: Why is this needed?

.travis.yml
```diff
@@ -3,7 +3,7 @@ language: c
 matrix:
     include:
         - os: linux
-          dist: trusty
```
Alexey-Yakovenko: Does it mean that deadbeef won't compile on trusty anymore?
As far as I remember, compiling on bionic breaks something in regards with binary compatibility, and the resulting build doesn't run on older ubuntu versions.

rindeal: The build failed back then on Trusty (https://travis-ci.org/DeaDBeeF-Player/deadbeef/jobs/595407067), but it seems it was just a transient Travis-CI bug. because it builds now.

configure.ac
rindeal: It sets `HAVE_ICONV` for:

https://github.com/DeaDBeeF-Player/deadbeef/blob/68e32fcb33fa1f9805d2a1b2cd894b68ceacd782/junklib.c#L34-L49

rindeal:
> Can I ask you to provide a hint how intl folder was updated? I would like to verify integrity of the files before merging.

`intl` dir was completely removed

Alexey-Yakovenko:
Oof. Sorry, I didn't pay enough attention, and thought the files were updated to a different libintl version.
Deleting libintl raises a different concern though...
Now deadbeef will link to whatever libintl is default in the system, which adds a new dependency, which will break portable builds.

Alexey-Yakovenko:
There are at least two things I need to verify before this can be merged:

1. I can't build a portable build locally anymore (ubuntu 18.04) -- it errors with gettext makefile version mismatch error
2. After I can build that, I need to verify that libintl is not linked dynamically, and if it does -- this needs to be fixed

Alexey-Yakovenko:
ok I don't know why this builds with normal configure && make, but there's a visible problem.
It's using system gettext / libintl, but the `po/Makefile.in.in` says it's expecting version 0.18.
My ubuntu is using gettext 0.19, so the build fails when I run `./scripts/static_build.sh`

Alexey-Yakovenko:
Looks like this difference comes from using autoreconf instead of autogen.sh, the generated (copied) po/Makefile.in.in is very different between these two cases.

Alexey-Yakovenko:
So when I'm replacing autoreconf with autogen.sh, and run static_build.sh, I get an infinite loop trying to do something with files in the po directory. Something is very wrong somewhere.

Alexey-Yakovenko:
After looking more careful at what's going on:

* Nuking libintl copy seems like a no go, because it's not guaranteed to be present on all targets
* We need to be sure it's linked when necessary. It was explicit before that patch, but with the patch it's not even clear how to verify that.
* I tried to make a smaller minimal version of this patch to use external argument of gettext macro. This still seems to remove the possibility of directly linking a copy of libintl.
* It might be possible to remove the AM_GNU_GETTEXT macro completely, and always explicitly link libintl instead. This may result in some defines to be removed, and I'll need to figure out what needs to be set. Trying all this now

Alexey-Yakovenko:
Now.. I looked at the source code of AM_GNU_GETTEXT macro. It appears as it supports building and linking the included libintl, and always assumes it to be in `${top_builddir}/intl`, so this case is covered. I'll try to make a working patch.

Alexey-Yakovenko:
I tried to come up with a minimal patch, but I couldn't get a build with working localization so far.
The localization files get built and installed, but interface always comes up in english.
Seems to be the same issue as the previous PR, where LOCALEDIR was set to `/usr/local//locale` instead of `/usr/local/share/locale`.
Trying to figure this part out.

Alexey-Yakovenko:
(Closed PR since it's not mergeable anyway)

Alexey-Yakovenko:
I think I nailed it. The localdir path is the variable @localedir@, but it can only correctly expand from Makefile.am, and not from configure.ac. The regular build works now, trying portable.

Alexey-Yakovenko:
Pushed on master, revision 70339732d8530570861d8fe26112b9f203077e91
Please let me know whether this works for you.

rindeal:
I'm not currently able to review/test/... it, hopefully will get to this by the end of the week.

rindeal:
Yes, it works in my build system. Thank you.

stefson:
@Alexey-Yakovenko current release v1.8.2 is broken with gettext-0.20, and this monster patch fixes it. So do you think you could push out a new release, say v1.8.3?

Alexey-Yakovenko:
@stefson The 1.8.3alpha1 includes this patch. I will release it as soon as it's ready.

Alexey-Yakovenko:
@stefson 
FYI: this change has caused issues with building deadbeef from tarball, therefore I have to revert a part of it.. I suppose certain compatibility issues will come back, but without this revert I can't figure out how to build a working tarball at all. Suggestions of a better fix are welcome.

Issue #2231, opened 2019-10-04
Start mimimized with GTK3 - empty window appears


spvkgn:
### Steps to reproduce the problem

Start player with clean config, set GTK3 GUI and check minimize on startup option, restart.

### What's going on? Describe the problem in as much detail as possible.

DeaDBeeF window is empty and not minimized:

![Снимок экрана в 2019-10-04 19-23-56](https://user-images.githubusercontent.com/4147135/66215367-eca03400-e6b2-11e9-89e2-50cc8bd30f90.png)

### Information about the software:

Deadbeef version: devel
OS: Ubuntu 18.04 MATE

Alexey-Yakovenko:
@kpcee do you want to look into this issue?
I didn't realize that I haven't tested the "start hidden" feature with GTK3, but turned out it doesn't work..

kpcee:
As far as I can see this is due to a changed behaviour of Gtk3. The following source code automatically displays the window independent of gtk_widget_show (mainwin);. That is why the window is displayed without layout.
```
#if GTK_CHECK_VERSION(3,10,0)
#if USE_GTK_APPLICATION
    // This must be called before window is shown
    gtk_application_add_window ( GTK_APPLICATION (gapp), GTK_WINDOW (mainwin));
#endif
#endif
```
But we can't do without it either, because otherwise the gui plugin is closed automatically. We can prevent this with g_application_hold(), but even here the window is displayed automatically. If this is no bug in Gtk3, then we can't repair this imho.
We can go back as it was before, but keep the changes in the preference dialog, or we have to consider Gtk2 and Gtk3 separately.

Alexey-Yakovenko:
Is it possible to defer this call to until the window needs to be shown?

Alexey-Yakovenko:
I haven't tried this, but it looks like it should be possible to group this call with `gtk_widget_show (mainwin);`, and possibly add a global variable flag to ensure it's called only once. something like a new function `gtkui_mainwindow_show`, which would do all the checks, call `gtk_application_add_window` if needed (once), and then `gtk_widget_show (mainwin);`.

That is of course assuming no other problems will happen.. I haven't tried anything. But I can try this later.

Alexey-Yakovenko:
ok I tried this, and you are correct. without any windows, the g_application_run would just return immediately, and the app would quit.

Alexey-Yakovenko:
[...]

PR #2243, opened 2019-10-20
Bugfix for #2231

kpcee:
Here is a possible bugfix for #2231 that restores the old behavior for Gtk3. With my older local copy it works fine. However, the current Git version has a problem and cannot remember the correct position of the window if you click on the tray icon several times. This is not due to my submitted patch because this behavior can be observed even before when you revert to a previous state. There must have been a change that caused this, because it still works in 1.82.

### DIFF
--- a/plugins/gtkui/gtkui.c
+++ b/plugins/gtkui/gtkui.c
@@ -304,6 +304,13 @@ show_traymenu (void) {
     g_idle_add (show_traymenu_cb, NULL);
 }
 
+#if GTK_CHECK_VERSION(3,0,0)
+static gboolean
+mainwin_hide_cb (gpointer data) {
+     gtk_widget_hide (mainwin);
+     return FALSE;
+ }
+#endif
 
 void
 mainwin_toggle_visible (void) {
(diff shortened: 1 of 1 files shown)


Alexey-Yakovenko:
It's quite possible, that the other problem was caused by the change that added "start hidden" option.

kpcee:
Are you able to reproduce the problem? I have now tried until I found the commit that triggers the problem. For me the problem starts with the commit:

> commit bbff7bbb80e041fadd91624f1ae11b97ddbbbc1c (HEAD -> Bugfix)
> Date:   Mon Jun 3 18:07:07 2019 +0000
> gtkui: restore window position relative to the monitor DeaDBeeF is currently on

Alexey-Yakovenko:
@saivert any comments to this? would it be possible to re-verify your change and fix it?
It's quite risky to accumulate changes in the same area, when it's impossible to verify their correctness due to tech debt.

Alexey-Yakovenko:
@kpcee sorry, I only started looking at a few PRs and bugs yesterday. Only got to reading the comments / summaries, haven't tried reproducing.

Alexey-Yakovenko:
@kpcee seems like the author of the original change is not available, and I will have to revert his commit.
It doesn't seem important to me to have multimonitor support for windows position saving, when singlemonitor is not working. But I will first need to test this anyway.

saivert:
Quite possible because I'm not zeroing out the struct for the monitor rect in that patch.
In https://github.com/DeaDBeeF-Player/deadbeef/commit/bbff7bbb80e041fadd91624f1ae11b97ddbbbc1c#diff-553100c0eadbbed1e24f24dcc5e1cf1dR49

Alexey-Yakovenko:
@saivert this change solves the problem indeed. great spot! thanks a lot. I'm adding the fix.

Alexey-Yakovenko:
@kpcee your fix works fine too! great job, merging.

PR #2251, opened 2019-10-28
Track Gain All Unscanned Files In Playlist

kpcee:
This Pull Request adds a new entry to the ReplayGain Scanner: Track Gain All Unscanned Files In Playlist.
It is not easy to see which files are replay gained and which are not. Especially in larger collections it quickly becomes chaotic. The new routine therefore checks all files in the current playlist to see if they have been replay gained, and only processes those that are not. This is fast and efficient.

### DIFF
--- a/plugins/gtkui/gtkui.c
+++ b/plugins/gtkui/gtkui.c
@@ -1592,12 +1592,20 @@ static DB_plugin_action_t action_rg_scan_per_file = {
     .next = &action_rg_scan_selection_as_album
 };
 
+static DB_plugin_action_t action_scan_all_tracks_without_Rg = {
+    .title = "ReplayGain/Track Gain All Unscanned Files In Playlist",
+    .name = "scan_all_tracks_without_Rg",
+    .flags = DB_ACTION_SINGLE_TRACK | DB_ACTION_MULTIPLE_TRACKS | DB_ACTION_ADD_MENU,
+    .callback2 = action_scan_all_tracks_without_Rg_handler ,
+    .next = &action_rg_scan_per_file
+};
+
 static DB_plugin_action_t action_deselect_all = {
     .title = "Edit/Deselect All",
     .name = "deselect_all",
     .flags = DB_ACTION_COMMON,
     .callback2 = action_deselect_all_handler,
-    .next = &action_rg_scan_per_file
+    .next = &action_scan_all_tracks_without_Rg
 };
 
 static DB_plugin_action_t action_select_all = {
--- a/plugins/gtkui/gtkui.c
+++ b/plugins/gtkui/gtkui.c
@@ -1592,11 +1592,11 @@ static DB_plugin_action_t action_rg_scan_per_file = {
     .next = &action_rg_scan_selection_as_album
 };
 
-static DB_plugin_action_t action_scan_all_tracks_without_Rg = {
-    .title = "ReplayGain/Track Gain All Unscanned Files In Playlist",
+static DB_plugin_action_t action_scan_all_tracks_without_rg = {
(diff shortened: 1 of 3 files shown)


plugins/gtkui/gtkui.c
```diff
@@ -1592,12 +1592,20 @@ static DB_plugin_action_t action_rg_scan_per_file = {
     .next = &action_rg_scan_selection_as_album
 };
 
+static DB_plugin_action_t action_scan_all_tracks_without_Rg = {
```
Alexey-Yakovenko: `Rg` should be changed to `rg` for consistency, everywhere.

plugins/gtkui/gtkui.c
```diff
@@ -1592,12 +1592,20 @@ static DB_plugin_action_t action_rg_scan_per_file = {
...
+    .title = "ReplayGain/Track Gain All Unscanned Files In Playlist",
```
Alexey-Yakovenko: This should better be named `Scan Per-file Track Gain If Not Scanned`.
TBH at this point I think it might be better to implement a function like "select tracks without track gain" or something like that, and then they could be scanned using the regular Track Gain menu item. What do you think?
Might be useful for other purposes too.
Seems like scanning with filter would be doing too many things which should be separate functions, and duplicating functionality somewhat.

plugins/gtkui/rg.c
```diff
@@ -674,3 +674,47 @@ action_rg_remove_info_handler (struct DB_plugin_action_s *action, int ctx) {
 
     return 0;
 }
+
+int
+action_scan_all_tracks_without_Rg_handler (struct DB_plugin_action_s *action, int ctx) {
+  int count = 0;
```
Alexey-Yakovenko: indentation

plugins/gtkui/gtkui.c
Alexey-Yakovenko: Right now, deadbeef doesn't have advanced search like that, therefore I will accept this PR, but for the future it's better to look at a bigger picture and try to solve problems at macro scale, instead of hardcoding filtering into every feature where subset of tracks is needed.

kpcee:
Updated.

plugins/gtkui/rg.c
```diff
@@ -674,3 +674,47 @@ action_rg_remove_info_handler (struct DB_plugin_action_s *action, int ctx) {
...
+    if (!plt) {
+        return 0;
+    }
+
+    ddb_replaygain_settings_t s;
+    s._size = sizeof (ddb_replaygain_settings_t);
+
+    deadbeef->pl_lock ();
+    int tc = deadbeef->plt_get_item_count (plt, PL_MAIN);
+    tracks = calloc (tc, sizeof (DB_playItem_t *));
```
Alexey-Yakovenko: This may cause unpredictable behavior, if a number of tracks is 0, e.g. if you run this from a playlist context menu on an empty playlist.

kpcee:
Okay, I've added a check.

plugins/gtkui/rg.c
```diff
@@ -674,3 +674,52 @@ action_rg_remove_info_handler (struct DB_plugin_action_s *action, int ctx) {
...
+        return 0;
+    }
+
+    ddb_replaygain_settings_t s;
+    s._size = sizeof (ddb_replaygain_settings_t);
+
+    deadbeef->pl_lock ();
+
+    int tc = deadbeef->plt_get_item_count (plt, PL_MAIN);
+    if (!tc) {
```
Alexey-Yakovenko: plt_unref is missing

kpcee: Okay I have added it and also in other places where it was missing in the source.

Issue #1602, opened 2016-06-29
Change Last.fm scrobbler URL to https://post.audioscrobbler.com (HTTPS)


anewuser:
The current default URL  is `http://post.audioscrobbler.com` . I've changed it to `https://post.audioscrobbler.com` and it works.

Alexey-Yakovenko:
http works fine here.
Seems like your local problem.
But it's great that you found how to fix it.

anewuser:
I didn't mean that http wasn't working. I meant that it's a better option, since the plugin sends our passwords.

anewuser:
I meant that HTTPS would be better since the plugin has our passwords, but after opening my profile I saw that only one of the albums I listened to was scrobbled. Maybe the HTTPS URL just redirects to the HTTP sometimes, and sometimes it gives an error. I can't investigate it now, and since you plan to eventually port the plugin to the new API, it's better to leave it as it is for now.

Alexey-Yakovenko:
Deadbeef's VFS_CURL doesn't support HTTPS on linux yet (in the official binary builds).

I'll consider changing this when this problem is fixed.

anewuser:
Thank you for your replies and for your work on this great player.

Alexey-Yakovenko:
Reopening as a feature request (I originally thought that it's a but report claiming it doesn't work without https)

spirine:
last.fm do not support HTTPS scrobbling: http://www.last.fm/api/intro
Maybe https://post.audioscrobbler.com just resolve as same IP for scrobbler

Alexey-Yakovenko:
@spirine I can't see any mention of last.fm not supporting scrobbling over HTTPS on your link.

I'm pretty sure it works fine. And yes, ip address can be the same for both HTTP and HTTPS, which doesn't really change the fact.

m0ngr31:
Mine doesn't work until I use HTTPS as well.

servimo:
Now scrobbling is not working in any way

servimo:
Deadbeef scrobbler is not working

4oo4:
@servimo The default is using the latest API endpoint (https://ws.audoscrobbler.com), which deadbeef can't even use since it doesn't work with HTTPS. So you need to use the plain HTTP endpoint (http://post.audioscrobbler.com). HTTPS support really needs to be something that gets looked at, especially nowadays, it's only a matter of time before the HTTP endpoint gets shut off.

[...]

PR #2263, opened 2019-11-08
WIP: Test Fix for #1602

4oo4:
Set `curl_easy_setopt` to use `CURLUSESSL_CONTROL`, where we force all scrobbling requests over HTTPS and fail the connection if we cannot establish TLS. Both last.fm and libre.fm allow for HTTPS scrobbling, so there's no reason to keep using plaintext HTTP.

https://curl.haxx.se/libcurl/c/CURLOPT_USE_SSL.html


4oo4:
Actually, I think `CURLUSESSL_ALL` is better, going to switch to that.

4oo4:
Yeah, I think this needs a rewrite to use the new version of the new version of the API before it can use TLS.

PR #2265, opened 2019-11-09
windows: is_relative_path fix for windows

kuba160:
- Fixed order so that `is_relative_path` is called after conversion
- More structured approach of detection. It is not covering all the cases but I guess it will be fine. We could also consider using [`PathIsRelative()`](https://docs.microsoft.com/en-us/windows/win32/api/shlwapi/nf-shlwapi-pathisrelativea) from windows api

### DIFF
--- a/playlist.c
+++ b/playlist.c
@@ -967,18 +967,6 @@ plt_insert_file_int (int visibility, playlist_t *playlist, playItem_t *after, co
         fname += 7;
     }
 
-    // now that it's known we're not dealing with URL, check if it's a relative path
-    if (is_relative_path (fname)) {
-        return NULL;
-    }
-
-    // detect decoder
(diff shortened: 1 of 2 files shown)


kuba160:
note: Build fails on current master

Alexey-Yakovenko:
I know.. I did some refactoring and added some deprecations which were missing, but now gcc freaks out. Still trying to figure out what's wrong.

Alexey-Yakovenko:
It seems that it's an older gcc (4.8.4) which can't handle attribute deprecated on enum values.. I can build fine locally with gcc 7.4.0. So.. I suppose I'll just remove the deprecations then.

Alexey-Yakovenko:
or.. I can change the macro to do nothing on older GCC. I think that's more portable.

PR #2269, opened 2019-11-16
streamer: fix allocation size

huglovefan:
the `scope` gui widget (and probably others) cause a segfault without this

might fix https://github.com/DeaDBeeF-Player/deadbeef/issues/2268


Alexey-Yakovenko:
Ugh. Thanks for spotting this. It's a really lame mistake.

PR #2270, opened 2019-11-19
Better Remove From Disk

kpcee:
Improved Remove From Disk routine 

* The warning dialog no longer shows a standard message, it shows what will be deleted 
* Deleted songs will be removed from all other playlists as well
* If desired the next song will be played when the currently played song is deleted (respects playback mode and play queue)
* Behavior is adjustable in the GUI (switched off by default)
pic

### DIFF
--- a/plugins/gtkui/actionhandlers.c
+++ b/plugins/gtkui/actionhandlers.c
@@ -46,6 +46,8 @@
 
 extern GtkWidget *mainwin;
 extern DB_functions_t *deadbeef;
+static DB_plugin_t *plugin;
+#define trace(...) { deadbeef->log_detailed (plugin, 0, __VA_ARGS__); }
 
 gboolean
 action_open_files_handler_cb (void *userdata) {
@@ -358,6 +360,27 @@ action_remove_from_playlist_handler (DB_plugin_action_t *act, int ctx) {
 }
 
 void
+remove_deleted_file_from_all_playlists (const char *search_uri) {
+    // The caller is responsible for pl_lock
+    int n = deadbeef->plt_get_count ();
+    for (int i = 0; i < n; ++i) {
+        ddb_playlist_t *plt = deadbeef->plt_get_for_idx (i);
+        DB_playItem_t *it = deadbeef->plt_get_first (plt, PL_MAIN);
+        while (it) {
+            DB_playItem_t *next = deadbeef->pl_get_next (it, PL_MAIN);
+            const char *uri = deadbeef->pl_find_meta (it, ":URI");
+            if (strcmp (uri, search_uri) == 0) {
+                deadbeef->plt_remove_item (plt, it);
+            }
+            deadbeef->pl_item_unref (it);
+            it = next;
+        }
+
+        deadbeef->plt_unref (plt);
+    }
+}
+
+void
 delete_and_remove_track (const char *uri, ddb_playlist_t *plt, ddb_playItem_t *it) {
     (void)unlink (uri);
 
@@ -365,31 +388,12 @@ delete_and_remove_track (const char *uri, ddb_playlist_t *plt, ddb_playItem_t *i
     struct stat buf;
     memset (&buf, 0, sizeof (buf));
     int stat_res = stat (uri, &buf);
+    
     if (stat_res != 0) {
         deadbeef->plt_remove_item (plt, it);
-    }
-}
-
-void remove_deleted_file_from_all_playlists (const char *search_uri) {
-    for (int i = 0; ; ++i) {
-        ddb_playlist_t *plt = deadbeef->plt_get_for_idx (i);
-        if (!plt) break;
-
-        int save_playlist = 0;
-        DB_playItem_t *it = deadbeef->plt_get_first (plt, PL_MAIN);
-        while (it) {
-            DB_playItem_t *next = deadbeef->pl_get_next (it, PL_MAIN);
-            const char *uri = deadbeef->pl_find_meta (it, ":URI");
-            if (strcmp (uri, search_uri) == 0) {
-                deadbeef->plt_remove_item (plt, it);
-                save_playlist = 1;
-            }
-            deadbeef->pl_item_unref (it);
-            it = next;
-        }
-
-        deadbeef->plt_unref (plt);
-        if (save_playlist) deadbeef->plt_save_n (i);
+        remove_deleted_file_from_all_playlists (uri);
+    } else {
+        trace("Failed to delete file: %s\n", uri);
     }
 }
 
@@ -399,28 +403,30 @@ action_delete_from_disk_handler_cb (void *data) {
     if (!plt) {
         return FALSE;
     }
-
-    int ctx = (intptr_t)data;
-
+    
+    int ctx = (int)(intptr_t)data;
     if (deadbeef->conf_get_int ("gtkui.delete_files_ask", 1)) {
         char buf[1000];
-        char buf2[500] = "\nAll files will be lost.\n\n(This dialog can be turned off in GTKUI plugin settings)";
+        const char *buf2 = _(" The files will be lost.\n\n(This dialog can be turned off in GTKUI plugin settings)");
 
         if (ctx == DDB_ACTION_CTX_SELECTION) {
             int selected_files = deadbeef->pl_getselcount ();
-            if (selected_files == 1) snprintf(buf, 1000, "Do you really want to delete the selected file?%s", buf2);
-            else snprintf(buf, 1000, "Do you really want to delete all %d selected files?%s", selected_files, buf2);
+            if (selected_files == 1) {
+                snprintf(buf, sizeof (buf), _("Do you really want to delete the selected file?%s"), buf2);
+            } else {
+                snprintf(buf, sizeof (buf), _("Do you really want to delete all %d selected files?%s"), selected_files, buf2);
+            }
         }
         else if (ctx == DDB_ACTION_CTX_PLAYLIST) {
             int files = deadbeef->plt_get_item_count (plt, PL_MAIN);
-            snprintf(buf, 1000, "Warning!\nDo you really want to delete the whole playlist? This applies to all %i files!%s", files, buf2);
+            snprintf(buf, sizeof (buf), _("Do you really want to delete all %d files from the current playlist?%s"), files, buf2);
         }
         else if (ctx == DDB_ACTION_CTX_NOWPLAYING) {
-            snprintf(buf, 1000, "Do you really want to delete the current played song?%s", buf2);
+            snprintf(buf, sizeof (buf), _("Do you really want to delete the currently playing file?%s"), buf2);
         }
 
         GtkWidget *dlg = gtk_message_dialog_new (GTK_WINDOW (mainwin), GTK_DIALOG_MODAL, GTK_MESSAGE_WARNING, GTK_BUTTONS_YES_NO, _("Delete files from disk"));
-        gtk_message_dialog_format_secondary_text (GTK_MESSAGE_DIALOG (dlg), _(buf));
+        gtk_message_dialog_format_secondary_text (GTK_MESSAGE_DIALOG (dlg), buf);
         gtk_window_set_title (GTK_WINDOW (dlg), _("Warning"));
 
         int response = gtk_dialog_run (GTK_DIALOG (dlg));
@@ -430,7 +436,6 @@ action_delete_from_disk_handler_cb (void *data) {
         }
     }
     deadbeef->pl_lock ();
-
     
     DB_playItem_t *it_current_song = deadbeef->streamer_get_playing_track ();
     int idx_current_song = -1;
@@ -440,15 +445,14 @@ action_delete_from_disk_handler_cb (void *data) {
             DB_playItem_t *next = deadbeef->pl_get_next (it, PL_MAIN);
             const char *uri = deadbeef->pl_find_meta (it, ":URI");
             if (deadbeef->pl_is_selected (it) && deadbeef->is_local_file (uri)) {
-                if (it == it_current_song) idx_current_song = deadbeef->plt_get_item_idx (plt, it, PL_MAIN);
+                if (it == it_current_song) {
+                    idx_current_song = deadbeef->plt_get_item_idx (plt, it, PL_MAIN);
+                }
                 delete_and_remove_track (uri, plt, it);
-                remove_deleted_file_from_all_playlists (uri);
             }
             deadbeef->pl_item_unref (it);
             it = next;
         }
-
-        deadbeef->plt_save_config (plt);
     }
     else if (ctx == DDB_ACTION_CTX_PLAYLIST) {
         DB_playItem_t *it = deadbeef->plt_get_first (plt, PL_MAIN);
@@ -457,23 +461,10 @@ action_delete_from_disk_handler_cb (void *data) {
             const char *uri = deadbeef->pl_find_meta (it, ":URI");
             if (deadbeef->is_local_file (uri)) {
                 delete_and_remove_track (uri, plt, it);
-                remove_deleted_file_from_all_playlists (uri);
-// FIXME: this dialog should allow something like "cancel" and "ignore all", then
-// it will be usable
-//                else {
-//                    GtkWidget *dlg = gtk_message_dialog_new (GTK_WINDOW (mainwin), GTK_DIALOG_MODAL, GTK_MESSAGE_WARNING, GTK_BUTTONS_OK, _("Can't delete the file. Perhaps it doesn't exist, read-only, or part of read-only VFS, or all of the above."));
-//                    gtk_message_dialog_format_secondary_text (GTK_MESSAGE_DIALOG (dlg), uri);
-//                    gtk_window_set_title (GTK_WINDOW (dlg), _("Warning"));
-//
-//                    int response = gtk_dialog_run (GTK_DIALOG (dlg));
-//                    gtk_widget_destroy (dlg);
-//                }
             }
             deadbeef->pl_item_unref (it);
             it = next;
         }
-
-        deadbeef->plt_save_config (plt);
     }
     else if (ctx == DDB_ACTION_CTX_NOWPLAYING) {
         DB_playItem_t *it = deadbeef->streamer_get_playing_track ();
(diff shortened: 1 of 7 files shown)


plugins/gtkui/actionhandlers.c
```diff
@@ -370,12 +370,57 @@ delete_and_remove_track (const char *uri, ddb_playlist_t *plt, ddb_playItem_t *i
     }
 }
 
+void remove_deleted_file_from_all_playlists (const char *search_uri) {
+    for (int i = 0; ; ++i) {
```
Alexey-Yakovenko: It's better to iterate over the pl_get_count, and pl_lock while this is executed, otherwise you're risking a race condition.

plugins/gtkui/actionhandlers.c
```diff
@@ -370,12 +370,57 @@ delete_and_remove_track (const char *uri, ddb_playlist_t *plt, ddb_playItem_t *i
...
+        ddb_playlist_t *plt = deadbeef->plt_get_for_idx (i);
+        if (!plt) break;
```
Alexey-Yakovenko: In Deadbeef codebase there's a convention to write each operator on separate line, and put curly braces around operators even when not mandatory, like in this case.

plugins/gtkui/actionhandlers.c
```diff
@@ -370,12 +370,57 @@ delete_and_remove_track (const char *uri, ddb_playlist_t *plt, ddb_playItem_t *i
...
+
     int ctx = (intptr_t)data;
+
     if (deadbeef->conf_get_int ("gtkui.delete_files_ask", 1)) {
+        char buf[1000];
+        char buf2[500] = "\nAll files will be lost.\n\n(This dialog can be turned off in GTKUI plugin settings)";
+
+        if (ctx == DDB_ACTION_CTX_SELECTION) {
+            int selected_files = deadbeef->pl_getselcount ();
+            if (selected_files == 1) snprintf(buf, 1000, "Do you really want to delete the selected file?%s", buf2);
```
Alexey-Yakovenko: It's better to use `sizeof (buf)` instead of hardcoding the size in multiple places.

Alexey-Yakovenko: You also need to wrap every localizable string literal in `_()`

plugins/gtkui/actionhandlers.c
```diff
@@ -370,12 +370,57 @@ delete_and_remove_track (const char *uri, ddb_playlist_t *plt, ddb_playItem_t *i
...
+            int files = deadbeef->plt_get_item_count (plt, PL_MAIN);
+            snprintf(buf, 1000, "Warning!\nDo you really want to delete the whole playlist? This applies to all %i files!%s", files, buf2);
+        }
+        else if (ctx == DDB_ACTION_CTX_NOWPLAYING) {
+            snprintf(buf, 1000, "Do you really want to delete the current played song?%s", buf2);
+        }
+
         GtkWidget *dlg = gtk_message_dialog_new (GTK_WINDOW (mainwin), GTK_DIALOG_MODAL, GTK_MESSAGE_WARNING, GTK_BUTTONS_YES_NO, _("Delete files from disk"));
-        gtk_message_dialog_format_secondary_text (GTK_MESSAGE_DIALOG (dlg), _("Files will be lost. Proceed?\n(This dialog can be turned off in GTKUI plugin settings)"));
+        gtk_message_dialog_format_secondary_text (GTK_MESSAGE_DIALOG (dlg), _(buf));
```
Alexey-Yakovenko: This doesn't need `_()`, xgettext requires this macro to be used on literals to work.

plugins/gtkui/actionhandlers.c
```diff
@@ -434,19 +480,31 @@ action_delete_from_disk_handler_cb (void *data) {
...
         }
         deadbeef->plt_save_config (plt);
     }
+    deadbeef->pl_item_unref (it_current_song);
     deadbeef->pl_unlock ();
     deadbeef->plt_unref (plt);
 
     deadbeef->sendmessage (DB_EV_PLAYLISTCHANGED, 0, DDB_PLAYLIST_CHANGE_CONTENT, 0);
+    
+    if (deadbeef->conf_get_int ("gtkui.skip_deleted_songs", 0) && deadbeef->plt_get_item_idx (plt, it_current_song, PL_MAIN) == -1 && deadbeef->streamer_get_current_playlist () == deadbeef->plt_get_curr_idx () && deadbeef->get_output ()->state () == OUTPUT_STATE_PLAYING) {
```
Alexey-Yakovenko: This needs to be split into multiple lines for better readability.

plugins/gtkui/actionhandlers.c
```diff
@@ -434,19 +480,31 @@ action_delete_from_disk_handler_cb (void *data) {
...
+        if (idx_current_song != -1 && deadbeef->playqueue_get_count () == 0 && deadbeef->conf_get_int ("playback.order", PLAYBACK_ORDER_LINEAR) == PLAYBACK_ORDER_LINEAR) {
```
Alexey-Yakovenko: This also needs splitting into multiple lines

plugins/gtkui/actionhandlers.c
```diff
@@ -434,19 +480,31 @@ action_delete_from_disk_handler_cb (void *data) {
...
+            deadbeef->sendmessage (DB_EV_PLAY_NUM, 0, idx_current_song, 0);
```
Alexey-Yakovenko: This wouldn't play anything if removed song is the last song, although maybe not a problem.

Alexey-Yakovenko: I don't see why this can't be used for both cases. It seems like you always want to ignore shuffle order.

Alexey-Yakovenko: Also getting playback.order via conf_get_int and the PLAYBACK_ORDER_* constants are deprecated.
There's now an API and new enum for that. See `streamer_get_shuffle` and `streamer_get_repeat`.

plugins/gtkui/actionhandlers.c
```diff
@@ -370,12 +370,57 @@ delete_and_remove_track (const char *uri, ddb_playlist_t *plt, ddb_playItem_t *i
...
+        char buf2[500] = "\nAll files will be lost.\n\n(This dialog can be turned off in GTKUI plugin settings)";
+
+        if (ctx == DDB_ACTION_CTX_SELECTION) {
+            int selected_files = deadbeef->pl_getselcount ();
+            if (selected_files == 1) snprintf(buf, 1000, "Do you really want to delete the selected file?%s", buf2);
+            else snprintf(buf, 1000, "Do you really want to delete all %d selected files?%s", selected_files, buf2);
+        }
+        else if (ctx == DDB_ACTION_CTX_PLAYLIST) {
+            int files = deadbeef->plt_get_item_count (plt, PL_MAIN);
+            snprintf(buf, 1000, "Warning!\nDo you really want to delete the whole playlist? This applies to all %i files!%s", files, buf2);
```
Alexey-Yakovenko: Inconsistent use of `%d` vs `%i`, would be better to pick one.

plugins/gtkui/actionhandlers.c
Alexey-Yakovenko: Nevermind the pl_lock request, I can now see that this function is called from while locked.

plugins/gtkui/actionhandlers.c
```diff
@@ -434,19 +480,31 @@ action_delete_from_disk_handler_cb (void *data) {
         if (it) {
             const char *uri = deadbeef->pl_find_meta (it, ":URI");
             if (deadbeef->is_local_file (uri)) {
-                int idx = deadbeef->plt_get_item_idx (plt, it, PL_MAIN);
+                int idx = idx_current_song = deadbeef->plt_get_item_idx (plt, it, PL_MAIN);
                 if (idx != -1) {
                     delete_and_remove_track (uri, plt, it);
```
Alexey-Yakovenko: The `delete_and_remove_track` will only remove the track from playlist if unlink was successful (verified via calling stat).
The next line would remove the track from other playlists even if deletion has failed.
Also, now that deadbeef got log window, it would make sense to print an error message if delete has failed.

plugins/gtkui/prefwin.c
```diff
@@ -1086,6 +1086,14 @@ on_hide_delete_from_disk_toggled       (GtkToggleButton *togglebutton,
     deadbeef->conf_set_int ("gtkui.hide_remove_from_disk", active);
 }
 
+void
+on_skip_deleted_songs_toggled          (GtkToggleButton *togglebutton,
+                                        gpointer         user_data)
+{
+    int active = gtk_toggle_button_get_active (GTK_TOGGLE_BUTTON (togglebutton));
+    deadbeef->conf_set_int ("gtkui.skip_deleted_songs", active);
```
Alexey-Yakovenko: I can't see the code which would set the checkbox to its initial state.

plugins/gtkui/actionhandlers.c
```diff
@@ -412,6 +457,7 @@ action_delete_from_disk_handler_cb (void *data) {
             const char *uri = deadbeef->pl_find_meta (it, ":URI");
             if (deadbeef->is_local_file (uri)) {
                 delete_and_remove_track (uri, plt, it);
+                remove_deleted_file_from_all_playlists (uri);
 // FIXME: this dialog should allow something like "cancel" and "ignore all", then
```
Alexey-Yakovenko: Please remove this commented out block.
It's becoming irrelevant, if the unlink errors are logged.

plugins/gtkui/actionhandlers.c
```diff
@@ -370,12 +370,57 @@ delete_and_remove_track (const char *uri, ddb_playlist_t *plt, ddb_playItem_t *i
...
+}
+
 gboolean
 action_delete_from_disk_handler_cb (void *data) {
+    ddb_playlist_t *plt = deadbeef->plt_get_curr ();
+    if (!plt) {
+        return FALSE;
+    }
+
     int ctx = (intptr_t)data;
```
Alexey-Yakovenko: This cast causes a compilation warning. Needs an explicit cast to int.

plugins/gtkui/actionhandlers.c
```diff
@@ -370,12 +370,57 @@ delete_and_remove_track (const char *uri, ddb_playlist_t *plt, ddb_playItem_t *i
...
+            int selected_files = deadbeef->pl_getselcount ();
+            if (selected_files == 1) snprintf(buf, 1000, "Do you really want to delete the selected file?%s", buf2);
+            else snprintf(buf, 1000, "Do you really want to delete all %d selected files?%s", selected_files, buf2);
+        }
+        else if (ctx == DDB_ACTION_CTX_PLAYLIST) {
+            int files = deadbeef->plt_get_item_count (plt, PL_MAIN);
+            snprintf(buf, 1000, "Warning!\nDo you really want to delete the whole playlist? This applies to all %i files!%s", files, buf2);
+        }
+        else if (ctx == DDB_ACTION_CTX_NOWPLAYING) {
+            snprintf(buf, 1000, "Do you really want to delete the current played song?%s", buf2);
```
Alexey-Yakovenko: Replace with `Do you really want to delete the currently playing file?`

plugins/gtkui/actionhandlers.c
Alexey-Yakovenko: Replace with `Do you really want to delete all %d files from the current playlist?%s`
The word "warning" is unnecessary either.

plugins/gtkui/actionhandlers.c
```diff
@@ -370,12 +370,57 @@ delete_and_remove_track (const char *uri, ddb_playlist_t *plt, ddb_playItem_t *i
...
+    ddb_playlist_t *plt = deadbeef->plt_get_curr ();
+    if (!plt) {
+        return FALSE;
+    }
+
     int ctx = (intptr_t)data;
+
     if (deadbeef->conf_get_int ("gtkui.delete_files_ask", 1)) {
+        char buf[1000];
+        char buf2[500] = "\nAll files will be lost.\n\n(This dialog can be turned off in GTKUI plugin settings)";
```
Alexey-Yakovenko: Replace with `The files will be lost.\n\n......`

Alexey-Yakovenko: buf2 doesn't need to be writable and fixed size.
Replace with `const char *buf2 = ....`

plugins/gtkui/actionhandlers.c
Alexey-Yakovenko: Please remember to assign the return value of pl_get_count to a variable, to avoid calling it many times.

plugins/gtkui/actionhandlers.c
```diff
@@ -370,12 +370,57 @@ delete_and_remove_track (const char *uri, ddb_playlist_t *plt, ddb_playItem_t *i
...
+        if (!plt) break;
+
+        int save_playlist = 0;
+        DB_playItem_t *it = deadbeef->plt_get_first (plt, PL_MAIN);
+        while (it) {
+            DB_playItem_t *next = deadbeef->pl_get_next (it, PL_MAIN);
+            const char *uri = deadbeef->pl_find_meta (it, ":URI");
+            if (strcmp (uri, search_uri) == 0) {
+                deadbeef->plt_remove_item (plt, it);
+                save_playlist = 1;
```
Alexey-Yakovenko: You need to use plt_modified instead, and then just call pl_save_all at the end.
This would ensure that all other plugins / modules will know that the playlist(s) has changed.

Alexey-Yakovenko: Hmm.. actually you don't. It's done automatically by plt_remove_item.
And you don't need to call plt_save_n.
Replace plt_save_config call below with pl_save_all.
That should cover it all.

plugins/gtkui/actionhandlers.c
```diff
@@ -370,12 +370,57 @@ delete_and_remove_track (const char *uri, ddb_playlist_t *plt, ddb_playItem_t *i
     }
 }
 
+void remove_deleted_file_from_all_playlists (const char *search_uri) {
```
Alexey-Yakovenko: The convention is to have return type on separate line.

plugins/gtkui/actionhandlers.c
kpcee: The reason why DB_EV_PLAY_NUM is used here is to play the next song below the current one if it is deleted and if DDB_SHUFFLE_OFF is set. DB_EV_NEXT always jumps to the beginning of the playlist which is not correct in this case and is annoying for larger playlists.

plugins/gtkui/actionhandlers.c
kpcee: I'm not quite sure how to do this correctly. When I define and use 
```
extern DB_plugin_t main_plugin;
#define trace(...) { deadbeef->log_detailed (&main_plugin, 0, __VA_ARGS__); }
```
the GUI plugin doesn't start anymore.

plugins/gtkui/actionhandlers.c
kpcee: You mean `int ctx = (int)(intptr_t)data;`?

plugins/gtkui/actionhandlers.c
Alexey-Yakovenko: @kpcee

You would need to do this instead:
```
extern ddb_gtkui_t plugin;
#define trace(...) { deadbeef->log_detailed (&plugin.plugin, 0, __VA_ARGS__); }
```

and also remove static qualifier from it in gtkui.c

Alexey-Yakovenko: You will probably need to enable logging via `.plugin.flags = DDB_PLUGIN_FLAG_LOGGING`

plugins/gtkui/actionhandlers.c
Alexey-Yakovenko: The question was about why you couldn't use DB_EV_PLAY_NUM unconditionally.

plugins/gtkui/actionhandlers.c
Alexey-Yakovenko: Yes.

plugins/gtkui/actionhandlers.c
kpcee: If we always use DB_EV_PLAY_NUM, the next song will be played automatically below the current one. This can be done certainly, but the selected playback order will be then ignored, as well as a play queue. If you switch off the feature and let the deleted song play, a play queue and the playback order will be respected.

kpcee:
So far so good, I hope I didn't miss anything.

plugins/gtkui/actionhandlers.c
```diff
@@ -46,6 +46,8 @@
 
 extern GtkWidget *mainwin;
 extern DB_functions_t *deadbeef;
+static DB_plugin_t *plugin;
```
Alexey-Yakovenko: This is not the intended way to do it.
The plan is to make logging configurable at some point.
Having such private plugin objects would hurt this.
You need to use plugin variable that's declared in gtkui.c

plugins/gtkui/actionhandlers.c
```diff
@@ -412,41 +461,47 @@ action_delete_from_disk_handler_cb (void *data) {
...
                     delete_and_remove_track (uri, plt, it);
                 }
             }
             deadbeef->pl_item_unref (it);
         }
-        deadbeef->plt_save_config (plt);
     }
+
+    deadbeef->pl_save_all ();
+    deadbeef->pl_item_unref (it_current_song);
```
Alexey-Yakovenko: This needs a null check.

plugins/gtkui/actionhandlers.c
Alexey-Yakovenko: Ah ok. I missed that last part.

plugins/gtkui/actionhandlers.c
```diff
@@ -412,41 +462,49 @@ action_delete_from_disk_handler_cb (void *data) {
...
+    if (it_current_song) {
+        deadbeef->pl_item_unref (it_current_song);
     }
     deadbeef->pl_unlock ();
     deadbeef->plt_unref (plt);
 
     deadbeef->sendmessage (DB_EV_PLAYLISTCHANGED, 0, DDB_PLAYLIST_CHANGE_CONTENT, 0);
+    
+    if (deadbeef->conf_get_int ("gtkui.skip_deleted_songs", 0) 
+        && deadbeef->plt_get_item_idx (plt, it_current_song, PL_MAIN) == -1
```
Alexey-Yakovenko: I think there's still some logical problem here. You're releasing (unref) the playItem and then keep using it. You probably need to move the unref call to be at the very end.

Alexey-Yakovenko:
Looks like all the issues are fixed!
Thanks for your collaboration!

One last note:

your commit messages are not that useful, and don't leave me other choice than to squash&merge your PR as 1 commit.

It would be useful if you start using the common convention of writing commit messages.
You can see how other commit messages are written to get the idea.

PR #2288, opened 2019-12-20
rg scanner: time calculation bug fixed

### LINKED ISSUE #2217: The elapsed time in replaygain scanner isn't correct.
Scan replaygain
The elapsed time is much much longer than real elapsed time.
It has only passed for 1 to 2 hours.
2019-09-16 15-55-53 的螢幕擷圖

Deadbeef version: deadbeef-git
OS: Manjaro

kpcee:
* time calculation fixed  #2217
* info label now shows more detailed info "(%i of %i files)"

### DIFF
--- a/plugins/gtkui/rg.c
+++ b/plugins/gtkui/rg.c
@@ -65,8 +65,8 @@ static void
 _formatTime (float sec, int extraPrecise, char *buf, int bufsize) {
     int hr;
     int min;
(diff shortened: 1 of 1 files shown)


Alexey-Yakovenko:
Oooh right.. This explains why I couldn't repro -- only affected hours calculation.
Thanks for looking into it!

PR #2289, opened 2019-12-21
unmark ddb_playback_state_t as deprecated

huglovefan:
probably didn't mean to deprecate this in df441929c30787b44a7cdb7c12069c8fa2b0bdc1


Alexey-Yakovenko:
Oops. Yes, this is wrong. Thanks!

PR #2299, opened 2019-12-30
gtkui: add "Remove dead items" feature; fixes half of https://github.…

JakeSFR:
…com/DeaDBeeF-Player/deadbeef/issues/504

### DIFF
--- a/plugins/gtkui/actionhandlers.c
+++ b/plugins/gtkui/actionhandlers.c
@@ -698,6 +698,49 @@ action_toggle_designmode_handler (DB_plugin_action_t *act, int ctx) {
 }
 
 gboolean
+action_remove_dead_items_handler_cb (void *data) {
+    deadbeef->pl_lock ();
+
+    ddb_playlist_t *plt = deadbeef->plt_get_curr ();
+    
+    if (plt) {
+        DB_playItem_t *it = deadbeef->pl_get_first (PL_MAIN);
+
+        while (it) {
+            const char *uri = deadbeef->pl_find_meta (it, ":URI");
+            DB_playItem_t *next = deadbeef->pl_get_next (it, PL_MAIN);
+
+            // Remove only local files, not network streams
+            if (deadbeef->is_local_file (uri)) {
+                // check if file exists
+                DB_FILE *fp = deadbeef->fopen (uri);
+                if (!fp) {
+                    //trace("Removing '%s' from playlist\n", uri);
+                    deadbeef->plt_remove_item (plt, it);
+                } else {
+                    deadbeef->fclose (fp);
+                }
+            }
+
+            deadbeef->pl_item_unref (it);
+            it = next;
+        }
+    }
+
+    deadbeef->plt_unref (plt);
+    deadbeef->pl_unlock ();
+    deadbeef->sendmessage (DB_EV_PLAYLISTCHANGED, 0, DDB_PLAYLIST_CHANGE_SELECTION, 0);
+
+    return FALSE;
+}
+
+int
+action_remove_dead_items_handler (DB_plugin_action_t *act, int ctx) {
+    gdk_threads_add_idle (action_remove_dead_items_handler_cb, NULL);
+    return 0;
+}
+
+gboolean
 action_preferences_handler_cb (void *data) {
     gtkui_run_preferences_dlg ();
     return FALSE;
--- a/plugins/gtkui/actionhandlers.h
+++ b/plugins/gtkui/actionhandlers.h
@@ -160,6 +160,12 @@ int
 action_toggle_designmode_handler (DB_plugin_action_t *act, int ctx);
 
 gboolean
+action_remove_dead_items_handler_cb (void *data);
+
+int
+action_remove_dead_items_handler (DB_plugin_action_t *act, int ctx);
+
+gboolean
 action_preferences_handler_cb (void *data);
 
 int
--- a/plugins/gtkui/actions.c
+++ b/plugins/gtkui/actions.c
@@ -148,7 +148,7 @@ add_mainmenu_actions (void)
                     if (prev_title && 0 == strcmp ("File", prev_title))
                         gtk_menu_shell_insert (GTK_MENU_SHELL (current), actionitem, 5);
                     else if (prev_title && 0 == strcmp ("Edit", prev_title))
-                        gtk_menu_shell_insert (GTK_MENU_SHELL (current), actionitem, 7);
+                        gtk_menu_shell_insert (GTK_MENU_SHELL (current), actionitem, 8);
                     else {
                         gtk_container_add (GTK_CONTAINER (current), actionitem);
                     }
--- a/plugins/gtkui/callbacks.c
+++ b/plugins/gtkui/callbacks.c
@@ -98,7 +98,12 @@ on_quit_activate                      (GtkMenuItem     *menuitem,
     gdk_threads_add_idle (action_quit_handler_cb, NULL);
 }
 
-
+void
+on_remove_dead_items_activate           (GtkMenuItem     *menuitem,
+                                        gpointer         user_data)
+{
+    gdk_threads_add_idle (action_remove_dead_items_handler_cb, NULL);
+}
 
 void
 on_select_all1_activate                (GtkMenuItem     *menuitem,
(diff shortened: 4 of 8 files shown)


plugins/gtkui/actionhandlers.c
```diff
@@ -697,6 +697,49 @@ action_toggle_designmode_handler (DB_plugin_action_t *act, int ctx) {
...
+    ddb_playlist_t *plt = deadbeef->plt_get_curr ();
+    
+    if (plt) {
+        DB_playItem_t *it = deadbeef->pl_get_first (PL_MAIN);
+
+        while (it) {
+            const char *uri = deadbeef->pl_find_meta (it, ":URI");
+            DB_playItem_t *next = deadbeef->pl_get_next (it, PL_MAIN);
+
+            // Remove only local files, not network streams
```
JakeSFR: Ouch, I forgot about Audio CD tracks.
I just checked and deadbeef->fopen() doesn't seem to open '/dev/cdrom#1.cda' kind of "files", so even if CD is in the drive, all tracks are being removed from the playlist. That's unforunate...
Frankly, I'm having second thoughts about it - now that there's the log window popping up when a track is unavailable, it's very easy to spot them, unlike in 0.7.2, so maybe there's no need for this feature after all.

Alexey-Yakovenko: Cd audio "file" naming is not implemented in the best way. It should have used URI scheme instead.
So I think you should not attempt to fix it here.

Another matter is that using `fopen` is not guaranteed to work correctly on all platforms for checking whether a file exists (e.g. if a file is temporarily unavailable). It's also very slow. You should use stat(2) or lstat(2) instead.
You'll need to strip the "file://" manually.

plugins/gtkui/actionhandlers.c
```diff
@@ -697,6 +697,49 @@ action_toggle_designmode_handler (DB_plugin_action_t *act, int ctx) {
...
+    deadbeef->plt_unref (plt);
+    deadbeef->pl_unlock ();
+    deadbeef->sendmessage (DB_EV_PLAYLISTCHANGED, 0, DDB_PLAYLIST_CHANGE_SELECTION, 0);
+
+    return FALSE;
+}
+
+int
+action_remove_dead_items_handler (DB_plugin_action_t *act, int ctx) {
+    gdk_threads_add_idle (action_remove_dead_items_handler_cb, NULL);
```
Alexey-Yakovenko: ~Please use `g_idle_add` instead. This is what the rest of the code is using.~

Alexey-Yakovenko: hmm.. nvm, actually in this particular case we HAVE to be using gdk_threads_add_idle because of Ubuntu legacy code in the external menu bar.

plugins/gtkui/actionhandlers.c
```diff
@@ -697,6 +697,49 @@ action_toggle_designmode_handler (DB_plugin_action_t *act, int ctx) {
...
+            }
+
+            deadbeef->pl_item_unref (it);
+            it = next;
+        }
+    }
+
+    deadbeef->plt_unref (plt);
+    deadbeef->pl_unlock ();
+    deadbeef->sendmessage (DB_EV_PLAYLISTCHANGED, 0, DDB_PLAYLIST_CHANGE_SELECTION, 0);
```
Alexey-Yakovenko: `DDB_PLAYLIST_CHANGE_SELECTION` doesn't seem to be a correct constant here.
You're changing playlist content, not just selection, therefore the correct constant is `DDB_PLAYLIST_CHANGE_CONTENT`

Alexey-Yakovenko:
Have you tried to run this on a huge playlist, like 50K items and more?
Is it very slow? I'm wondering if it needs a progress bar.

plugins/gtkui/actionhandlers.c
JakeSFR: > Cd audio "file" naming is not implemented in the best way. It should have used URI scheme instead.
> So I think you should not attempt to fix it here.

Hmm, the best I can think of is simply ignoring such "files" e.g.:

```
const char *ftype = deadbeef->pl_find_meta (it, ":FILETYPE");
...
if (deadbeef->is_local_file (uri) && strcmp(ftype, "cdda") != 0) {
...
```
But either way, it will surely, sooner or later, result in some bug reports from confused users.


> Another matter is that using fopen is not guaranteed to work correctly on all platforms for checking whether a file exists (e.g. if a file is temporarily unavailable). It's also very slow. You should use stat(2) or lstat(2) instead.
> You'll need to strip the "file://" manually.

Yes, 'stat' was what I used in the first version of the patch:

```
if (deadbeef->is_local_file (uri)) {
    // check if file exists
    memset (&buf, 0, sizeof (buf));
    int stat_res = stat (uri, &buf);
    
    if (stat_res != 0) {
        deadbeef->plt_remove_item (plt, it);
    }
}
```
but it obviously didn't work with ZIP/RAR archives.
Then I found about 'fopen' and it does, so I switched to it; didn't know about its limitations.

plugins/gtkui/actionhandlers.c
JakeSFR: Oh, ok, I must've blindly copy/paste it without a thought.

JakeSFR:
> Have you tried to run this on a huge playlist, like 50K items and more?
> Is it very slow? I'm wondering if it needs a progress bar.

Initially tried with ~1000 items and it was instant.
But just now tried with a set of small WAVs and MP3s (due to the lack of enough free space on my disks), which I duplicated up to 63014 files (~5GiB):
- just clicking 'Remove dead items', while the files are still on HDD freezes the playlist for ~1.5 sec.,
- clicking 'Remove dead items' after actually deleting all the files from HDD freezed DB for ~2 sec.

That's on my old Intel-i3 laptop. FS is ext4.
I don't know, maybe OS keeps these files' metadata or something in some buffer in RAM and that's why it's so fast...


Btw, at first I had also some .amr files in the set, but while I was adding them to the playlist, at some point DB stopped with these error msgs:

> ERROR: could not load: /mnt/sda5/ZZZ/249/Kos.amr

Looks like DB, while loading AMRs, still keeps them open:

```
# lsof | grep '\.amr' | wc -l
10139
```

Even removing the playlist doesn't help, only closing DB actually closes them, too.
Should I open a separate issue concerning this?


Anyway, I have to admit that I'm not happy with this PR and (especially if only 'stat' is the way to go, instead of 'fopen', what in addition to AudioCDs exludes also ZIP/RAR archives) I think it's not good enough to be merged. What do you think?
I'm closing it, at least for now...
Well, the bright side of it all is the discovery of the AMR bug, though. ;)

Thanks!

plugins/gtkui/actionhandlers.c
Alexey-Yakovenko: > Hmm, the best I can think of is simply ignoring such "files" e.g.:

You shouldn't do that -- I'd rather change the file naming scheme in the CDDA plugin to make the files be reported as "non-local".

> Then I found about 'fopen' and it does, so I switched to it; didn't know about its limitations.

Yeah, it's not simple. With archives, you would have to use fopen, and it will remain slow. There's no way around it. But with regular files -- stat is more reliable, and faster.

Alexey-Yakovenko:
> Should I open a separate issue concerning this?

Yes, please do. Sounds like a bug.

Alexey-Yakovenko:
Btw you shouldn't close the PR -- you can update it with more commits instead.

Alexey-Yakovenko:
> Anyway, I have to admit that I'm not happy with this PR

If it freezes for only 2 sec on 60K files -- I'm actually happy, even without using stat.
I can also extend deadbeef API with a "fileExists" call, and then this can be implemented in various VFS plugins with speed optimizations.

JakeSFR:
>> Should I open a separate issue concerning this?
> 
> Yes, please do. Sounds like a bug.


Ok, done: https://github.com/DeaDBeeF-Player/deadbeef/issues/2303


> Btw you shouldn't close the PR -- you can update it with more commits instead.

Ok, if/when I get back to it, I'll reopen this one.


> > Anyway, I have to admit that I'm not happy with this PR
> 
> If it freezes for only 2 sec on 60K files -- I'm actually happy, even without using stat.

I checked again, but this time on NTFS (ntfs-3g) with 91653 files and it's no longer that impressive, unfortunately:
- with all files remaining on HDD: ~52 sec. (heavy disk activity),
- again, but after removing all the files from HDD: ~10 sec (no disk activity).

> I can also extend deadbeef API with a "fileExists" call, and then this can be implemented in various VFS plugins with speed optimizations.

Oh, that would be useful!

PR #2302, opened 2020-01-03
rg scanner: updates now identical files in all playlists

kpcee:
The rg scanner will now check all playlists and update the metadata for all identical tracks.


Alexey-Yakovenko:
It's an interesting change, but I disagree that it's a good idea to do it this way.

There are currently many ways in the UI to change metadata of 1 track and not update the same track across the playlist.

You make it "sync" the tracks for the replay gain use case, but not other cases.

To make it right, the system needs to be in place, which would do this automatically in all cases, but for that we would need to have media library since it would link individual tracks in the playlists with the library "master" tracks.

Until this happens, it doesn't make sense to fix individual cases.
This is why I'm rejecting it.

PR #2310, opened 2020-01-16
fix potential buffer overflow in on_group_by_custom_activate

olafhering:
Repeated strncat calls may overflow the buffer, which is what gcc
complains about.  While in practice the buffer might be large enough,
the compiler can not possibly know that.

Rearrange the code to fill the buffer with its size in mind.

Signed-off-by: Olaf Hering <olaf@aepfle.de>

### DIFF
--- a/plugins/gtkui/plcommon.c
+++ b/plugins/gtkui/plcommon.c
@@ -1548,9 +1548,9 @@ on_group_by_custom_activate            (GtkMenuItem     *menuitem,
     DdbListviewGroupFormat *fmt = listview->group_formats;
     while (fmt) {
         if (format[0] != 0) {
-            strncat(format, SUBGROUP_DELIMITER, sizeof(format) - 1);
+            strncat(format, SUBGROUP_DELIMITER, sizeof(format) - strlen(format) - 1);
         }
-        strncat(format, fmt->format, sizeof(format) - 1);
+        strncat(format, fmt->format, sizeof(format) - strlen(format) - 1);
         fmt = fmt->next;
     }
     gtk_entry_set_text (GTK_ENTRY (entry), format);


olafhering:
(Not runtime tested)

Alexey-Yakovenko:
I acknowledge the bug, but I think this could be solved in a much simpler way.

Replacing `sizeof(format) - 1` with `sizeof(format) - strlen(format) - 1` in the original code would do the right thing, and will be only 2 lines of changes.

olafhering:
Would that be more correct?
After doing the PR I thought that perhaps  `g_string_sized_new` and `g_string_append` should be used.

Alexey-Yakovenko:
> Would that be more correct?

I think less code is better in this case, especially since there are no unit tests for that.

> After doing the PR I thought that perhaps g_string_sized_new and g_string_append should be used.

Using glib would diminish reuse value of the code, and using g_string would introduce many unnecessary heap allocations and deallocations, which may cause significant performance problems in this particular use case. I don't recommend that.

olafhering:
That `GString tmpstr` would go into `DdbListview`, which gets allocated once in `ddb_listview_init` and released in `ddb_listview_destroy`.

Alexey-Yakovenko:
Then I don't understand what you intend to do with GString. If you want to go this route, and the work is not too complex -- it might make sense to just do it and make another PR. But I still think it's the easiest to keep the original code, and fix the 2 wrong lines.

olafhering:
I have updated the change, not runtime tested.

Alexey-Yakovenko:
OK now I see what you meant. I'm sorry, but I disagree with this solution.
It doesn't make much sense to me the way it's done. It's putting a local variable of one specific method into the object. It's not like the code wouldn't work -- it will.. It just doesn't make sense from design standpoint.

olafhering:
Reverted.

Alexey-Yakovenko:
Thanks for your input, and reporting the issue. I applied basically the same fix, just in a slightly different way.

Issue #504, opened 2015-03-12
Remove dead and duplicate items feature


Oleksiy-Yakovenko:
Original [issue 602](https://code.google.com/p/ddb/issues/detail?id=602) created by Alexey-Yakovenko on 2012-04-19T11:49:43.000Z:

submitted by: starws
Date: 2010-05-11 18:47:14

&quot;Remove dead items&quot; and &quot;Remove duplicate items&quot; features for active playlist

wdlkmpx:
hmm this should be very easy to implement in the gtkui. 

Looking at some foobar2000 screenshots I see:
http://news.softpedia.com/images/extra/WINDOWS3/large/foobar2000/foobar2000 edit menu.png
Menu Edit
-> Remove duplicates
-> Remove dead items

Remove duplicates = coreutils 'uniq' (with absolute filenames??)
Remove dead items = verify that all files in the playlist actually exist, just one operation i suppose

When I close deadbeef and i open it again... renamed/deleted files are still in the playlist, would be really to just wipe all the offending items at the same time.. "Remove dead items"

--

Right now the right click menu of a playlist item (in deadbeef) shows "Remove" and "Remove from disk..."

Implementing this way would be really cool..
http://news.softpedia.com/images/extra/WINDOWS3/large/foobar2000/foobar2000 file operations context menu.png

yssmcl:
Hi! Any update on this?

DaarkWel:
Yes, it would be nice. This function existed even in olde antique Winamp :-)

JakeSFR:
Yeah, I agree that especially "Remove dead items" would be a useful feature.

So, I was dabbling with this idea a little today and here's my attempt to implement it.
Not knowing DB's codebase and not being very proficient in C, it's just a simple "copy/paste/adjust" type of work, but, as far as I can tell after a few tests, it seems to work fine and might be at least something to start with.

Patch (against https://github.com/DeaDBeeF-Player/deadbeef/commit/fe514eeb3346ff86bd4866038cc822a8a32968db revision): [db-remove-dead-items.patch.gz](https://github.com/DeaDBeeF-Player/deadbeef/files/4008204/db-remove-dead-items.patch.gz)

Btw, one thing that bugs me is that I'd like to see this option between "Sort by" and "Configure Custom Shell Commands", but couldn't figure it out...

Alexey-Yakovenko:
[...]

PR #2311, opened 2020-01-16
Ignore duplicate items

kpcee:
This patch adds a new entry "Ignore Duplicates" to the playlist context menu. It prevents that the user can add files twice, the setting have to set for each playlist separately. Possible partial solution for feature request #504. 

Bugfix: Changes to the "Enable Autosort" setting were not always saved, because a plt_modified (plt); was missing.

### DIFF
--- a/playlist.c
+++ b/playlist.c
@@ -105,6 +105,7 @@ static int playlists_count = 0;
 static playlist_t *playlists_head = NULL;
 static playlist_t *playlist = NULL; // current playlist
 static int plt_loading = 0; // disable sending event about playlist switch, config regen, etc
+static int ignore_duplicates = 0;
 
 #if !DISABLE_LOCKING
 static uintptr_t mutex;
@@ -975,6 +976,16 @@ plt_insert_file_int (int visibility, playlist_t *playlist, playItem_t *after, co
     // now that it's known we're not dealing with URL, check if it's a relative path
     if (is_relative_path (fname)) {
         return NULL;
+    }
+
+    // Skip double tracks
+    if (ignore_duplicates) {
+        for (playItem_t *it = playlist->head[PL_MAIN]; it; it = it->next[PL_MAIN]) {
+            const char *uri = pl_find_meta (it, ":URI");
+            if (strcmp (uri, fname) == 0) {
+                return NULL;
+            }
+        }
     }
 
     // detect decoder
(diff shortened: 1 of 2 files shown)


Alexey-Yakovenko:
The playlist context menu is a wrong place for this option.
It needs to be in the settings, or perhaps in the file menu, but not in the context menu.

Alexey-Yakovenko:
Although, if it's a playlist-specific option, this might be viable, but I'm still not sure it's a good idea. It's possibly even better to add a checkbox to the file dialog.

kpcee:
I like it the way it is now, but of course it's not my decision. If it is, I would add it in the settings and it would apply to all playlists. To round things up, we could add a "Remove Duplicate Items" in the Edit menu, in addition to the PR #2299.

Alexey-Yakovenko:
There are a few things that needed to be discussed before starting to work on the feature.
It is also written in the [contributing guide](https://github.com/DeaDBeeF-Player/deadbeef/blob/master/CONTRIBUTING.md).

Not discussing design leads to big code changes, because redesign will be needed.

For me, it's pointless to review the code implementing the feature, since it isn't backed by any design.

* I disagree that the option needs to be named "Ignore duplicates" -- it's not what the feature is doing
* I disagree that it's put in the playlist tab context menu -- it needs to be discussed and decided where is the best place
* I disagree with making this option a global variable. It needs to be some sort of a parameter, maybe a flag, to the function which adds the files. It needs to be testable in some way, and it needs to be configurable without going through the config & the global variable.

kpcee:
> For me, it's pointless to review the code implementing the feature, since it isn't backed by any design.

I think the design fits perfectly, as it fits perfectly with the Autosort feature. If you would move "Ignore Duplicates" to the main menu, you would distribute related functions across the user interface.

> * I disagree that the option needs to be named "Ignore duplicates" -- it's not what the feature is doing

This feature prevents a file from being added to the playlist twice, whether this is called "(Ignore|Do Not Add|Filter Out|Disallow) Dublicates" or similar is a matter of taste.

I don't want to argue, especially over something like this. Since the feature was requested several times, I took a look at it. No more, no less.

Alexey-Yakovenko:
This attitude is not going to help anything.
I don't mind other people working on features, but this work requires some coordination and discussion before jumping to code.

I need to consider more things -- like how easy it is to discover and use, what problems it may cause in the future, how to reuse code and how to fit the design on other platforms, and so on.

PR #2315, opened 2020-01-21
hotkeys: don't fire global hotkeys with the system typematic

abextm:
If an action has the flag DB_ACTION_HOTKEY_REPEAT it will be called
at a regular interval, regardless of the system typematic setting. If
it doesn't have DB_ACTION_HOTKEY_REPEAT it will only be called once per
key down/up cycle

### DIFF
--- a/deadbeef.h
+++ b/deadbeef.h
@@ -1555,7 +1555,12 @@ enum {
 
 #if (DDB_API_LEVEL >= 10)
     // Don't allow running this action in playlist context, even if it supports multiple selection
-    DB_ACTION_EXCLUDE_FROM_CTX_PLAYLIST = 1 << 7
+    DB_ACTION_EXCLUDE_FROM_CTX_PLAYLIST = 1 << 7,
+#endif
+
+#if (DDB_API_LEVEL >= 11)
+    // If this has a hotkey assigned and it is held down, the callback should be called at a regular interval
+    DB_ACTION_HOTKEY_REPEAT = 1 << 8
 #endif
 };
 
--- a/plugins/hotkeys/Makefile.am
+++ b/plugins/hotkeys/Makefile.am
@@ -1,7 +1,7 @@
 if HAVE_HOTKEYS
 pkglib_LTLIBRARIES = hotkeys.la
 hotkeys_la_SOURCES = hotkeys.c hotkeys.h actionhandlers.c actionhandlers.h
-hotkeys_la_LDFLAGS = -module -avoid-version
+hotkeys_la_LDFLAGS = -lrt -module -avoid-version
 
 EXTRA_hotkeys_la_SOURCES = keysyms.inc
 
--- a/plugins/hotkeys/hotkeys.c
+++ b/plugins/hotkeys/hotkeys.c
@@ -30,6 +30,8 @@
 #endif
 #ifndef NO_XLIB_H
 #include <X11/Xlib.h>
+#include <X11/XKBlib.h>
+#include <time.h>
 #endif
 #include <ctype.h>
 #ifdef __linux__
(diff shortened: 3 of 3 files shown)


kuba160:
I tested the changes and it looks like on my system `DB_ACTION_HOTKEY_REPEAT` is not working as you described (or maybe I misunderstood something). I removed `DB_ACTION_HOTKEY_REPEAT` from volume up/down action and volume changed when I hold the hotkey (although maybe at slower rate).

abextm:
What exactly was happening? It only changed once per key down/up, or did it not change the behavior at all? Note that this change only affects global hotkeys.

kuba160:
Ok then, I wasn't testing the behavior on global hotkeys, I will check soon how it behaves on global hotkeys.

kuba160:
Ok, now I see. I found another problem: I tried to use 2 global hotkeys at once (Volume up and down) and sometimes it can lock on one hotkey. From what I've seen problem didn't exist before. If you need more info for debug please say, I didn't look through code good enough.

abextm:
I can't get that to happen, can you turn on trace and give me logs?

abextm:
Nevermind, I got it to break right after posting that message; It seems that X is just not sending a KeyRelease event.

abextm:
Turns out Xorg follows some twisted logic which makes it not send the second KeyRelease unless you XGrabKeyboard manually, so I do that now and it seems to work.

Alexey-Yakovenko:
@abextm what was the behavior before that? did the keys repeat always, or never?

abextm:
Before this it would not repeat until typematic kicks in, then it would repeat at the system typematic rate, which for me is much faster than the hotkey poll rate, making it difficult to control the volume.

Alexey-Yakovenko:
So why do you think it's a good idea to change the normal system behavior? I just tried global hotkeys from the system, and they work the same way as normal keys, with this extra delay -- so that the key can be pressed either once, or be hold for autorepeat.

Although I understand the issue you're trying to solve, and it can be valuable to look into it.
I mean, it shouldn't be much more than reducing the sleep amount?

abextm:
I'm basically trying to have the same behavior I am used to with foobar2000. For volume keys especially, its of very limited use to adjust a single tick up/down. I typically want to move it atleast 3 or 4 steps, which this allows me to do much faster, without having to change my system typematic speed which is already where I want it for basically every other use.

Alexey-Yakovenko:
> For volume keys especially, its of very limited use to adjust a single tick up/down.

I tend to disagree, since this is the only way I control the volume.

> I typically want to move it atleast 3 or 4 steps, which this allows me to do much faster, without having to change my system typematic speed which is already where I want it for basically every other use.

If there's such use case -- we need to think how to support it without affecting other use cases.
There doesn't seem to be one size fits all.

You seem to believe that there are specific hotkeys that need to be "continuous" without typematic delay, but that is subjective, and doesn't mean all users will want it (I won't).

I can see how that can be useful to have an option per-hotkey, but I don't like to just change it globally for all users for specific subjective set of hotkeys.

abextm:
I will look into making it user configurable then

Alexey-Yakovenko:
My input:

- You should make it configurable per-hotkey.
- The hotkey config is not very flexible, but the additional "NoDelay" flag can be added as a modifier, in a backwards compatible way.
- It seems that this will also be forward compatible, and won't need migration - but needs to be tested, that previous versions can load the config with this new flag. I think it will be ignored with the current code. Otherwise we might need to find a different way.

Alexey-Yakovenko:
More input: I think it's fine to decrease the sleep amount without adding an option for that.
Even better if you can find a way to wait for events, instead of polling. I'm not very good at Xlib.

Alexey-Yakovenko:
Also, although I think I made this clear enough, the repeat delay flag should be a property of hotkey configuration, and not affect actions themselves.

Alexey-Yakovenko:
@abextm 
Closing the PR for now, since it doesn't seem any progress is being made.
Please let me know if there are any news.

PR #2318, opened 2020-01-27
Illumos/SVR4 build fixes

EubiquitouS:
Tested on OpenIndiana 2019.10, been running DeaDBeeF for about an hour and everything seems pretty stable. 

## Workarounds
`CFLAGS=-D__EXTENSIONS__ -DHAVE_ALLOCA_H` during configure phase. From my understanding, autoconf is supposed to detect those and set up the macros automatically, but it doesn't. I have no idea about how autoconf works/is set up/whatever, but I might try taking a look at this later.

## What doesn't build
- **ffmpeg module:** Builds, but OpenIndana's ffmpeg package doesn't include `libav*.so`, so linking fails. Might test again with a custom ffmpeg build when I have time.
- **sc68 module:** Build fails because the definition of getopt on `dial_conf.c` conflicts with the one on `stdlib.h`.


playlist.c
```diff
@@ -1201,8 +1201,8 @@ plt_insert_dir_int (int visibility, playlist_t *playlist, DB_vfs_t *vfs, playIte
     if (vfs && vfs->scandir) {
         n = vfs->scandir (dirname, &namelist, NULL, dirent_alphasort);
         // we can't rely on vfs plugins to set d_type
-        // windows: missing dirent[]->d_type
-        #ifndef __MINGW32__
+        // windows/svr4 unixes: missing dirent[]->d_type
+        #if !(defined(__MINGW32__) || defined(__SVR4))
```
Alexey-Yakovenko: This looks wrong. Please double check if this is what you wanted.

Alexey-Yakovenko:
> From my understanding, autoconf is supposed to detect those and set up the macros automatically, but it doesn't

Yes, there's already a check.

```
AC_CHECK_HEADER([alloca.h],[],[alloca.h not found.])
```

It looks incomplete though.
It needs `HAVE_ALLOCA_H`=yes in the empty `[]`, then it will probably work.

Alexey-Yakovenko:
Actually, this would not be enough.
This looks like what you would need:

```
AC_CHECK_HEADER([alloca.h],[AC_DEFINE(HAVE_LIBICONV,1)],[alloca.h not found.])
```

playlist.c
EubiquitouS: It is right; `!(A || B) == !A && !B` (or I'm not thinking right). Should be more readable now, anyway, I wrote this kinda late in the night after trying to get a bunch of stuff to work, sorry for that.

EubiquitouS:
> Actually, this would not be enough.
> This looks like what you would need:
> 
> ```
> AC_CHECK_HEADER([alloca.h],[AC_DEFINE(HAVE_ALLOCA_H,1)],[alloca.h not found.])
> ```

Yup, this works, though autoconf apparently errors out if you don't pass a description for some reason (judging from [this](https://www.gnu.org/software/autoconf/manual/autoconf-2.66/html_node/Defining-Symbols.html), it looks like the description is optional, but it errors out anyway).

EubiquitouS:
Also, it seems like `AC_USE_SYSTEM_EXTENSIONS` already defines `__EXTENSIONS__`, but on `config.h` rather than "globally", so I just built with `-Werror=implicit-function-declaration` and (somewhat) blindly added/moved a import to `config.h` as necessary (which included some files apparently imported from other projects). Think it would be a better idea to define flags like those on CFLAGS instead, but I haven't found a way to do it (moving `AC_USE_SYSTEM_EXTENSIONS` before `AC_CONFIG_HEADER()` on configure.ac changes nothing, and information I've found haven't been of much use)

Alexey-Yakovenko:
Why did you need to include config.h in so many files? I don't think this is needed.

Alexey-Yakovenko:
I will try to find a way to transfer the AC_DEFINE values set by AC_USE_SYSTEM_EXTENSIONS to CFLAGS -D. This is actually needed for -D_GNU_SOURCE too.

Alexey-Yakovenko:
I got a few ideas how to do that, gonna try them today.
Also, it's a really great idea to add `-Werror=implicit-function-declaration`, I'm suffering from this a lot, since these problems often occur with the older libraries used for making CI builds.

Alexey-Yakovenko:
Sorry for the delay, I needed to catch up with a few other things.
I'll follow up on this as soon as I can. Soon.

Alexey-Yakovenko:
After digging into the internals of autoconf, I figured that AC_DEFINE simply appends a #define string to the config header variable, and there's no way to extract any information back from there.

The `AC_USE_SYSTEM_EXTENSIONS` internally tests for the `ac_cv_safe_to_define___extensions__=yes`, and then AC_DEFINES __EXTENSIONS__.

It's not portable, but let's see if that approach would work on your host.

The `_GNU_SOURCE` is AC_DEFINED unconditionally, so we can just have it in CFLAGS.

Please try with latest master branch, without all the config.h includes.
Commit b3478bbc8b8606c3c2ce73167a4d10b8fc640cb5

Alexey-Yakovenko:
It may help to add this line somewhere in configure.ac to see what CFLAGS you will get:
`echo "CFLAGS: $CFLAGS"`

Remember to rerun `autogen.sh`

EubiquitouS:
Merged changes from master and reverted most includes, `-D__EXTENSIONS` gets into `CFLAGS` and builds/runs fine on my end.

plugins/dumb/dumb-kode54/src/helpers/lpc.c
```diff
@@ -43,6 +43,10 @@ Carsten Bormann
 
 *********************************************************************/
 
+#include "../../../../config.h"
```
Alexey-Yakovenko: Build fails on mac, since there's no config.h
You need to add #ifdef HAVE_CONFIG_H everywhere.

EubiquitouS: Completely forgot that, sorry. Should be fixed now.

Alexey-Yakovenko:
Thanks for cooperation and the patience! :)

PR #2352, opened 2020-03-05
fix syntax error in deadbeef.h

huglovefan:
line 1601 ends an enum declaration but it's missing the semicolon

```
% clang -xc -DDDB_API_LEVEL=10 /usr/include/deadbeef/deadbeef.h
/usr/include/deadbeef/deadbeef.h:1601:2: error: expected ';' after enum
}
 ^
 ;
```


Alexey-Yakovenko:
Thanks!

PR #2363, opened 2020-04-02
Add soundtouch to premake / windows_postbuild fix

kuba160:
Comments on soundtouch plugin:
- Music is choppy when soundtouch added first to dsp. Restarting deadbeef does not help. Enabling and disabling "Use Quickseek" fixes it for some reason. This has to be done every time soundtouch is added to dsp.
- Setting tempo over ±10% starts making music unstable/choppy. I'm not familiar with soundtouch so I'm not sure if it's fault of the this plugin, output plugin or soundtouch library limitation
- In gtk + and - buttons  on pitch change give 0.0000001 difference, which is kind of slow


Alexey-Yakovenko:
Certain settings cause sound stuttering for me to. Decrease the Time stretch sequence length should help.

Alexey-Yakovenko:
> In gtk + and - buttons on pitch change give 0.0000001 difference, which is kind of slow

I'm intending to change them to sliders.

Alexey-Yakovenko:
> I'm intending to change them to sliders.

Just to share the intentions:

The current implementation of config dialogs in GTKUI is very dated.
It's using the ok/apply/cancel pattern, so interactive changes are not possible.

On Mac there's a more modern approach, where there's no OK or Apply buttons, and the changes are applied immediately.

This relies on the new backend called "Scriptable", that is cross platform (written in C).

So to make use of all this, the current GTKUI code handling config dialogs needs to be converted to use Scriptable, and then the UI can be updated to apply changes immediately, support "Restore defaults", and make other UI improvements.

But right now, even making a change to use sliders instead of spin buttons is hard.
This breaks certain use cases, since GtkScale doesn't have a direct-input text field,
and it's quite tricky to add it without first converting the code to use Scriptable backend.

So it will take some work.

PR #2376, opened 2020-04-27
alac: fix for no decoder config

kuba160:
Some mp4 files cause Segmentation Fault without this check. Turns out that `mp4ff_get_decoder_config()` succeeds every time request is valid, but only sets buff and buff_size if these values are known.


Alexey-Yakovenko:
I think it's better to fix the issue in the mp4ff_get_decoder_config itself, since it's not supposed to return 0 and not set the values.

kuba160:
Done.

PR #2395, opened 2020-06-01
main: fix hang when no gui available

kuba160:


Alexey-Yakovenko:
hmm wait.. I think what you see as hanging is an intended behavior? it just runs in the console, you can control playback via command line from another console etc. Why do you need it to quit when there's no GUI plugin?

kuba160:
Ah, sorry, I forgot about that use case. I just noticed this behaviour when I didn't send `DB_EV_TERMINATE` signal from gui plugin. Thought it was not intended.

PR #2419, opened 2020-08-01
Windows branch merge

### LINKED ISSUE #2413: Windows + UNC paths
Hi - not sure if issue or feature missing.
But since I expected it to work - rather an issue.

Drag and drop a file or folder from a UNC network share into the list. 
OR
Select a file via the add file menu and browse to a UNC path.

Result: nothing gets added
Win has a security mechanism in place, preventing from eg. copying UNC files to the desktop: 
I already took steps to rule out the problem of windows security. (add unc to intranet manually)

I then added a Network Drive (eg. Z:\) mounting said UNC path, and here goes, all functions properly. 
So there must be a different file handling, and not done via OS calls (not a developer, just trying my best to describe here)

Tl;dr: playing via win network paths do not work, except when adding a network drive with drive letter. 

Deadbeef version: latest devel:  SHA1: a18c62cab9dd5a75f197cbab781fddaa3234362d
OS: Win 10 latest.

kuba160:
Changes:
- junklib: fix when editing tags on windows
- fix for adding playlists
- fix for adding files through unc paths (fixes #2413)
- artwork: various porting fixes. artwork will also be included in next build 
- ctx type fix for hotkeys and gtkui
- adplug for premake
- rewrite of scandir
- fix for failing windows travis build (I hope?) If it doesn't fix add `ps` command and see what else is running in background


junklib.c
```diff
@@ -4668,7 +4668,12 @@ junk_rewrite_tags (playItem_t *it, uint32_t junk_flags, int id3v2_version, const
     if (stat(fname, &stat_struct) != 0) {
         stat_struct.st_mode = 00640;
     }
-    out = open (tmppath, O_CREAT | O_LARGEFILE | O_WRONLY, stat_struct.st_mode);
+#ifdef __MINGW32__
```
Alexey-Yakovenko, line 4671: I suggest to do this instead:

```
#ifndef _O_BINARY
#define _O_BINARY 0
#endif
```

and then use it directly, instead of ifdef in the function bodies.

plugins/converter/converter.c
```diff
@@ -53,6 +53,12 @@ static ddb_converter_t plugin;
 #define trace(...) { deadbeef->log_detailed (&plugin.misc.plugin, 0, __VA_ARGS__); }
 #define trace_err(...) { deadbeef->log_detailed (&plugin.misc.plugin, DDB_LOG_LAYER_DEFAULT, __VA_ARGS__); }
 
+#ifdef __MINGW32__
+#define OFLAGS O_LARGEFILE | O_WRONLY | O_CREAT | O_TRUNC | _O_BINARY
```
Alexey-Yakovenko, line 57: Same as the other comment

kuba160: Done

PR #2421, opened 2020-08-01
artwork small fixes

kuba160:
- fix for double `deadbeef` in cache path
- move slash conversion to main program
- remove path generation from `HOMEDIR`, it is already done for `DDB_SYS_DIR_CACHE` in main

### DIFF
--- a/plugins/artwork-legacy/cache.c
+++ b/plugins/artwork-legacy/cache.c
@@ -79,7 +79,7 @@ int make_cache_root_path (char *path, const size_t size)
     #endif
 
     const char *cache_root = xdg_cache ? xdg_cache : getenv (HOMEDIR);
-    if (snprintf (path, size, xdg_cache ? "%s/deadbeef/" : "%s/.cache/deadbeef/", cache_root) >= size) {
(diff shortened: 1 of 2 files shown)


plugins/artwork-legacy/cache.c
```diff
@@ -61,28 +61,15 @@ void cache_unlock (void)
...
-            slash_p++;
-        }
-        xdg_cache = xdg_cache_conv;
-    }
-    #endif
+    const char *ddb_cache = deadbeef->get_system_dir (DDB_SYS_DIR_CACHE);
 
-    const char *cache_root = xdg_cache ? xdg_cache : getenv (HOMEDIR);
-    if (snprintf (path, size, xdg_cache ? "%s/deadbeef/" : "%s/.cache/deadbeef/", cache_root) >= size) {
+    if (strlen(ddb_cache) >= size) {
```
Alexey-Yakovenko, line 66: This approach is better than truncating -- since truncated path doesn't have any value.
But please update the trace message to say that the path is too long instead.
Truncation won't be occurring anymore.

kuba160: Done

PR #2422, opened 2020-08-02
artwork: more updates

kuba160:
- artwork: fix cache path generation for some builds
- windows: include and use certs for vfs_curl (fixes last.fm cover loading for artwork plugin)

### DIFF
--- a/plugins/artwork-legacy/cache.c
+++ b/plugins/artwork-legacy/cache.c
@@ -62,23 +62,11 @@ void cache_unlock (void)
 int make_cache_root_path (char *path, const size_t size)
 {
     const char *ddb_cache = deadbeef->get_system_dir (DDB_SYS_DIR_CACHE);
-    size_t cachepath_len = strlen(ddb_cache);
 
-    // append '/' later if missing
-    if (ddb_cache[cachepath_len-1] != '/') {
-        cachepath_len++;
-    }
-
-    if (cachepath_len >= size) {
+    int ret = snprintf (path, size, (ddb_cache[strlen(ddb_cache)-1] == '/') ? "%s" : "%s/", ddb_cache);
+    if (ret >= 0 && ret > size) {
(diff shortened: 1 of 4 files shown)


plugins/artwork-legacy/cache.c
```diff
@@ -62,14 +62,25 @@ void cache_unlock (void)
 int make_cache_root_path (char *path, const size_t size)
 {
     const char *ddb_cache = deadbeef->get_system_dir (DDB_SYS_DIR_CACHE);
+    size_t cachepath_len = strlen(ddb_cache);
 
-    if (strlen(ddb_cache) >= size) {
+    // append '/' later if missing
+    if (ddb_cache[cachepath_len-1] != '/') {
```
Alexey-Yakovenko, line 68: You should replace most of this code with a single `snprintf`.
There's no need to check if the `/` is already present -- double `//` is not a problem.

plugins/vfs_curl/vfs_curl.c
```diff
@@ -660,6 +660,9 @@ http_thread_func (void *ctx) {
         headers = curl_slist_append (headers, "Icy-Metadata:1");
         curl_easy_setopt (curl, CURLOPT_HTTPHEADER, headers);
         curl_easy_setopt (curl, CURLOPT_HTTP200ALIASES, ok_aliases);
+#ifdef __MINGW32__
+        curl_easy_setopt (curl,CURLOPT_CAINFO, getenv("CURL_CA_BUNDLE"));
```
Alexey-Yakovenko, line 664: nice.. you figured out how to make a curl build with SSL... I need to do the same for linux smh..
👍

plugins/artwork-legacy/cache.c
kuba160: I converted code to use `snprintf` while still avoiding double `//`.

PR #2441, opened 2020-09-07
Linux: Move files to trash

kpcee:
Instead of deleting files directly the files are moved to the recycle bin with [GIO](https://developer.gnome.org/gio/stable/gio.html) (if available, but should be present in all GTK based distributions).

### DIFF
--- a/plugins/gtkui/actionhandlers.c
+++ b/plugins/gtkui/actionhandlers.c
@@ -41,6 +41,7 @@
 #include "callbacks.h"
 #include <sys/stat.h>
 #include "gtkui_api.h"
+#include <gio/gio.h>
 
 // disable custom title function, until we have new title formatting (0.7)
 #define DISABLE_CUSTOM_TITLE
@@ -385,7 +386,14 @@ _remove_file_from_all_playlists (const char *search_uri) {
 
 static void
 _delete_and_remove_track_from_all_playlists (const char *uri, ddb_playlist_t *plt, ddb_playItem_t *it) {
-    (void)unlink (uri);
+
+    if (deadbeef->conf_get_int ("gtkui.move_to_trash", 1)) {
+        GFile *file = g_file_new_for_path (uri);
+        g_file_trash (file, NULL, NULL);
+        g_object_unref (file);
+    } else {
+        (void)unlink (uri);
+    }
 
     // check if file exists
     struct stat buf;
@@ -409,7 +417,9 @@ action_delete_from_disk_handler_cb (void *data) {
     ddb_action_context_t ctx = (int)(intptr_t)data;
     if (deadbeef->conf_get_int ("gtkui.delete_files_ask", 1)) {
         char buf[1000];
-        const char *buf2 = _(" The files will be lost.\n\n(This dialog can be turned off in GTKUI plugin settings)");
+        const char *buf2 = deadbeef->conf_get_int ("gtkui.move_to_trash", 1) ?
+        _(" The files will be moved to trash.\n\n(This dialog can be turned off in GTKUI plugin settings)") : 
+        _(" The files will be lost.\n\n(This dialog can be turned off in GTKUI plugin settings)");
 
         if (ctx == DDB_ACTION_CTX_SELECTION) {
             int selected_files = deadbeef->pl_getselcount ();
--- a/plugins/gtkui/deadbeef.glade
+++ b/plugins/gtkui/deadbeef.glade
@@ -3775,6 +3775,26 @@ Only prevent clipping</property>
 		  </child>
 
 		  <child>
+		    <widget class="GtkCheckButton" id="move_to_trash">
+		      <property name="visible">True</property>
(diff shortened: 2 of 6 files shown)


Alexey-Yakovenko:
1. It's a nice feature, but I don't like that it's using the command line utility instead of directly calling GIO API.
2. Also not quite sure why it has to be limited to Linux. To my best understanding, it should work on any system where GIO is used, and GTKUI works on BSD and Mac systems.
3. I'm not sure whether it's good or bad idea, but something tells me there will be users who would prefer to delete directly without Trash, e.g. the people who are not using DEs like GNOME or KDE.. there's no UI for Trash access/cleaning. So it seems reasonable to provide an option to delete the files directly. Although having Trash enabled by default is good.

plugins/gtkui/actionhandlers.c
```diff
@@ -385,7 +385,20 @@ _remove_file_from_all_playlists (const char *search_uri) {
 
 static void
 _delete_and_remove_track_from_all_playlists (const char *uri, ddb_playlist_t *plt, ddb_playItem_t *it) {
+
+#ifdef __linux__
+    struct stat buffer;
+    int exist = stat("/usr/bin/gio", &buffer);
+    if (exist == 0) {
+        char command[2000];
+        snprintf (command, sizeof (command), "/usr/bin/gio trash \"%s\"", uri);
```
Alexey-Yakovenko, line 394: Aside from other comments, using double quotes in this case is unsafe, since the filename can contain a double quote in its name, which would open a possibility of arbitrary code execution in certain scenario.

kpcee:
> 1. It's a nice feature, but I don't like that it's using the command line utility instead of directly calling GIO API.

Using the API is much easier than expected and the changes are minimal:
```
GFile *file = g_file_new_for_path (uri);
GError *error = NULL;
g_file_trash (file, NULL, &error);
```
> 2. Also not quite sure why it has to be limited to Linux. To my best understanding, it should work on any system where GIO is used, and GTKUI works on BSD and Mac systems (edit: and Windows!).

If it really works on all platforms, an entry in the settings would certainly make sense. I will revise the PR.

Alexey-Yakovenko:
nice!

kpcee:
Okay updated.

plugins/gtkui/deadbeef.glade
```diff
@@ -3774,6 +3774,26 @@ Only prevent clipping</property>
 		    </packing>
 		  </child>
 
+		  <child>
+		    <widget class="GtkCheckButton" id="move_to_trash">
+		      <property name="visible">True</property>
+		      <property name="can_focus">True</property>
+		      <property name="label" translatable="yes">Move files to trash</property>
```
Alexey-Yakovenko, line 3781: A better name for this is: "Delete from disk should use Trash"
Otherwise it's completely unclear what this option refers to.

plugins/gtkui/actionhandlers.c
```diff
@@ -409,7 +417,7 @@ action_delete_from_disk_handler_cb (void *data) {
     ddb_action_context_t ctx = (int)(intptr_t)data;
     if (deadbeef->conf_get_int ("gtkui.delete_files_ask", 1)) {
         char buf[1000];
-        const char *buf2 = _(" The files will be lost.\n\n(This dialog can be turned off in GTKUI plugin settings)");
+        const char *buf2 = deadbeef->conf_get_int ("gtkui.move_to_trash", 1) == 1 ? _(" The file(s) will be moved to trash.\n\n(This dialog can be turned off in GTKUI plugin settings)") : _(" The file(s) will be lost.\n\n(This dialog can be turned off in GTKUI plugin settings)");
```
Alexey-Yakovenko, line 420: I don't like "file(s)".
It's ok to use "files" even in singular form when there's no specific count displayed.

Alexey-Yakovenko (changes requested):

plugins/gtkui/actionhandlers.c
Alexey-Yakovenko, line 420: `== 1` is unnecessary. It's better to test for "nonzero".
If you prefer explicit equality test, it should be `!= 0`.

Alexey-Yakovenko, line 420: It would be great for readability to split this into 3 lines:
```
const char *buf2 = deadbeef->conf_get_int ("gtkui.move_to_trash", 1) == 1 ?
 _(" The file(s) will be moved to trash.\n\n(This dialog can be turned off in GTKUI plugin settings)") :
 _(" The file(s) will be lost.\n\n(This dialog can be turned off in GTKUI plugin settings)");
```

Alexey-Yakovenko (changes requested):

kpcee:
Okay done.

Alexey-Yakovenko:
Great. Merging.

Alexey-Yakovenko:
Great job, thanks for cooperating :)

kpcee:
Thanks for the kind words.

PR #2445, opened 2020-09-14
medialib: get medialib building and running on Linux

ToadKing:
image

Some changes to the autoconf scripts needed to be made and some Jannson build flags needed to be added to the plugin Makefile.

I have to note that I have to explicitly enable the plugin with `--enable-medialib` despite it looking like it should be detected automatically in the autoconf script.


Alexey-Yakovenko:
Thanks!

PR #2467, opened 2020-10-18
Prevent leaking file descriptors when loading and playing ALAC files

tomjepp:
This fixes the underlying issue causing the crash in https://github.com/DeaDBeeF-Player/deadbeef/issues/2466.

Also fixes a second leak that happened on each ALAC track playback.

### DIFF
--- a/plugins/alac/alac_plugin.c
+++ b/plugins/alac/alac_plugin.c
@@ -111,8 +111,7 @@ alacplug_init (DB_fileinfo_t *_info, DB_playItem_t *it) {
     int bps = 0;
     float duration = 0;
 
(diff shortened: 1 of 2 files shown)


plugins/alac/alac_plugin.c
```diff
@@ -111,8 +111,7 @@ alacplug_init (DB_fileinfo_t *_info, DB_playItem_t *it) {
     int bps = 0;
     float duration = 0;
 
-    DB_FILE *file = deadbeef->fopen (deadbeef->pl_find_meta (it, ":URI"));
```
Alexey-Yakovenko, line 114: Uhh nice find 👍

PR #2489, opened 2020-12-05
gtkui: Update titlebar based on TF update interval

### LINKED ISSUE #2448: %playback_time% does not refresh when in window title.
1. Configure title bar formatting as e.g. : **%playback_time% - %artist% - %title% - DeaDBeeF**
2. Play some file.
3. Wait, do not click anything and observe the window title bar.
4. Interact with playlist, like clicking already selected file (except switching between two playlists) or open preferences dialog.

When at point 3, the title bar does not change showing 0:00 constantly while playing. It refreshes, and once, only when playlist is interacted with or configuration window is called.
Should be refreshed more or less once per second.

Deadbeef version: 1.8.3
OS: Debian 10.5 Stable

Thanks in advance.
M.

saivert:
Fixes #2448

### DIFF
--- a/plugins/gtkui/gtkui.c
+++ b/plugins/gtkui/gtkui.c
@@ -424,6 +424,12 @@ gtkui_titlebar_tf_init (void) {
     statusbar_stopped_bc = deadbeef->tf_compile (statusbar_stopped_tf);
 }
 
+static gboolean
+set_title_cb (gpointer data) {
+    gtkui_set_titlebar (NULL);
+    return FALSE;
+}
+
 void
 gtkui_set_titlebar (DB_playItem_t *it) {
     if (!it) {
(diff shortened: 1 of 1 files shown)


plugins/gtkui/gtkui.c
```diff
@@ -450,6 +456,11 @@ gtkui_set_titlebar (DB_playItem_t *it) {
         deadbeef->pl_item_unref (it);
     }
     set_tray_tooltip (str);
+    if (ctx.update > 0) {
+        g_timeout_add(ctx.update, set_title_cb, NULL);
+    } else if (ctx.update < 0) {
```
Alexey-Yakovenko, line 461: Hmm.. this seems to be handling a case, which is written in a comment, but isn't really supported -- i.e. there's no way deadbeef would set this to a value less than 0.. At least I can't find one. Do you think we should be doing this?

saivert: In that case there is no need to check for a negative ctx.update.

Alexey-Yakovenko (changes requested):

PR #2515, opened 2021-01-28
ID3v2 frame sizes are before unsynchronisation

sktt:
This fixes https://github.com/DeaDBeeF-Player/deadbeef/issues/2514

### DIFF
--- a/junklib.c
+++ b/junklib.c
@@ -3570,7 +3570,6 @@ junklib_id3v2_sync_frame (uint8_t *data, int size) {
         *writeptr = *data;
         if (data[0] == 0xff && size >= 2 && data[1] == 0) {
             data++;
-            size--;
         }
         writeptr++;
         data++;


Alexey-Yakovenko:
Thank you for in-depth investigation and the fix. I don't have time to verify it today, but I'll try my best to do it as soon as possible!

sktt:
My pleasure! I hope that I got it right. The more time I spend on this the more it appears that the unsync flag is for the entire tag (for 2.3) so any de-unsync should thus be done before parsing the frames. But not everyone is doing that... Probably it's problematic if some frame (APIC?) is huge...

Mediainfo increments Size when it finds the 0xff00 pattern:
https://github.com/MediaArea/MediaInfoLib/blob/master/Source/MediaInfo/Tag/File_Id3v2.cpp#L513

IdSharp clearly neglects unsync bytes from size:
https://github.com/judwhite/IdSharp/blob/master/IdSharp.Tagging/ID3v2/ID3v2Utils.cs#L70

libid3tag de-unsync the entire tag at once for 2.3 
https://github.com/markjeee/libid3tag/blob/master/tag.c#L456 This works well because the spec clearly states that the ID3v2 tag size is for the entire tag *after* unsync bytes are added.

For 2.4 it appears to be for either the entire tag OR per frame.. I hope it's invalid to unsynch twice then (per frame and per tag)..
Also found for 2.4 that
```
   The frame ID is followed by a size descriptor containing the size of
   the data in the final frame, after encryption, compression and
   unsynchronisation.
 ```
 
So in the 2.4 the frame size should be the actual byte size on disk, whereas for 2.3 it is the size after the entire tag has been deunsynchronized. In this case this patch might break id3v2.4 tags... It could also be that this only refers to the added per-frame unsync, in which case this patch might still be correct.

Also it seems, since the entire tag is unsynched, that also frame sizes must be deunsynched which is not supported by junklib..

Alexey-Yakovenko:
The main problem with this PR is that it's incredibly difficult to test..

id3v2.3 is relatively simple -- I just need a tagger that would use unsynchronization, and write some unicode field, and hope that utf16 bom will be added.. then check that all fields show up correctly.

id3v2.4 is more tricky -- all taggers will prefer to write utf8 instead, so the unsynchronization won't happen.

Because of the complexity of validation, I'll assume that id3v2 specification is correct, and apply the fix you provided only for 2.3, and keep the original code for 2.4.

Alexey-Yakovenko:
Also, I verified that id3v2.2 follows the same unsynchronization scheme as 2.3

Alexey-Yakovenko:
Merged manually.

sktt:
Looks good i think. I just switched off unsync in my own clone of
flac2mp3.pl, made it optional  and disabled by default, because it's
outdated that mp3 players don't know how to handle id3 tags at all.

About testing.. is it possible to test encoding/decoding thru end to end
tests that uses popular taggers for generating fixtures?(didn't look at
existing ones yet)..

Alexey-Yakovenko:
I don't have any existing tests of similar nature, i.e. not end to end.. but I can quite easily add tests which verify pre-made mp3 files. So if you have easy way of creating such files, I would be willing to add the necessary tests.
(one limitation though.. all deadbeef tests are running only on mac, since they're written in ObjC and using XCTest)

Alexey-Yakovenko:
If you decide to go for it: the test files must be "empty", e.g. only contain silence or a tone signal, and be small (1 mpeg frame is enough)

Alexey-Yakovenko:
It would be nice to cover at least 2.3 and 2.4 with unsynchronization, using UTF16 and to have BOM.
Even better if we could get a 2.2 file.. but I don't know if any tools can do this.

I remember that you mentioned that frame size can contain unsync bytes... but I think this is a wrong assumption, since the size is encoded in a special way, and can't contain the FF FE sequence.

PR #2539, opened 2021-02-25
gtkui: New plugin preferences UI

saivert:
Notes:

I had to change the code in `apply_conf` from using `lookup_widget` glade utility function to `g_object_get_data` directly as it needs to work with any container and not just top levels where it can find the outermost parent. `gtkui_run_dialog` expected a parent dialog passed to it which remains until none of the child objects are needed. I hope this change has no side effects this time.

### DIFF
--- a/plugins/gtkui/callbacks.h
+++ b/plugins/gtkui/callbacks.h
@@ -1419,3 +1419,22 @@ on_minimize_on_startup_clicked         (GtkButton       *button,
 void
 on_move_to_trash_clicked               (GtkButton       *button,
                                         gpointer         user_data);
+
+void
+on_plugin_conf_tab_btn_clicked         (GtkRadioButton       *button,
+                                        gpointer         user_data);
+
+void
+on_plugin_info_tab_btn_clicked         (GtkRadioButton       *button,
+                                        gpointer         user_data);
+
+void
+on_plugin_license_tab_btn_clicked      (GtkRadioButton       *button,
+                                        gpointer         user_data);
+
+
+void
+on_plugin_notebook_switch_page         (GtkNotebook     *notebook,
+                                        GtkWidget       *page,
+                                        guint            page_num,
+                                        gpointer         user_data);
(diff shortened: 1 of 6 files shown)


kuba160:
Look on Windows (GTK2):
image
GTK3:
image

Previous look for reference:
image

saivert:
Not sure what to do about the inset border on GTK2 really. It looks okay here with Adwaita GTK2 theme on Gnome desktop. Styling and themes is pain.
Screenshot from 2021-02-26 08-21-50

saivert:
@kuba160 so I made some fixes to the styling. Should look better on GTK2 now. The buttons also look more appropriate on GTK3 now.

Alexey-Yakovenko:
Looks great! I still didn't have time to verify or look at the code, but looking forward to it :D

Alexey-Yakovenko:
So, it actually works quite great, but I see 2 issues.. although I'm not sure what's the best way to fix it and whether it makes sense.

1. Some widgets have too large min width, so they cause horizontal scrolling.. probably the min-width should be reduced. I can't see where it's set though, maybe it's added by default by Gtk to GtkEntry, or something like this.
2. The scrollbars act super weird -- although this is probably something that only happens with Mac version of GTK, and maybe is irrelevant. I don't care as long as it works well on Linux. Here's what's happening:
    * Initially, I can only use the Vertical scrollbar, and it's invisible.. Only becomes visible when I start scrolling
    * I can't use horizontal scrollbar with the mouse wheel - holding any modifier keys (shift/cmd/opt/ctrl) doesn't help.
    * I can scroll horizontally by moving the mouse towards the bottom edge of the view, then the horizontal scrollbar appears, and I can drag, but:
    * After I drag the horizontal one -- I can no longer scroll the vertical one.
    * Also, if the mouse pointer is over any dropdown control, using the mouse wheel will change the value of the dropdown, instead of scrolling the view.

Alexey-Yakovenko:
Either way, I think it's usable as it is, so if you're ready -- I can hit the merge button.

PR #2542, opened 2021-02-27
premake5: export all symbols on windows

kuba160:
Fixes issues with symbols not exporting correctly on local build on windows.

### DIFF
--- a/scripts/windows_postbuild.sh
+++ b/scripts/windows_postbuild.sh
@@ -31,9 +31,6 @@ done
 cp -uv translation/help.ru.txt  "$1/doc/"
 
 # Libraries
-rm -fv "$1"/plugins/*.lib | true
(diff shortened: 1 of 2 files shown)


kuba160:
It looks like curl dll does not include ssl built-in anymore. Because of that we need to lookup for dependencies for the libraries aswell.

scripts/windows_postbuild.sh
```diff
@@ -44,8 +42,23 @@ ldd "$1/plugins/"*.dll "$1/deadbeef.exe" | awk 'NF == 4 {print $3}; NF == 2 {pri
 
 cp -uv `cat .libraries.tmp` "$1/"
 
+# Libraries of libraries :)
```
Alexey-Yakovenko, line 45: Would be great to add a comment what actually happens here :)

PR #2545, opened 2021-03-03
tf: Implement $rgb() function

saivert:


saivert:
Screenshot from 2021-03-03 14-11-26

PR #2549, opened 2021-03-06
gtkui: plugin report menu item

saivert:


plugins.c
```diff
@@ -94,6 +94,7 @@ const char *lowprio_plugin_ids[] = {
 // internal plugin list
 typedef struct plugin_s {
     void *handle;
+    const char *filepath;
```
Alexey-Yakovenko, line 97: This should not be const. By convention, pointers which need to be freed are non-const. It's makes the code easier to read, and you wouldn't need to cast.

plugins.c
```diff
@@ -1308,6 +1312,9 @@ plug_unload_all (void) {
         if (plugins->handle) {
             dlclose (plugins->handle);
         }
+        if (plugins->filepath) {
```
Alexey-Yakovenko, line 1315: You don't need this if. `free` does this check by itself.

Alexey-Yakovenko (changes requested):

PR #2551, opened 2021-03-06
gtkui: fix rename the right plt_idx

sktt:
I messed up a bit with my last pr. Now renames the right playlist

### DIFF
--- a/plugins/gtkui/gtkui.c
+++ b/plugins/gtkui/gtkui.c
@@ -576,30 +576,27 @@ gtkui_get_curr_playlist_mod (void) {
 }
 
 int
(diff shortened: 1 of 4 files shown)


plugins/gtkui/gtkui.c
```diff
@@ -576,30 +576,27 @@ gtkui_get_curr_playlist_mod (void) {
 }
 
 int
-gtkui_rename_current_playlist (void) {
-    int idx = deadbeef->plt_get_curr_idx ();
-    if (idx != -1) {
```
sktt, line 581: essentially just moving this check to one level up

Issue #1201, opened 2015-03-26
Tags not working for .wav files


marblenix:
I'm not sure if I have configured it correctly but deadbeef seems incapable of reading tags of .wav files. .mp3 files work just fine though.

`DeaDBeeF 0.6.2`

`mediainfo 01\ Modest\ Mouse\ -\ Strangers\ to\ Ourselves.wav`

```
General
Complete name                            : 01 Modest Mouse - Strangers to Ourselves.wav
Format                                   : Wave
File size                                : 34.4 MiB
Duration                                 : 3mn 24s
Overall bit rate mode                    : Constant
Overall bit rate                         : 1 411 Kbps
Album                                    : Strangers to Ourselves
Album/Performer                          : Modest Mouse
Track name                               : Strangers to Ourselves
Track name/Position                      : 1
Track name/Total                         : 15
Performer                                : Modest Mouse
Composer                                 : Modest Mouse
Director                                 : Modest Mouse
Genre                                    : Alternative Rock
Recorded date                            : 2015-03-13
...

Audio
Format                                   : PCM
Format settings, Endianness              : Little
Format settings, Sign                    : Signed
Codec ID                                 : 1
Duration                                 : 3mn 24s
Bit rate mode                            : Constant
Bit rate                                 : 1 411.2 Kbps
Channel(s)                               : 2 channels
Sampling rate                            : 44.1 KHz
Bit depth                                : 16 bits
Stream size                              : 34.4 MiB (100%)
```

![deadbeef](https://cloud.githubusercontent.com/assets/6401427/6858010/1eeafab8-d3d9-11e4-9d0f-e7a8181da8fb.png)

![deadbeef2](https://cloud.githubusercontent.com/assets/6401427/6858032/53ea0100-d3d9-11e4-8392-81b63f2bb0ac.png)

Alexey-Yakovenko:
What tag format do you use for tagging the wav files?
Or which tagging tool?

marblenix:
The files were ripped directly from CD using [XLD](http://sourceforge.net/projects/xld/), which queries freedb.freedb.org for tags. The tag file format appears to be RIFF.

```
hexdump -C 01\ Modest\ Mouse\ -\ Strangers\ to\ Ourselves.wav | head -3
00000000  52 49 46 46 9c 67 26 02  57 41 56 45 66 6d 74 20  |RIFF.g&.WAVEfmt |
[...]

PR #2575, opened 2021-04-02
sndfile: inital support of RIFF tags.

xinyazhang:
This add *reading* support of RIFF tags for sndfile.
This is a partial solution of #1201.

### DIFF
--- a/plugins/sndfile/sndfile.c
+++ b/plugins/sndfile/sndfile.c
@@ -352,6 +352,9 @@ sndfile_seek (DB_fileinfo_t *_info, float sec) {
     return sndfile_seek_sample (_info, sec * _info->fmt.samplerate);
 }
 
+static int
+_sndfile_ctx_read_tags (DB_playItem_t *it, SNDFILE *ctx);
+
 static DB_playItem_t *
 sndfile_insert (ddb_playlist_t *plt, DB_playItem_t *after, const char *fname) {
     trace ("adding file %s\n", fname);
@@ -382,8 +385,6 @@ sndfile_insert (ddb_playlist_t *plt, DB_playItem_t *after, const char *fname) {
     trace ("calling sf_open_virtual ok\n");
     int64_t totalsamples = inf.frames;
     int samplerate = inf.samplerate;
-    sf_close (info.ctx);
-    deadbeef->fclose (info.file);
 
     float duration = (float)totalsamples / samplerate;
     DB_playItem_t *it = deadbeef->pl_item_alloc_init (fname, plugin.plugin.id);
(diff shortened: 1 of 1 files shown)


plugins/sndfile/sndfile.c
```diff
@@ -512,12 +513,79 @@ sndfile_insert (ddb_playlist_t *plt, DB_playItem_t *after, const char *fname) {
...
+    DB_fileinfo_t* info = sndfile_open (0);
+    int res = sndfile_init (info, it);
+    if (res) {
+        return -1;
+    }
+    SNDFILE *ctx = ((sndfile_info_t*)info)->ctx;
+    _sndfile_ctx_read_tags (it, ctx);
+
+    sndfile_free (info);
+    free (info);
```
Alexey-Yakovenko, line 556: The need to do this seems to be a bug / memory leak.
The `sndfile_free` should be calling it instead.

plugins/sndfile/sndfile.c
```diff
@@ -512,12 +513,79 @@ sndfile_insert (ddb_playlist_t *plt, DB_playItem_t *after, const char *fname) {
...
+    { SF_STR_TRACKNUMBER,   "track" },
+    { SF_STR_GENRE,         "genre" },
+};
+
+static int
+sndfile_read_metadata (DB_playItem_t *it) {
+    DB_fileinfo_t* info = sndfile_open (0);
+    int res = sndfile_init (info, it);
+    if (res) {
+        return -1;
```
Alexey-Yakovenko, line 550: `sndfile_free` needs to be called here

plugins/sndfile/sndfile.c
```diff
@@ -512,12 +513,79 @@ sndfile_insert (ddb_playlist_t *plt, DB_playItem_t *after, const char *fname) {
...
+        if (charset) {
+            // recode non-UTF-8 metadata string.
+            size_t l = strlen (meta_str);
+            char recbuf[l*10];
+            int res = deadbeef->junk_recode (meta_str, (int)l, recbuf, (int)(l * 10 - 1), charset);
+            if (res >= 0) {
+                deadbeef->pl_add_meta (it, kv->meta_name, recbuf);
+            }
+            else {
+                deadbeef->pl_add_meta (it, kv->meta_name, "<UNRECOGNIZED CHARSET>");
```
Alexey-Yakovenko, line 579: This should not be needed. If something can't be loaded -- just skip it.

xinyazhang:
The build failed on Windows. Any ideas?

plugins/sndfile/sndfile.c
```diff
@@ -512,12 +513,79 @@ sndfile_insert (ddb_playlist_t *plt, DB_playItem_t *after, const char *fname) {
...
+    for (i = 0; i < ARRAY_LEN(metamap_items); i++) {
+        const struct metamap_s *kv = &metamap_items[i];
+        const char *meta_str = sf_get_string (ctx, kv->str_type);
+        if (!meta_str)
+            continue ;
+        const char *charset = deadbeef->junk_detect_charset (meta_str);
+        if (charset) {
+            // recode non-UTF-8 metadata string.
+            size_t l = strlen (meta_str);
+            char recbuf[l*10];
```
Alexey-Yakovenko, line 573: This should use malloc/free here instead, since the `l` can be any size, and may be larger than available stack.
In such case it would produce a null pointer and cause a crash, without any possibility to handle it.

plugins/sndfile/sndfile.c
```diff
@@ -512,12 +513,79 @@ sndfile_insert (ddb_playlist_t *plt, DB_playItem_t *after, const char *fname) {
     }
 
     deadbeef->pl_add_meta (it, "title", NULL);
+
+    // delay the closing to the end of the function to load RIFF tags.
```
Alexey-Yakovenko, line 517: This comment is unnecessary. It kinda explains what the PR does, but it's clear enough from the code.

Alexey-Yakovenko (changes requested):

Alexey-Yakovenko:
> The build failed on Windows, but there isn't too much info in the build log. Any ideas?

I re-ran the builds, seems like a random failure which isn't related to your change.

plugins/sndfile/sndfile.c
xinyazhang: I've created another PR to fix `sndfile_free`.

PR #2580, opened 2021-04-10
netbsd build warnings/errors fixes

Alexey-Yakovenko:


Alexey-Yakovenko:
To: Onno van der Linden

First thing first, it seems that you changed the build scripts to produce and use all libraries as dynamic instead of static.
I'm not sure why you'd want this, but this seems wrong to me.

However, it seems that with your changes everything is still working, so I will merge it and see what happens.
In the worst case scenario, I'd have to revert.

PR #2586, opened 2021-04-17
gtkui: accept drag and drop on main window

saivert:


kuba160:
Works well, the only thing I would like to comment on: Dialog "Add files..." pops out for a short moment when dragging item from playlist and dropping on other GUI element. There is also a green indicator above the menu bar when holding it. (Tested on Windows)

saivert:
Well it needs some more work. It currently allows dragging a subsong from a cuesheet from the playlist window onto the main window titlebar and then that results in adding the full mp3 file (not just the subsong being dragged) to the current playlist with wrong cached metadata.

I would have to disable drag when source is coming from itself as it makes no sense to do this and I don't know how to fix the subsong issue.

PR #2587, opened 2021-04-19
gtkui: Add scroll event handler to seekbar widget

saivert:


plugins/gtkui/ddbseekbar.c
```diff
@@ -153,6 +153,41 @@ static gboolean ddb_seekbar_real_configure_event (GtkWidget* base, GdkEventConfi
 	return result;
 }
 
+// Copied from hotkeys.c
+static int
+seek_sec (float sec) {
+    deadbeef->pl_lock ();
```
Alexey-Yakovenko, line 159: why do you need a pl_lock here?

Issue #2494, opened 2021-01-07
[Linux / Fedora][DeadBeeF 1.8.4] Local file mask not respected for album art


pihug12:
### Steps to reproduce the problem

When I play a file where there should be no artwork, DeaDBeeF picks a random JPG file in the same folder.

The pattern is set to: `*cover*.jpg;*front*.jpg;*folder*.jpg;*cover*.png;*front*.png;*folder*.png`

The file name of the pictures picked don't respect this mask (`mpv-shot0002.jpg`, `62423173-7736d800-b6f0-11e9-9ddd-e48832259695.jpg`).

### Information about the software:

DeadBeeF version: 1.8.4
OS: Fedora 32

kuba160:
I think the issue is present on Windows aswell, had some tracks in Downloads folder and it picked some pictures I had (1st one alphabetically I think). Thought at first that it was Windows only problem, maybe some regression happened when it was ported to Windows.

PR #2595, opened 2021-04-30
Sanitize paths for artwork-legacy

kuba160:
These commits were sent as PR to DeaDBeeF-for-Windows by Keith. I told him it would be more appropriate to sent it straight here. It sadly never happened but now I'm doing this because I still think these changes are useful.

Please let me know what you think about these changes and request changes if necessary (`is_illegal_windows_name` could be rewritten for better readability for sure 🙂).

### DIFF
--- a/plugins/artwork-legacy/artwork.c
+++ b/plugins/artwork-legacy/artwork.c
@@ -925,6 +925,83 @@ is_illegal_windows_name (const char* str) {
     return 0;
 }
 
+// Maybe put this in utils if useful in other places?
+static size_t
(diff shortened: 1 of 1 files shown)


kuba160:
I also found that artwork-legacy has hardcoded filemasks to pick any image (commit: https://github.com/DeaDBeeF-for-Windows/deadbeef/commit/3c294c062aa6505a4de0669dc057ac6175974d09 ), related to issue #2494 . I'm not sure if you want to squash this PR, so I'm not adding it here. Also, do you want to add these filemaks to `DEFAULT_FILEMASK`?

PR #2624, opened 2021-06-30
pulse: try to use sample rates above 192kHz if possible

ToadKing:


Alexey-Yakovenko:
I didn't like this fix, so I fixed it differently. Hope it works for you as well.

PR #2642, opened 2021-08-05
Implement column alignment for cocoaui on medialib branch

tomjepp:
Implement left, center and right alignment for text in playlist cells.

### DIFF
--- a/plugins/cocoaui/Playlist/PlaylistViewController.h
+++ b/plugins/cocoaui/Playlist/PlaylistViewController.h
@@ -26,7 +26,6 @@
 #import "ConverterWindowController.h"
 
 typedef NS_ENUM(NSInteger, PlaylistColumnAlignment) {
-    ColumnAlignmentUnknown = -1,
     ColumnAlignmentLeft = 0,
     ColumnAlignmentCenter = 1,
     ColumnAlignmentRight = 2,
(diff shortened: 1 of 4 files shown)


plugins/cocoaui/Playlist/PlaylistViewController.h
```diff
@@ -25,12 +25,19 @@
 #import "PlaylistView.h"
 #import "ConverterWindowController.h"
 
+typedef NS_ENUM(NSInteger, PlaylistColumnAlignment) {
+    ColumnAlignmentUnknown = -1,
```
Alexey-Yakovenko, line 29: This should not be allowed. When the value is loaded from file, it should be converted into representable range of the enum, with ColumnAlignmentLeft being the default.

Alexey-Yakovenko (changes requested):

plugins/cocoaui/Playlist/PlaylistViewController.h
tomjepp: This is the default value that's set by the add new column UI now. I can change that to left if you'd prefer?

tomjepp: I've implemented that.

Alexey-Yakovenko:
Thanks, merged!

Issue #2636, opened 2021-07-14
Unable to seek after ~12:25h / 13:31h


SuperVirus:
DeaDBeeF is unable to seek beyond a certain timestamp in very long audio files.
The limit is about 13:31h for stereo files with a sample rate of 44.100 and about 12:25h for stereo files with a sample rate of 48.000.
Only seeking in the file is affected. Playback will continue beyond those limits without problems.

It guess a 32bit integer is used for the seek position:
44100×2×3600×13,5 => 4 286 520 000 samples
48000×2×3600×12,4 => 4 285 440 000 samples
Both values are close the maximum of 4 294 967 295 for a 32bit integer.

### Steps to reproduce the problem
- Open a very long (min 14h audio file in DeaDBeeF)
- Seek to a timestamp > 13:40h

### What's going on? Describe the problem in as much detail as possible
Depending on the decoder plugin the behavior differs:
- For FLAC and OPUS the seek does not happen and the playback continues.
- For ffmpeg (tested with MP4-AAC and OPUS) the playback position jumps to the start of the file.

### Information about the software:
Deadbeef version: 1.8.7 and git  6d3b11262d5fbe0326b408bc73ae01861505d4ef
OS: Arch Linux 64bit

Here are some 14h test files (FLAC, only silence)
:[testfiles.zip](https://github.com/DeaDBeeF-Player/deadbeef/files/6812680/testfiles.zip)

Alexey-Yakovenko:
It's the same with any files.. It's not hard to fix, but a little time consuming. In the meantime I'm doing 2 things:

1. adding a new extended decoder plugin interface, to add 64 bit seek offset API.
2. implementing the interface and using 64 bit offset in the flac plugin.

Alexey-Yakovenko:
(not closing the issue yet, until it's fixed in all relevant decoders).

Alexey-Yakovenko:
Sorry, it was closed by the git merge

Alexey-Yakovenko:
A list of other codecs that could potentially benefit from the same fix, alphabetical order:
* aac
* alac
* ape (seektable is 32bit)
* mp3
* musepack
* sndfile
* tta (seektable is 32 bit)
* wavpack
* wma

PR #2660, opened 2021-08-23
Updated ffmpeg plugin to implement decoder2 API

SuperVirus:
Implementation decoder2 API to support 64bit seek function in ffmpeg plugin.

Related to #2636

### DIFF
--- a/plugins/ffmpeg/ffmpeg.c
+++ b/plugins/ffmpeg/ffmpeg.c
@@ -44,7 +44,7 @@
 #define min(x,y) ((x)<(y)?(x):(y))
 #define max(x,y) ((x)>(y)?(x):(y))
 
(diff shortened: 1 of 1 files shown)


plugins/ffmpeg/ffmpeg.c
```diff
@@ -460,7 +460,7 @@ ffmpeg_read (DB_fileinfo_t *_info, char *bytes, int size) {
 }
 
 static int
-ffmpeg_seek_sample (DB_fileinfo_t *_info, int sample) {
+ffmpeg_seek_sample64 (DB_fileinfo_t *_info, int64_t sample) {
```
ToadKing, line 463: The cast below on line 473 isn't required anymore with this change.

PR #2686, opened 2021-10-23
gtkui: extend gui syntax with a label item

saivert:
Allows you to break up options groups or place a label above an entry box to save on space.

Syntax:

   label "\<label text\>" \<left|center|right\>

### DIFF
--- a/plugins/gtkui/gtkui.c
+++ b/plugins/gtkui/gtkui.c
@@ -2079,8 +2079,10 @@ ddb_gui_GTK3_load (DB_functions_t *api) {
 
 static const char settings_dlg[] =
     "property \"Ask confirmation to delete files from disk\" checkbox gtkui.delete_files_ask 1;\n"
+    "label \"Status icon settings:\" left;\n"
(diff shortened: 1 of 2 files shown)


Alexey-Yakovenko:
The syntax looks off to me. Until now, the syntax was always starting with a "property":
`property text type identifier parameters;`

Please change the syntax to match everything else.

Alexey-Yakovenko:
I'm not saying that the original syntax is very good, considering that nobody even tried to think how to make it good -- but it's better to have ugly and consistent, than ugly and random.

Alexey-Yakovenko:
The diff looks surprisingly large for such a small change. Is there a reason for that?

Alexey-Yakovenko:
Because the change touches so much code, I'd not be able to merge this until I'm able to test as much as possible for regressions. There are quite a few dialogs in the player that rely on this functionality, and I don't want them to break apart.

Alexey-Yakovenko:
(and currently I'm stuck without a linux VM, so it'll have to wait for a bit)

saivert:
It indents a lot of existing code to put it inside an if clause.

I can of course change it to always start with property even if it isn't really defining a property. It would simplify the code some.

PR #2687, opened 2021-10-23
gtkui: implement scale property for DdbVolumeBar

saivert:
Now you can right click the volume bar widget in design mode and pick the scale you want.


Alexey-Yakovenko:
> Now you can right click the volume bar widget in design mode and pick the scale you want.

why in design mode though, and not in the normal mode?

PR #2749, opened 2022-02-17
streamer: stop after current now works like fb2k (#2124)

### LINKED ISSUE #2124: Переход выделения на следующий трек при включенной опции "Stop after current"
Есть ли аналог опции "Cursor follows playback" (из проигрывателя Foobar)?
При включенной опции "Stop after current" хотелось бы, чтобы по окончании трека (и остановки воспроизведения) выделение переходило на следующий трек. И по нажатии на кнопку Play (клавишу Enter) - начинал звучать следующий трек.

Deadbeef version: 1.8.0
OS: Debian

rsekman:
after the streamer stops due to stop after current (track/album), hitting play will now play the next track. before it would play the track that just ended

Resolves #2124

Unfortunately, the selection indicated in `gtkui` is not updated to reflect the new cursor position. I think this may require new signals to implement.

### DIFF
--- a/streamer.c
+++ b/streamer.c
@@ -1654,9 +1654,9 @@ streamer_thread (void *unused) {
             if (stop) {
                 stream_track (NULL, 0);
                 if (next) {
-                    playlist_t* cur_pl = plt_get_curr();
-                    int cursor = plt_get_item_idx(cur_pl, next, PL_MAIN);
-                    plt_deselect_all(cur_pl);
+                    playlist_t* cur_pl = plt_get_curr ();
+                    int cursor = plt_get_item_idx (cur_pl, next, PL_MAIN);
+                    plt_deselect_all (cur_pl);
                     pl_set_selected_in_playlist (cur_pl, next, 1);
                     messagepump_push (DB_EV_PLAYLISTCHANGED, 0, DDB_PLAYLIST_CHANGE_SELECTION, 0);
                     plt_set_cursor (cur_pl, PL_MAIN, cursor);
--- a/streamer.c
+++ b/streamer.c
@@ -1656,7 +1656,10 @@ streamer_thread (void *unused) {
                 if (next) {
                     playlist_t* cur_pl = plt_get_curr();
                     int cursor = plt_get_item_idx(cur_pl, next, PL_MAIN);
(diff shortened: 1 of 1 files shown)


Alexey-Yakovenko:
You can change the selection by running something like this
```
plt_deselect_all(plt);
plt_item_set_selected(plt, next, 1);
sendmessage (DB_EV_PLAYLISTCHANGED, 0, DDB_PLAYLIST_CHANGE_SELECTION, 0);
```
There's no need to propagate this as an event.

rsekman:
> You can change the selection by running something like this

Oh, excellent. I'm not familiar enough with the codebase yet so i was trying with `DB_EV_SONGCHANGED`.
There is a bit of a timing issue with this: the selection will be advanced when the last streaming block is loaded, not when it finishes playing. As a result the selection marker advances just before the playing track indicator. A video may be easier to understand:

https://user-images.githubusercontent.com/29686694/154492250-dd6939e3-282c-44c3-b932-76d8b4a5faf3.mp4

The next track, Run To The Hills, becomes highlighted before The Number of the Beast stops being bold. Of course this is only an issue if one happens to be looking at the playlist just as the track ends...

> I believe, this behavior is configurable in fb2k, but not in deadbeef.

I don't think it is. But it has been almost a year since I last used fb2k. fb2k has both "cursor follows playback" and "playback follows cursor" and I was never quite able to figure out the.

Alexey-Yakovenko:
> There is a bit of a timing issue with this: the selection will be advanced when the last streaming block is loaded, not when it finishes playing. As a result the selection marker advances just before the playing track indicator.

If you want to have this in sync - it will be quite hard.
You would have to put some sort of a special marker into the last streaming block, and ensure that the marker is carried to the decoded blocks queue.
And then when this decoded block is consumed -- you would know that the cursor and selection needs to jump to the next track.

Maybe you should give it a try.

rsekman:
> Maybe you should give it a try.

I might when I don't have anything with a higher reward/effort ratio to work on. For now I am not bothered enough by it. It is more important to me that "play" behaves as I expect it to. Most of the time I am not looking at the playlist, after all.

Oleksiy-Yakovenko:
Closing: out of date

PR #2821, opened 2022-06-16
add selection metadata widget

elementdavv:
There is a selection properties widget, but no selection metadata one.
I like to see the metadata as well, so add it.
Maybe it will make sense to others.


Alexey-Yakovenko:
Overall looks very good. There are no crashes, and address sanitizer is quiet.

But I noticed 2 things.

1. In foobar2000, there's a single widget "Track Properties", which is configurable to display all the various things. (In the case of deadbeef, there's no separation in that many categories -- there are only 2, which can be Metadata and General).

image

Do you think it would be hard to implement something like that?

2. The meaning of Track Properties and Track Metadata is swapped, compared to the tab names in the Track Properties 
dialog - see the screenshot below.

3. The new widget displays ":" symbol on all property names. The ":" prefix is supposed to be removed before displaying.
image

Alexey-Yakovenko:
Another problem spotted:

while using the new widget, there are errors in the log:
```
(deadbeef:15357): Gtk-[1;33mWARNING[0m **: [34m20:49:45.648[0m: ../gtk/gtkliststore.c:1029: Invalid column number 2 added to iter (remember to end your list of columns with a -1)
```

This occurs on every selection change.

Alexey-Yakovenko:
It occurs from this callstack
image

elementdavv:
The key name issue and invalid column warning fixed.
And a foobar2000 style Track Properties widget added as well.
Please check it.

Alexey-Yakovenko:
Is there any reason why you added 2 more widgets (resulting in 3 total widgets doing mostly the same thing), instead of having just one configurable "Selection Properties"?

Regarding the current implementation, it's a shame that GtkTreeView doesn't support group headers / column spanning separators, so I guess we'll have to live with how it looks. This is not going to be implemented in GTK any time soon (even if it does -- probably will only come to GTK4)

image

However, it should look slightly cleaner if the separator line "------" is not used at all, better to keep it blank.
And the text "Metadata" and "Properties" could be rendered in bold, if possible.

Other suggestion:

You did a great job saving the settings inside of the widget the layout -- this is the correct way of doing this.
However, I think it should be more expandable for future changes.

I suggest to change the section config to some kind of array..
It's a bit more tricky than it should be, since the keyvalues abstract too much of JSON, and there's no way to write a proper JSON array.

We may consider either expanding this API, in order to write a proper array, or we can write the array as comma separated string

Option1: more work, save proper JSON array: `"sections": [ "metadata", "properties" ]`
Option2: potentially less work, save array as string: `"sections": "metadata,properties"`
This way we can easily add more sections as needed.

Alexey-Yakovenko:
The `properties_mode_t` enum, in particular, is not quite so flexible.
Perhaps, consider changing it to a bit set?

```typedef enum {
    SECTION_METADATA = 1<<0,
    SECTION_PROPERTIES = 1<<1,
} properties_mode_t;
```

Alexey-Yakovenko:
While you're at it -- it would be great to change this global setting to be stored in the layout as well:
"gtkui.track_properties.show_headers"

elementdavv:
> While you're at it -- it would be great to change this global setting to be stored in the layout as well: "gtkui.track_properties.show_headers"

Are you sure to do that? I define this config as conforming with old config of "gtkui.selection_properties.show_headers". May be more similar items live somewhere. The change may breaks functioning of existing users.

Alexey-Yakovenko:
> Are you sure to do that? I define this config as conforming with old config of "gtkui.selection_properties.show_headers". May be more similar items live somewhere. The change may breaks functioning of existing users.

Fairly sure that this would be better, since currently this setting is not synchronized across widgets anyway.

elementdavv:
Now it is clear to be only one Selecton Properties widget.
The exapi framework for widget itself use json config object, but expects a string format of config/value from widgets and does not parse it to json object. So using json config inside widgets will result in many backslacks in the config file, that is so ugly.

Alexey-Yakovenko:
Thanks, this is much better.
You need to cleanup some json line, since they cause compile error on mac (from the unused variable warning):

```
--- a/plugins/gtkui/selpropertieswidget.c
+++ b/plugins/gtkui/selpropertieswidget.c
@@ -22,7 +22,6 @@
 */
 
 #include <assert.h>
-#include <jansson.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -196,7 +195,6 @@ _deserialize_from_keyvalues (ddb_gtkui_widget_t *widget, const char **keyvalues)
     for (int i = 0; keyvalues[i]; i += 2) {
         if (!strcmp (keyvalues[i], "section")) {
                        live = TRUE;
-                       json_t *section= json_loads(keyvalues[i+1], 0, NULL);
                        char *sect = strdup(keyvalues[i+1]);
                        char *token = strtok(sect, ",");
```

The section variable should be `uint32_t` instead of the enum type -- since it's a bit set, and not an enum value.
It also makes sense to rename it to "visible_sections" for clarity: `uint32_t visible_sections;`

Bits are typically combined with a `|` instead of `+`:
```
-       w->section = SECTION_METADATA + SECTION_PROPERTIES;
+    w->visible_sections = SECTION_METADATA | SECTION_PROPERTIES;
```
(encountered 3 times)

Aside of this, it seems like your text editor is configured to use tabs instead of spaces. Please replace tabs with 4 spaces, since this is the convention used in deadbeef source code.

elementdavv:
Updated.

PR #2839, opened 2022-08-07
gtkui: improve sort by custom dialog

rsekman:
### Before

 2022-08-07T01:38:41+02:00

### After
2022-08-07T07:37:27+02:00


Oleksiy-Yakovenko:
The interface.c file is generated from the plugins/gtkui/deadbeef.glade file.
Therefore you can't modify it directly, and it needs to be done using Glade, the version present in deadbeef source tree under the `tools/glade` folder.

PR #2840, opened 2022-08-07
gtkui: improve sort by custom dialog

rsekman:
Like #2839, but with glade

### Before

2022-08-07T17:38:14+02:00

### After 

2022-08-07T17:39:09+02:00

### DIFF
--- a/plugins/gtkui/actionhandlers.c
+++ b/plugins/gtkui/actionhandlers.c
@@ -624,28 +624,37 @@ action_sort_custom_handler_cb (void *data) {
     gtk_dialog_set_default_response (GTK_DIALOG (dlg), GTK_RESPONSE_OK);
 
     GtkComboBox *combo = GTK_COMBO_BOX (lookup_widget (dlg, "sortorder"));
-    GtkEntry *entry = GTK_ENTRY (lookup_widget (dlg, "sortfmt"));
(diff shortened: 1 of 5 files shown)


plugins/gtkui/actionhandlers.c
```diff
@@ -624,29 +624,38 @@ action_sort_custom_handler_cb (void *data) {
...
         deadbeef->conf_set_int ("gtkui.sortby_order", order);
         deadbeef->conf_set_str ("gtkui.sortby_fmt_v2", fmt);
 
+
         ddb_playlist_t *plt = deadbeef->plt_get_curr ();
         deadbeef->plt_sort_v2 (plt, PL_MAIN, -1, fmt, order == 0 ? DDB_SORT_ASCENDING : DDB_SORT_DESCENDING);
         deadbeef->plt_save_config (plt);
         deadbeef->plt_unref (plt);
 
+        free( (void*) fmt);
```
Oleksiy-Yakovenko, line 657: Why do you need the cast here?

Oleksiy-Yakovenko (approved):
The code looks ok, I'll merge after testing it locally.

plugins/gtkui/actionhandlers.c
rsekman: Ah, I don't. I used to have `const char *fmt = gtk_text_buffer_get_text ..`  on line 646 which would generate a compiler warning. I shut that up with the cast but the correct thing to do is of course to drop `const`.

PR #2861, opened 2022-10-01
Fix build with ffmpeg 5.0+

xuzhen:
av_iformat_next was replaced with av_demuxer_iterate. https://trac.ffmpeg.org/wiki/Bump59


xuzhen:
Updated. Fixed the argument of ```av_demuxer_iterate```. And there is no need to call ```av_register_all``` for ffmpeg 4.0+ https://github.com/leandromoreira/ffmpeg-libav-tutorial/issues/29

Issue #2847, opened 2022-08-24
[Request] Finding cover art in parent directory


jieweix:
There's often this kind of multi-disc structure:
```
.
├── cover.jpg
├── Disc 1
│   └── 01. For Bunita Marcus (1985).flac
└── Disc 2
    ├── 01. Palais de Mari (1986).flac
    └── 02. Triadic Memories (1981).flac
```
where cover art is in the parent directory of audio files in every disc folder. It would be nice if DeaDBeeF can find cover art in the parent directory when there's no image file in the current folder.

PR #2883, opened 2022-11-06
Load artwork from parent folder

pappnu:
Allows using ".." as a search folder in the album artwork plugin to load art from a parent folder.

This feature was requested in issue #2847.


Oleksiy-Yakovenko:
I can't approve this:

* It removes any possibility of loading artwork from archives via VFS API
* It seems like it breaks finding artwork folders ignoring case -- "Covers" vs "covers" etc

PR #2896, opened 2022-11-26
Shuffle before play

pappnu:
Adds a new event that shuffles the current playlist and then plays the first track. Also modifies the playlist double click setting to optionally use the new event.

The current implementation seems to always play the first track in the playlist and afterwards proceeds in the shuffle order but I'd prefer for the shuffling to determine the initial track, so I made this.

If you'd like something to be changed let me know.

### DIFF
--- a/plugins/pltbrowser/pltbrowser.c
+++ b/plugins/pltbrowser/pltbrowser.c
@@ -786,8 +786,10 @@ on_pltbrowser_popup_menu (GtkWidget *widget, gpointer user_data) {
 
 static void
 on_pltbrowser_row_activated (GtkTreeView *tree_view, GtkTreePath *path, GtkTreeViewColumn *column, gpointer user_data) {
-    if (deadbeef->conf_get_int ("gtkui.pltbrowser.play_on_double_click", 1))
-        deadbeef->sendmessage (DB_EV_PLAY_NUM, 0, 0, 0);
+    if (deadbeef->conf_get_int ("gtkui.pltbrowser.play_on_double_click", 1)) {
+        deadbeef->sendmessage (DB_EV_STOP, 0, 0, 0);
+        deadbeef->sendmessage (DB_EV_NEXT, 0, 0, 0);
+    }
 }
 
 static void


Oleksiy-Yakovenko:
This new feature does almost the same thing as "Play random track" menu item.
Is this functionality not enough for your use case?
Did you consider changing the behavior of play random track to work this way when Playback Order is set to shuffle?

Oleksiy-Yakovenko:
Regarding the configurability of the playlist browser double click action -- I think it should be fine, but I think that there might be a way to do the same thing without introducing this new behavior.

pappnu:
Play random track fills my use case but I thought that it wouldn't suit the Albums shuffle mode, since in that case only the first tracks of albums should be possible candidates and I assume the "play radom track" event doesn't make that distinction.

Would it be okay to use the "play random track" event when double clicking if the shuffle mode is set to "Tracks" or "Play random tracks"? If I use the "play random track" event how can I tell the streamer to switch to the selected playlist? Should I modify the event's behaviour or is the functionality already exposed in the API?

Oleksiy-Yakovenko:
TBH I'm not completely sure what would be the best in that situation. Perhaps a new action "play random album" would make sense.

As far as I understand, you want to be able to double-click playlists, and when you do -- you expect a random track or random album to play (depending on Playback Order mode).

It does sound correct, that you would need one or more new streamer events to make this work (They are more of "streamer commands" than events anyway).

However, I don't think you need to make a new main thread events for that.. It's enough to have a new API function that would send the command to streamer, then call this function directly from the UI code.

pappnu:
Is this now in line with your suggested changes?

Oleksiy-Yakovenko:
I think that your whole problem can be solved in a different way.
First you need to expose the plt_reshuffle function to the API.
Then you can use this code.

Have you tried this?

Your approach -- it doesn't look correct to me. In the way you're doing it, I don't see how to make it reliable.

```
case ON_CLICK_SHUFFLE:
        plt = deadbeef->plt_get_curr ();
        deadbeef->plt_resuffle (plt, NULL, NULL);
        deadbeef->plt_unref (plt);
        deadbeef->sendmessage (DB_EV_STOP, 0, 0, 0);
        deadbeef->sendmessage (DB_EV_NEXT, 0, 0, 0);
        break;
```

pappnu:
With that specific approach, if I start a playlist and then double click another playlist deadbeef doesn't play the new playlist but instead reshuffles and plays the already playing one, so I think it is necessary to set the streamer playlist one way or another. Should I expose streamer_set_streamer_playlist to the API?

I also found it necessary to call streamer_set_last_played (NULL). Otherwise when double clicking the same playlist repeatedly while using album shuffle mode deadbeef would just alternate between the first two tracks of some album instead of playing one of the first tracks of any album in the playlist. I guess the reshuffling or the play_next actions take the previous track into account somehow and cause the aformentioned behaviour but since I'm not quite sure what causes it I don't know what is the best way to work around it. If it's acceptable to call streamer_set_last_played should I expose it too in the API and then make all the necessary calls in on_pltbrowser_row_activated?

Oleksiy-Yakovenko:
> With that specific approach, if I start a playlist and then double click another playlist deadbeef doesn't play the new playlist but instead plays the already playing one

I traced the code for `play_index` and `_handle_playback_stopped`, and they are not resetting the streamer_playlist to NULL when playback stops - which is the reason why you're observing this behaviour.

And here's where we have a problem:

1. If we reset streamer_playlist to NULL in response to DB_EV_STOP -- this will cause "continuity" problems, i.e. deadbeef will lose track of what's the next track to play after user presses "next" or "prev" after stopping playback.
2. If we don't reset -- then DB_EV_NEXT will use the previous streamer playlist to find the next track, which breaks your use case.

One option to solve this is to get a mechanism to force-reset the streamer_playlist to NULL, but I really don't want this to exist. Changing this variable from outside the streamer code is basically a bug.

I believe, a better option is to pass it as a flag to DB_EV_STOP, and pass the flag to the streamer, so that when this flag is set -- it will tell the _handle_playback_stopped function to reset the streamer_playlist.

in other words, try something like this:

1. In the main message handler
```
case DB_EV_STOP:
                    int full_stop = p1; // new flag to ask the streamer to reset streamer_playlist
                    streamer_stop_playback(full_stop); // new function -- see below
                    break;
```

2.
```
// new streamer signal:
STR_EV_STOP, // p1: whether to do a full stop, setting streamer_playlist to NULL
...

void streamer_stop_playback(int full_stop) {
    streamer_abort_files ();
    handler_reset (handler);
    handler_push (handler, STR_EV_STOP, 0, full_stop, 0); // hand over to the streamer thread
}
```

3.
```
// Handling the STR_EV_STOP: New function -- see below
case STR_EV_STOP:
                streamer_stop (p1, 0);
                break;

// extract the relevant code from play_index
static void
streamer_stop(int full_stop) {
    DB_output_t *output = plug_get_output ();
    playqueue_clear ();
    output->stop ();
    streamer_reset (1);
    viz_reset();
    viz_process(NULL, 0, output, 0, 0);

    streamer_lock();
    _handle_playback_stopped ();
    stream_track (NULL, 0);
    if (plt) {
        plt_unref (plt);
    }
    if (full_stop) {
         streamer_set_current_playlist_real(NULL); // <--------- here's where it should be done
    }
    streamer_unlock();
}

```

Oleksiy-Yakovenko:
If you prefer this -- I can apply this partial solution on my side, since this would be a useful cleanup anyway, and then you can try again. What do you think?

Oleksiy-Yakovenko:
I found a much smaller solution to this, trying it now.

Oleksiy-Yakovenko:
I just pushed a change.
This change seems fairly harmless to me -- but needs to be tested.
Now you should be able to do this:
```
        deadbeef->sendmessage (DB_EV_STOP, 0, 0, 0);
        deadbeef->sendmessage (DB_EV_NEXT, 0, 0, 0);
```

This would do exactly what you need, without calling any extra code

Oleksiy-Yakovenko:
One question that still needs to be answered -- whether we can call reshuffle in this case (automatically), or it breaks too much, and needs to be moved to be called by the UI code.

pappnu:
> ```
>         deadbeef->sendmessage (DB_EV_STOP, 0, 0, 0);
>         deadbeef->sendmessage (DB_EV_NEXT, 0, 0, 0);
> ```
> 
> This would do exactly what you need, without calling any extra code
This approach works but the problem with the album shuffle mode, that I mentioned in my previous comment, still exists.

Oleksiy-Yakovenko:
> This approach works but the problem with the album shuffle mode, that I mentioned in my previous comment, still exists.

Are you referring to the comment that it causes 2 alternating tracks to be triggered every time you double click a playlist?

Can you try to add this?
```
diff --git a/streamer.c b/streamer.c
index 5459820e6..936a5d276 100644
--- a/streamer.c
+++ b/streamer.c
@@ -2476,6 +2476,7 @@ play_index (int idx, int startpaused) {
     pl_lock ();
     if (plt != streamer_playlist) {
         streamer_set_streamer_playlist (plt);
+        streamer_set_last_played(NULL);
     }
     pl_unlock();
 
```

Oleksiy-Yakovenko:
I pushed one more commit that should take care of this as well

Oleksiy-Yakovenko:
Sorry for the swarm of commits -- I realized that the previous one was doing a wrong thing, so I did a revert and cleanup.
Can you see if this works? :)

pappnu:
Now it works just like I had hoped. I'll include only the pltbrowser changes to this pull request or do you want do those too on your end since they are so minimal? Do you think it's necessary to make the shuffled play on double click a new option or would it be okay to replace the current behaviour with it altogether. Setting shuffle to Off would still give the original behaviour.

Thank you for making these changes and guiding me along.

Oleksiy-Yakovenko:
I'm busy working on something else, so please finalize your UI changes :)

pappnu:
I replaced the old behaviour as I think most would expect the shuffling to apply to the first track, if shuffling is turned on. If you think this new approach should be optional, let me know and I'll make it so.

Oleksiy-Yakovenko:
> I replaced the old behaviour as I think most would expect the shuffling to apply to the first track, if shuffling is turned on. If you think this new approach should be optional, let me know and I'll make it so.

Thanks, I'm ok with this.

PR #2900, opened 2022-12-01
cocoaui: add "Show in Finder" context menu item

percontation:
In macOS, adds a "Show in Finder" track context menu item, that opens the file location of the track in Finder.app

I was missing this feature from other music players (e.g. iTunes, VLC, IINA)


Oleksiy-Yakovenko:
Works well, great addition :)
thanks

PR #2906, opened 2022-12-24
Feature/multithreaded converter

dakeryas:
* Add a configurable multi-threaded converter
* Minimal changes to the existing structures and interfaces, added new structures on top of the existing ones.
* Added a `convert3` method to be able to pass the `rw_lock` on the `abort` variable. The former calls the older `convert2` method.
* Re-used the existing-yet-commented GUI features for selecting the number of threads in the GUI dialogue.
* Looked for any memory leak (found an old one) using the ASan sanitiser. 
* Added some static functions for clarity.

### DIFF
--- a/plugins/converter/converter.c
+++ b/plugins/converter/converter.c
@@ -40,6 +40,7 @@
 #include <unistd.h>
 #include <inttypes.h>
 #include <errno.h>
+#include <pthread.h>
 #include "../../deadbeef.h"
 #include "converter.h"
 #include "../../strdupa.h"
@@ -855,8 +856,27 @@ static const unsigned char format_id_pcm[] = {
     0x80, 0x00, 0x00, 0xAA, 0x00, 0x38, 0x9B, 0x71
 };
 
+static int
+get_abort (int *pabort) {
+    return pabort ? *pabort : 1;
+}
+
+static int
+get_abort_lock (pthread_rwlock_t *abort_lock, int *pabort) {
+    int abort = 0;
+    if (abort_lock) {
+        pthread_rwlock_rdlock(abort_lock);
+        abort = get_abort (pabort);
+        pthread_rwlock_unlock(abort_lock);
+    }
+    else {
+        abort = get_abort (pabort);
+    }
+    return abort;
+}
+
 static int64_t
-_write_wav (DB_playItem_t *it, DB_decoder_t *dec, DB_fileinfo_t *fileinfo, ddb_dsp_preset_t *dsp_preset, ddb_encoder_preset_t *encoder_preset, int *abort, int fd, int output_bps, int output_is_float) {
+_write_wav (DB_playItem_t *it, DB_decoder_t *dec, DB_fileinfo_t *fileinfo, ddb_dsp_preset_t *dsp_preset, ddb_encoder_preset_t *encoder_preset, pthread_rwlock_t *abort_lock, int *abort, int fd, int output_bps, int output_is_float) {
     int64_t res = -1;
     char *buffer = NULL;
     char *dspbuffer = NULL;
@@ -885,7 +905,7 @@ _write_wav (DB_playItem_t *it, DB_decoder_t *dec, DB_fileinfo_t *fileinfo, ddb_d
         if (eof) {
             break;
         }
-        if (abort && *abort) {
+        if (get_abort_lock (abort_lock, abort)) {
             break;
         }
         int sz = dec->read (fileinfo, buffer, bs);
@@ -1343,7 +1363,7 @@ ddb_mktemp(char *template, size_t template_size, char *suffix) {
 }
 
 static int
-convert2 (ddb_converter_settings_t *settings, DB_playItem_t *it, const char *out, int *pabort) {
+convert3 (ddb_converter_settings_t *settings, DB_playItem_t *it, const char *out, pthread_rwlock_t *abort_lock, int *pabort) {
     int output_bps = settings->output_bps;
     int output_is_float = settings->output_is_float;
     ddb_encoder_preset_t *encoder_preset = settings->encoder_preset;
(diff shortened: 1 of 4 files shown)


Oleksiy-Yakovenko:
I will not accept a pull request that adds a new thread pool to the process instead of using libdispatch.
I understand that you put a lot of work into this, and if you wish so -- you are welcome to maintain your own fork.
Please see CONTRIBUTING.md:
> Before working on big changes, talk with the devs on Slack.
(well, now it's discord, but the concept doesn't change)

dakeryas:
Is there any chance you would reconsider, considering POSIX threads are standard and better supported on Linux and Windows? In addition, a good part of my work is already valid if you just change the threading model. 

Thank you for the Discord tips.

Oleksiy-Yakovenko:
It doesn't have anything to do which API is better or more standard. It's about avoiding thread explosion problem.

dakeryas:
Mmmmh, but in that case users have control when to start the converter, and how many threads to utilise. It's not really a background process issue, users are free to customise (reduce) the number of threads at any time should they encounter any issues. The current converter already uses a POSIX threads, so users might as well keep it to 1 thread if they so see fit.

Oleksiy-Yakovenko:
> The current converter already uses a POSIX thread

It uses pthread through an abstraction layer. That's a bit different.
(This allows e.g. to use a native win32 threads api on Windows -- even though it's not used right now, but it's possible).

However, none of this matters.
It's legacy code which eventually will be converted to libdispatch, just like all other code in Deadbeef.
Adding new code using old-school threads is not allowed in Deadbeef codebase.

dakeryas:
All right, I will try to move to libdispatch in the coming weeks, when I have some free time...

Oleksiy-Yakovenko:
> Mmmmh, but in that case users have control on when to start the converter, and how many threads to utilise.

Try to think what happens when a user runs multiple converters at the same time?

Oleksiy-Yakovenko:
And what happens if the user runs a converter with 8 threads, but there's another plugin (e.g. replaygain scanner) that's doing the same?

dakeryas:
> Try to think what happens when a user runs multiple converters at the same time?
Well, it's obvious that if you have 8 cores and run the convert another time with 8 threads, things will slow down. When you run something multiple times at maximum 'speed', you do expect a slowdown. The same if you run 100 tabs in three browsers...  Besides, it is the same with the current converter, you can open it 10 times even if you have 2 cores.

dakeryas:
> And what happens if the user runs a converter with 8 threads, but there's another plugin (e.g. replaygain scanner) that's doing the same?

Well, at least the converter *is* configurable, which is not the case of the ReplayGain scanner (which uses four hard-coded threads, spawning them again for every track without re-using the existing ones in a pool). I mean, if you know the ReplayGain scanner is a thread sucker, then you just don't start the converter with many threads. Again, it is nothing critical, and nothing that the users doesn't see or cannot customise. At the lowest level, the OS manages the threads.

Oleksiy-Yakovenko:
> Well, it's obvious that if you have 8 cores and run the convert another time with 8 threads, things will slow down. When you run something multiple times at maximum 'speed', you do expect a slowdown. The same if you run 100 tabs in three browsers... 

Well, this is not what I expect as a user from an application.
In the browser -- it's a bit different. I do kind of expect this, since each tab is basically a separate application (typically running in its own process). But I don't want each tab to run 100 threads.

Here's what I expect from an application:

The application should utilize all available threads / or as many threads as the user has configured.
If the number of needed threads exceeds this number -- the application should not attempt to run more threads, but instead it should work with what is configured.

> Besides, it is the same with the current converter, you can open it 10 times even if you have 2 cores.

Yes, exactly -- and this is why the entire codebase is being migrated to libdispatch. To solve that problem.

> I mean, if you know the ReplayGain scanner is a thread sucker, then you just don't start the converter with many threads at the same time. Again, it is nothing critical, and nothing that the users doesn't see or cannot customise. At the lowest level, the OS manages the threads.

I want both ReplayGain and Converter to utilize all available / configured threads, and I don't want them to run double of max thread count.

dakeryas:
> The application should utilize all available threads / or as many threads as the user has configured.
If the number of needed threads exceeds this number -- the application should not attempt to run more threads, but instead it should work with what is configured.

Agreed, in the ideal case. I would still argue that the proposed patch covers most user needs, and  is more convenient than running the converter n-times to achieve some sort or parallelism. :-)

> I want both ReplayGain and Converter to utilize all available / configured threads, and I don't want them to run double of max thread count.

That also sounds good, and is ideal. I would argue that something configurable in the meantime is already a step forward ! :-)

Oleksiy-Yakovenko:
> Agreed, in the ideal case. I would still argue that the proposed patch covers most user needs, and is more convenient than running the converter n-times to achieve some sort or parallelism. :-)

I never said that the intention is wrong. I'm only saying that using pthread is not acceptable.

> That also sounds good, and is ideal. I would argue that something configurable in the meantime is already a step forward ! :-)

It is a step forward in feature development, but a step back in repaying technical debt.

Oleksiy-Yakovenko:
I would say, the right way to do what you did is 2 steps:

1. convert the existing single-threaded converter code to libdispatch (which is quite trivial)
2. add concurrent flag to the queue, with a counted semaphore to limit thread count - which is probably 5 minutes of work

dakeryas:
Ideally, I would not lose all the current work... 

The GUI-update bits are annoying... And so is the ability to cancel all the converting threads at once upon clicking. Not sure how that pans out with libdispatch.

Oleksiy-Yakovenko:
@dakeryas I converted rg_scanner to use libdispatch, just so that you can see how simple it is.
You can see the diff here:
https://github.com/DeaDBeeF-Player/deadbeef/commit/4caa469476dc5489d16c58fd7c7bfeff72b96765

I didn't fully test is yet -- that's why it is on a branch, but it works in my minimal testing so far.

Oleksiy-Yakovenko:
also it has some weird variable names, since I copy-pasted some of the code from artwork plugin -- I'll clean it up

dakeryas:
Thanks, I will look at it in the coming days as I have never implemented anything with libdispatch.

PR #2938, opened 2023-04-10
Cast to long long when printing %lld

esainane:
On the linux CI system, the relevant format specifier for `int64_t` is `%ld`, rather than `%lld`. This generates some harmless warnings.

This PR changes current use of `%lld` with `int64_t` to use the portable `PRId64` constant instead, removing the warnings.

### DIFF
--- a/plugins/aac/aac.c
+++ b/plugins/aac/aac.c
@@ -28,7 +28,6 @@
 #include "../../config.h"
 #endif
 #include <stdlib.h>
-#include <inttypes.h>
 #include <math.h>
 #include <deadbeef/deadbeef.h>
 #include <deadbeef/strdupa.h>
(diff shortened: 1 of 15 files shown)


Oleksiy-Yakovenko:
@esainane hey, can you please rebase on master and rerun the builds?
Master was in a failing state on windows at the time when you submitted the PR -- I only got to fixing this failure today.

Oleksiy-Yakovenko:
tbh, I personally prefer a more readable solution to this problem, which is doing this:`"%lld", (long long)value`

esainane:
Done!

There are a few cases of `(type) expr` in the codebase, rather than all being `(type)expr`, but I've used the more frequently used `(type)expr` form also used in your comment. Let me know if you'd rather it be the other way!

PR #2945, opened 2023-05-04
Don't negate ignore_archives value when setting flags.

fudanchii:
I notice weirdness when adding folders with (.zip) archives, we now have the preference for ignoring archives when adding folders and it's active (value == 1) by default.

But what happened is actually the reverse, archives will added if this config is set. I tried to track this issue and found that the config passing for this is quite redundant:

The first assignment is correct since we set `plt->ignore_archives` directly from config
Screenshot at 2023-05-04 12-36-49

but then we call `_get_insert_file_flags_from_config` and do another assignment to `plt->ignore_archives` at `plt_insert_dir_int`, but this time by asserting the flags, effectively negating the correct value
Screenshot at 2023-05-04 12-37-31

notice the value for `plt->ignore_archives` changed from 1 to 0.

This patch fix this issue.

Thanks!


fudanchii:
After taking another look, it makes more sense why we negate the config value at `_get_insert_file_flags_from_config`, it's because the flag mask is named: `ENTER_ARCHIVES` which is the opposite of `ignore_archives` from the config, but if that's the case then the `0` and `1` here at `plt_insert_dir_int` will need to be flipped:
```
plt->ignore_archives = (flags&DDB_INSERT_FILE_ENTER_ARCHIVES) ? 0 : 1; // instead of ` 1 : 0`
```
Just to avoid further confusion I think the flag needs to be named `IGNORE_ARCHIVES` instead.

Updating this PR to reflect that.

Oleksiy-Yakovenko:
It's not possible to change the meaning and names of constants in deadbeef.h -- this may break existing plugins.
I don't think you really need to rename anything, it looks like the only thing you need to change is the evaluation of the flag meaning in the code which uses it.

Oleksiy-Yakovenko:
Wouldn't it be enough to just do this to resolve the issue?
```
-    plt->ignore_archives = (flags&DDB_INSERT_FILE_FLAG_ENTER_ARCHIVES) ? 1 : 0;
+    plt->ignore_archives = (flags&DDB_INSERT_FILE_FLAG_ENTER_ARCHIVES) ? 0 : 1;
```

fudanchii:
Yup, understood that the flag is part of public header, updated the PR just to flip the bit from flag checking.

PR #2946, opened 2023-05-06
adplug: Fix compile error with C++17

jaimeMF:
The "register" storage class specifier is not allowed


Oleksiy-Yakovenko:
There should not be a need for that, it should be enough to pass "-std=c++11" in compile flags.

jaimeMF:
> There should not be a need for that, it should be enough to pass "-std=c++11" in compile flags.

I have picked https://github.com/DeaDBeeF-Player/deadbeef/commit/ec70bec029c379ea17442c95b2ad346a0a68d9d0 and it builds fine on openSUSE Tumbleweed. Thank you.

PR #2952, opened 2023-05-22
Add bit depth override option.

HiFiPhile:
Add a bit depth selection box, similar to foobar.

Objective is to have a finer control of bit depth, which allows:
- Keep maximum resolution while doing volume math
- Lower resolution for testing/benchmark purpose

image

image

### DIFF
--- a/plugins/gtkui/callbacks.h
+++ b/plugins/gtkui/callbacks.h
@@ -1049,10 +1049,6 @@ on_convert16to24_toggled                (GtkToggleButton *togglebutton,
                                          gpointer       user_data);
 
 void
-on_combo_bit_override_changed                (GtkToggleButton *togglebutton,
-                                         gpointer       user_data);
-
-void
 on_reset_autostop_toggled              (GtkToggleButton *togglebutton,
                                         gpointer         user_data);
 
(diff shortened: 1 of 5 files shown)


Oleksiy-Yakovenko:
Overall looks fine, but before it can be merged -- the UI needs to be re-done using glade2,
since the interface.c file is generated by glade, and can't be modified directly.
You will need to use the version of glade from the tools folder in deadbeef git repo -- build as usual using configure / make / make install.

HiFiPhile:
Hi, I've updated the UI with glade.

Oleksiy-Yakovenko:
* Please set spacing of 8px between the elements in the new hbox
* Please change the titles to more readable "16 bit", "24 bit", "32 bit" etc -- without dash or capitalization
* I don't think the "32 bit fixed" is a thing in this context - it shoud be simply "32 bit"
* Changing this option repeatedly will crash pipewire plugin (when using pipewire output). Although it might be a pipewire bug, or pipewire plugin bug, or virtualbox bug.

HiFiPhile:
Update done.
> Changing this option repeatedly will crash pipewire plugin (when using pipewire output). Although it might be a pipewire bug, or pipewire plugin bug, or virtualbox bug.

I think it could be something else, since changing this option nearly does the same thing as check `always convert 16 bit input signal to 24 bit`

PR #2954, opened 2023-05-25
Add experimental DoP support.

HiFiPhile:
Based on @cqpwx's fork and https://github.com/marcoc1712/squeezelite-R2/blob/Release/dsd.c

Tested both DSF and DFF format.

To use it check `Enable DoP output` option in ffmpeg plugin.

### DIFF
--- a/plugins/alsa/alsa.c
+++ b/plugins/alsa/alsa.c
@@ -114,10 +114,10 @@ palsa_set_hw_params (ddb_waveformat_t *fmt) {
         // generic format
         plugin.fmt.bps = 16;
         plugin.fmt.is_float = 0;
-        plugin.fmt.is_dop = 0;
         plugin.fmt.channels = 2;
         plugin.fmt.samplerate = 44100;
         plugin.fmt.channelmask = 3;
+        plugin.fmt.flags &=~DDB_WAVEFORMAT_FLAG_IS_DOP;
     }
 
     snd_pcm_nonblock(audio, 0);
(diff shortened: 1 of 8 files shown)


Oleksiy-Yakovenko:
We can't add this `is_dop` field to the `ddb_waveformat_t` structure, since this would break plugin API.
(My fault, since at the time when this struct was added - I didn't make it extensible).

We'll have to figure out a way to pass this option either outside of this struct, or repurpose one of the existing fields for dual meaning.

(I didn't have time to think about it, so I don't have any ready solution at this point)

HiFiPhile:
I think it's possible to use `samplerate` to identify DSD, for example 2822400 for DSD64 and 5644800 for DSD128 etc.

Oleksiy-Yakovenko:
this would interfere with many things, I don't think samplerate is a good choice.
Perhaps the most feasible choice is the is_bigendian, which can be converted into a 32 bit flags field.
I don't think it was ever used for anything (in deadbeef itself). Basically, currently it is always zero.
It should be safe to even rename it into "flags", and then use 1 bit of that as DOP marker.

HiFiPhile:
Something like this ?
```
typedef struct {
    int bps;
    int channels;
    int samplerate;
    uint32_t channelmask;
    int is_float; // bps must be 32 if this is true
    struct 
    {
        int is_bigendian:1;
        int is_dop:1;
    };
} ddb_waveformat_t;
```

Oleksiy-Yakovenko:
It's best to kill the `is_bigendian` entirely, and remove all of its mentions from the code.
Replace it with `uint32_t flags`, and define the constant `DDB_WAVEFORMAT_FLAG_IS_DOP = 1`.
See how other similar constants are defined -- using an unnamed enum.

HiFiPhile:
I suppose increase `DDB_API_LEVEL` to 17 and wrap the enum inside ?

Oleksiy-Yakovenko:
Yes please

HiFiPhile:
API update done.

Oleksiy-Yakovenko:
I'll test it as soon as I can and will get back to you!

plugins/alsa/alsa.c
```diff
@@ -225,18 +241,28 @@ palsa_set_hw_params (ddb_waveformat_t *fmt) {
...
-    if ((err = snd_pcm_hw_params_set_rate_near (audio, hw_params, &val, &ret)) < 0) {
-        fprintf (stderr, "cannot set sample rate (%s)\n",
-                snd_strerror (err));
-        goto error;
+    if(plugin.fmt.flags & DDB_WAVEFORMAT_FLAG_IS_DOP) {
+        if ((err = snd_pcm_hw_params_set_rate (audio, hw_params, val, 0)) < 0) {
+            fprintf (stderr, "cannot set sample rate (%s)\n",
+                    snd_strerror (err));
+            goto error;
+        }
```
Oleksiy-Yakovenko, line 256: Wouldn't you want to take the returned rate here as well, and assign it to the format?
Or is it guaranteed to be the same?

Oleksiy-Yakovenko:
So, I'm not able to test this with DSD-capable DAC since I don't have one. But I tested all I could, and everything seems to work fine - merging!

plugins/alsa/alsa.c
HiFiPhile: `snd_pcm_hw_params_set_rate()` with `dir` set to 0 will use exactly the same rate.

PR #2957, opened 2023-06-12
configure: Use AC_SYS_LARGEFILE

orbea:
With musl-1.2.4 LFS support is being disabled and will later be entirely removed where it is expected to use off_t and lseek rather than off64_t and lseek64.

With glibc these types are also obsolete where it is expected to use the _FILE_OFFSET_BITS define where off_t and lseek will become 64-bit compatible versions when needed. This is set by AC_SYS_LARGEFILE.


Oleksiy-Yakovenko:
Merging this would break compatibility with 32 bit systems, where lseek is 32 bit.

orbea:
@Oleksiy-Yakovenko Are you saying deadbeef needs a 64-bit lseek even on 32-bit systems? Or am I not understanding?

Oleksiy-Yakovenko:
Yes, lseek64 works on 32 bit systems to seek in large files which are larger than 4GB

orbea:
Perhaps it would be better to define _FILE_OFFSET_BITS=64 unconditionally?

Oleksiy-Yakovenko:
I checked the docs, and it seems like adding AC_SYS_LARGEFILE should take care of 64 bit file offsets on 32 bit systems as well, but it needs to be tested before merging. (I'll try to find time for this)

orbea:
@Oleksiy-Yakovenko Any chance you had time to investigate this yet?

Oleksiy-Yakovenko:
I haven't, and I'm not sure whether I will. Let's assume that it works as it should. There are not that many 32bit users, if they will complain we'll revert.

PR #2977, opened 2023-08-05
package: ship common.h, too

rsekman:


Oleksiy-Yakovenko:
This change doesn't seem to make any sense. You can't really use this header from plugins -- so you don't need to install it as an SDK header.

rsekman:
The tests in `Tests/` include this file with `<>`. So either those need to be changed to relative imports or the header shipped.
https://github.com/DeaDBeeF-Player/deadbeef/blob/9d299f395f11176648fa756e024cd2d674f502aa/Tests/MP3DecoderTests.cpp#L10

Oleksiy-Yakovenko:
The tests need to run in the source tree, they don't / should not require deadbeef sdk headers to be installed.
Can you show how you run tests and what kind of errors are you getting?

Oleksiy-Yakovenko:
FYI: to run the tests you simply need to run `./scripts/test.sh` -- it has all the necessary compiler options to make sure the headers will be found.

Oleksiy-Yakovenko:
When running the script, it is expected that static-deps folder is present in the root of source tree.
You can get this automatically setup by running `travis/download-linux-static-deps.sh`

rsekman:
I was trying to write a new test and my editor complained about missing
includes.. I guess I need to work relative to the root directory.
Thanks.

PR #2978, opened 2023-08-05
tf: implement and test $get(), $put(), and $puts()

rsekman:


src/tf.c
```diff
@@ -76,13 +76,88 @@
 #define TEMP_BUFFER_SIZE 1000
 #define TF_INTERNAL_FLAG_LOCKED (1<<16)
 
+typedef struct {
+    // NULL-terminated C strings
+    char* key;
```
Oleksiy-Yakovenko, line 81: Deadbeef uses coding convention where the `*` is placed on the right hand side, so please correct in the new code -- it should look like this `char *key` etc

src/tf.c
```diff
@@ -76,13 +76,88 @@
...
+    char* value;
+} ddb_tf_var_t;
+
+typedef struct ddb_tf_vars_s {
+    struct ddb_tf_vars_s* next;
+    ddb_tf_var_t var;
+} ddb_tf_vars_t;
+
+ddb_tf_var_t
+tf_var_new(const char* key, const char* value){
```
Oleksiy-Yakovenko, line 91: It would be more explicit and more performant to convert this code to be initializer style, i.e. pass a pointer to the struct as argument, and initialize it there.
Then the calls would like this: `tf_var_init(&var, key, value)`, and there would not be unnecessary copying.

src/tf.c
```diff
@@ -76,13 +76,88 @@
...
+    out.key = malloc(strlen(key) + 1);
+    out.value = malloc(strlen(value) + 1);
+    strcpy(out.key, key);
+    strcpy(out.value, value);
+
+    return out;
+}
+
+void
+tf_var_free(ddb_tf_var_t var) {
```
Oleksiy-Yakovenko, line 103: Same here, this function looks like it would free an object, but really -- this is a "deinitializer" (also known as destructor), it is preferred to call these "deinit: `tf_var_deinit(&var)`

src/tf.c
```diff
@@ -76,13 +76,88 @@
...
+    free(var.key);
+    free(var.value);
+}
+
+char*
```
Oleksiy-Yakovenko, line 108: This function should be returning `const char *`

src/tf.c
```diff
@@ -76,13 +76,88 @@
...
+tf_vars_lookup(ddb_tf_vars_t* vars, const char* key) {
+    char* tmpkey = malloc(strlen(key)+1);
```
Oleksiy-Yakovenko, line 110: I don't see any reason for this tmpkey variable

src/tf.c
```diff
@@ -76,13 +76,88 @@
...
+    strcpy(tmpkey, key);
+    while (vars != NULL) {
+        if (!u8_strcasecmp (tmpkey, vars->var.key)) {
```
Oleksiy-Yakovenko, line 113: According to foobar2000 specification, the variables are case-sensitive, so you should be using strcmp instead.

src/tf.c
```diff
@@ -76,13 +76,88 @@
...
+    while (cur != NULL) {
+        if (!u8_strcasecmp (key, cur->var.key)) {
+            cur->var.value = realloc(cur->var.value, strlen(value)+1);
+            strcpy(cur->var.value, value);
+            return cur;
+        }
+        cur = cur->next;
+    }
+
+    ddb_tf_vars_t* out = malloc(sizeof(ddb_tf_var_t));
```
Oleksiy-Yakovenko, line 136: It's preferred to use calloc, to automatically initialize the allocated memory with zeroes

src/tf.c
```diff
@@ -76,13 +76,88 @@
...
+    out->next = vars;
+    out->var = tf_var_new(key, value);
+
+    return out;
+}
+
+void
+tf_vars_free(ddb_tf_vars_t* vars) {
+    if (vars != NULL) {
+        tf_vars_free(vars->next);
```
Oleksiy-Yakovenko, line 146: It's completely unnecessary to use recursion here.
It's less efficient, harder to read, and potentially may cause stack overflow if the list is large enough.

Oleksiy-Yakovenko (changes requested):
First round of review:
Overall makes sense, so no big changes requested - mostly stylistic.
After the code is cleaned up I'll do a 2nd round + testing / validation.

src/tf.c
Oleksiy-Yakovenko: sorry, I got confused and misread the spec. Please ignore this comment.

src/tf.c
```diff
@@ -2507,6 +2584,66 @@ tf_func_itematindex (ddb_tf_context_t *ctx, int argc, const uint16_t *arglens, c
...
+    int len;
+    char temp_key[TEMP_BUFFER_SIZE];
+    char temp_value[TEMP_BUFFER_SIZE];
+    TF_EVAL_CHECK(len, ctx, args,            arglens[0], temp_key,   sizeof (temp_key)   - 1, fail_on_undef);
+    TF_EVAL_CHECK(len, ctx, args+arglens[0], arglens[1], temp_value, sizeof (temp_value) - 1, fail_on_undef);
+
+    ddb_tf_context_int_t *priv = (ddb_tf_context_int_t *)ctx;
+    priv->vars = tf_vars_set(priv->vars, temp_key, temp_value);
+    if (should_return) {
+        char* value = tf_vars_lookup(priv->vars, temp_key);
```
Oleksiy-Yakovenko, line 2630: This line causes "unused variable" compiler warning

src/tf.c
```diff
@@ -2507,6 +2584,66 @@ tf_func_itematindex (ddb_tf_context_t *ctx, int argc, const uint16_t *arglens, c
...
+        *out = '\0';
+        return 0;
+    }
+    len = u8_strnbcpy(out, value, outlen);
+
+    return len;
+}
+
+int
+tf_func_put_ (ddb_tf_context_t *ctx, int argc, const uint16_t *arglens, const char *args, char *out, int outlen, int fail_on_undef, int should_return) {
```
Oleksiy-Yakovenko, line 2614: It's non-conventional in deadbeef to have `_` suffix in function names. What does it mean here?

Oleksiy-Yakovenko: (it seems like it's an internal / private function. In this case you need to add `static` to it, and use prefix `_` instead of suffix.

rsekman: The only difference between `$put` and `$puts` is whether the stored value should be returned. So it is DRY to to have an internal function that takes an extra `bool should_return`. Fixed.

src/tf.c
rsekman: It is also superfluous. Fixed.

src/tf.c
rsekman: Mistake left over, fixed.

src/tf.c
rsekman: You mean copying of the struct, not the strings, right? Because the strings are going to be read into buffers of size `TEMP_BUFFER_SIZE` on the stack. The buffers could, of course, be put on the heap. But we can't know how long they are before evaluating them, so we should shrink to fit with `realloc` anyway... this should not be the responsibility of the caller. So keep `strcpy` in the body, yes?

Oleksiy-Yakovenko: I mean this:

```
void
 tf_var_init(ddb_tf_var_t *var, const char* key, const char* value) {
    var->key = strdup(key);
    var->value = strdup(value);
}
```

Oleksiy-Yakovenko:
@rsekman hey, are you planning to finalize the PR?
It seems like it's failing on some unit tests, which use too little buffer space (8 bytes). You should replace all those buffer size parameters with `sizeof(buffer)` to fix it.
However, the tests also don't finish even if I fix this, and they cause memory corruption.
I think your linked list code is not right.

Oleksiy-Yakovenko:
(I'll try to fix it on my own, since I think explaining what needs to be fixed will take longer time)

rsekman:
Hmm, the tests pass on Linux but not on Mac. I don't know why that would be and don't have a Mac to test with. UTF-8 vs UTF-16?

Oleksiy-Yakovenko:
I have already refactored / fixed / tested / merged everything, so nothing left to do - unless you find some more issues.

PR #2981, opened 2023-08-08
Remove the register keyword in C++ code

xuzhen:
The `register` keyword was removed in the C++17 standard [1], and Clang 16 builds C++ code according to the C++17 standard by default [2].

This patch removed the register keyword in C++ codes, to make DeaDBeeF can be built using Clang 16.

[1] https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/p0001r1.html
[2] https://clang.llvm.org/cxx_status.html#cxx17


xuzhen:
https://github.com/DeaDBeeF-Player/deadbeef/commit/ec70bec029c379ea17442c95b2ad346a0a68d9d0 set standard to C++11, no need this patch

PR #2998, opened 2023-09-29
tests: propagate exit code from running tests properly

rsekman:
246e3b1 introduced a bug in the test script. If the test suite fails the script still continues, restores the cpp libs that were moved out of the `static-deps` folder, and exits with the exit code of that move. This hides failing or crashing tests from CI. Oops. Luckily it's easily fixed.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -66,5 +66,7 @@ if [ $code -ne 0 ]; then
 fi
 
 "$BUILD/runtests"
+code=$?
 
 restore_cpp_libs
+exit $code


Oleksiy-Yakovenko:
Did you get a callstack of the crash? This PR doesn't solve the issue - instead, it makes the tests fail on CI for linux.
(Which, of course, it the right thing - but wouldn't you want to fix the crash before merging?)

Oleksiy-Yakovenko:
nvm, I see the callstack in the relevant issue now. will take a look into it...

PR #3007, opened 2023-10-09
feat: commands for next/previous album that respect shuffle settings

rsekman:
# Feature

This pull request implements commands that can be bound to hotkeys: `Playback/Play Next Album`,  `Playback/Play Previous Album`, `Playback/Play Random Album`. The latter goes to first track of a random album in the current playlist. The behaviour of the first two depends on the shuffle setting and is as follows.
* `DDB_SHUFFLE_RANDOM`, `DDB_SHUFFLE_TRACKS`: same as  Play Next Track and Play Previous Track.
* `DDB_SHUFFLE_OFF`, `DDB_SHUFFLE_ALBUMS`:
*  * Play Next Album is the same as calling Play Next Track until reaching a new album.
* * If *not* on the first track of the current album, Play Previous Track goes to the first track on the current album
* * If on the first track of the current album, Play Previous Track is the same as calling Play Previous Track until reaching the first track of the previous album
* * The reasoning behind this is mimicking pressing Back once to go the beginning of the current track and twice to go to the beginning of the previous track.

# Implementation

While Play (Next/Previous) Album mimic repeated calls to Play (Next/Previous) Track, they are implemented with separate functions to avoid having to search the playlist repeatedly. This would be `O(#tracks on album * #tracks in playlist)` in `DDB_SHUFFLE_ALBUMS`; now they are `O(#tracks in playlist)`.

Thus the pull request consists of 4 commits, viz., in order:
1. Preparatory refactoring checking if two playlist items belong to the same album into a function. This was already implemented in three places in three different ways.
2. Preparatory refactoring breaking searching the playlist for minimal notplayed or maximal played out into some helper functions.
3. Implementation of the feature and a test suite. Note that 2/3 of the new code is in tests as there are quite a lot of cases to test.
4. Trivial linting.

# To discuss

I've tried to follow DeaDBeeF's style as best as I understand it. :^) But of course I'll be happy to lint before merging if I haven't been fully consistent.

Because the pull request introduces new signals `DB_EV_PLAY_(NEXT|PREV|RANDOM)_ALBUM`  I think we need to bump the API version before merging. I've left it until then because I thought it best to receive signoff first.

### DIFF
--- a/Tests/AlbumNavigationTests.cpp
+++ b/Tests/AlbumNavigationTests.cpp
@@ -1,0 +1,678 @@
+//
+//  AlbumNavigationTests.m
+//  Tests
+//
+//  Created by Robin Ekman on 05/10/23.
+//  Copyright © 2023 Robin Ekman. All rights reserved.
+//
+
+#include "conf.h"
+#include "logger.h"
+#include <deadbeef/common.h>
+#include <deadbeef/deadbeef.h>
+#include "conf.h"
+#include "fakeout.h"
+#include "messagepump.h"
+#include "playlist.h"
+#include "playmodes.h"
+#include "plmeta.h"
+#include "plugins.h"
+#include "streamer.h"
+#include "threading.h"
+#include <stdio.h>
+#include <gtest/gtest.h>
+
+extern "C" DB_plugin_t * fakeout_load (DB_functions_t *api);
+// super oversimplified mainloop
+static void
+mainloop (void *ctx) {
+    for (;;) {
+        uint32_t msg;
+        uintptr_t ctx;
+        uint32_t p1;
+        uint32_t p2;
+        int term = 0;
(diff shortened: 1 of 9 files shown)


Oleksiy-Yakovenko:
You don't necessarily need to bump API version, since there's still a possibility to get onto 1.9.6 train.
But you need to add version check to the new enum cases.

Oleksiy-Yakovenko:
The main concern for me is whether the function `pl_items_from_same_album` does the same thing as the code was doing before. I'll need to carefully read and understand the code to figure this out.

rsekman:
> The main concern for me is whether the function `pl_items_from_same_album` does the same thing as the code was doing before. I'll need to carefully read and understand the code to figure this out.

The body is the same as the check in `streamer.c:stop_after_album_check`. Other than in the new code it's called in three places: `stop_after_album_check, plt_insert_item, plt_reshuffle`. I could write tests for the latter two?

PR #3073, opened 2024-04-04
Update docs

dark-penguin:
- Encountering the phrase "See build instructions in the README file" while reading the README file is confusing, as is two README files. I've considered renaming README to BUILDING, but on second thought, it would make more sense to just put the *nix build instructions right there in README.md, along with Windows/macOS instructions.
- Specify correct Debian/Ubuntu package names for dependencies to make things easier
- Fix wrong paths in outdated Docker build instructions (but the instructions don't work for me anyway)


README.md
```diff
@@ -29,13 +29,54 @@ If you wish to chat with developers, join us on [Discord](https://discord.gg/GTV
...
+libpulse-dev - for PulseAudio support
+libfaad-dev - for AAC support via libFAAD
+zlib1g-dev - for psf/psf2/vgz support
+libzip-dev - for vfs_zip support
+libpipewire-0.3-dev - for Pipewire support
+libnotify-dev - for notification popups
+libopusfile-dev - for Opus support
+libcdio-dev, libcddb2-dev, libcdio19 - for AudioCD support
+```
+There is also a GTK2 interface that you can use instead of GTK3, but GTK2 has been dropped from Debian/Ubuntu since a long time ago.
```
Oleksiy-Yakovenko, line 70: This statement is highly opinionated, and technically is not true. From the last time I tried using Ubuntu, GTK2 was available, and worked just fine.

tools/docker/README.md
```diff
@@ -1,15 +1,15 @@
 # Building portable build using docker
 
-1. Install and run docker
-2. Make sure current directory is the root of the repo
-3. Run once: `./tools/docker-build/docker-bootstrap.sh`
-4. For each build: `./tools/docker-build/docker-build.sh`
-5. Find build artifacts in docker-artifacts folder
-
-To speed up the builds, make sure to download and unpack static-deps. Otherwise, docker-build.sh will do that automatically for each build.
+1. Make sure you have `docker` installed
```
Oleksiy-Yakovenko, line 3: This approach is not something that should be recommended to the users.
It's an script that I'm using for quick testing that deadbeef is building on Ubuntu 18.04, but there's no practical reason to use this script outside of this scope.

Oleksiy-Yakovenko (changes requested):
In summary:

I don't like this change.

It goes into too much details about how to build deadbeef on one specific Linux distribution (Debian), and this sounds more like a wiki page material, not the main project's README.md file.

dark-penguin:
I am all for the "more broadly applicable rather than highly specific" approach! However, I believe this change *is* more broadly applicable. It's not about Debian the distro - it's about all DEB-based distros, which is likely at least about half the users. And consider this:
- "Install Pipewire" - all users have to wonder which package exactly (Pipewire? You mean the library or the binary package?)
- "Install libpipewire-0.3-dev" - some users are already done, and the rest of them go "OK, so we need the headers for the pipewire library - I know what that should be called in my distro, BTW v0.3 should be good enough".

I would definitely prefer detailed instructions about "this is how I did it 15 years ago on RHEL5" rather than abstract "configure your xxx to do yyy, you'll figure it out for your distro". And the only "specific" part here is package names - if anyone provides a tested list of packages for another distro, it would only take one to add them all.

If you are still not convinced, then how about we keep the "generic" library names for the big list ("GTK3 - for GTK3 UI support"), but also have one line of "So, the complete list for most DEB-based distros (tested on Ubuntu 22.04) is: ... " ? But I think there is no clear benefit at all in providing *less* specific information that is already included in the *more* specific one.

Oleksiy-Yakovenko:
Yes, it has to be specific for each distro. There used to be a wiki page on that (and for more distros, not just Ubuntu). I don't know where did it go.. 
I created a new one and linked it from README.md, so for the time being it's going to be enough.
I don't want the README.md to become a Linux dependency database, which is why I strongly insist to keeping this kind of documentation in the wiki.

dark-penguin:
Oh, I see how it's done for Windows - there's a link to the wiki page with specific deps! I agree that my Linux build section is too large to be included in the main readme. But maybe delete the "other readme" that actually only contains build instructions? Or at least rename it to "BUILDING", and maybe put distro-specific instructions there if you want it to be in the project itself rather than in wiki?

README.md
dark-penguin: I haven't based it on conjecture. I was unable to find the required GTK2 package in Ubuntu 22.04 , that's the only reason I've left one item from this list without a clear result. But this was because I'm an idiot; I've found it now, it's `libgtk2.0-dev`!

tools/docker/README.md
dark-penguin: It's just that there are wrong paths there. I've figured out what you meant and corrected them.

dark-penguin:
I see the wiki page linked in README.md, however it is not listed in the list of wiki pages on the right. :slightly_smiling_face:

Oleksiy-Yakovenko:
> Or at least rename it to "BUILDING", and maybe put distro-specific instructions there if you want it to be in the project itself rather than in wiki?

I don't want to have any distro-specific instructions written in files, since this is the kind of information that becomes outdates super quickly, and wiki is the perfect place for that.

Renaming makes sense though, I'll do it.

PR #3076, opened 2024-04-09
plugins: cdda: cdtext: Try reading magic track 0 first for album artist / title

Warepire:
Since cdio API version 6 (introduced in 2012) the magic track index 0 represents the album wide CD-TEXT. Trying to read album artist / title from the track_nr index will result in the track artist / track title being selected.

Having the Artist - Album playlist column enabled and relying on CD-TEXT would previously display the track title instead of the album title after the dash.

API ref: http://git.savannah.gnu.org/cgit/libcdio.git/tree/include/cdio/cdtext.h#n305

Example using the Memory Garden - 1349 CD with CD-Text, added to playlist before and after fix:

deadbeef-cdtext-before-after


Oleksiy-Yakovenko:
I have to blindly trust this one, since I can't test or validate in any way, but it looks legit :)

PR #3101, opened 2024-06-22
ffmpeg: update API to support FFMPEG 7

kode54:
Added some const attributes here and there to satisfy some warnings, conditionally used new structures for the channel count based on an API deprecated long enough ago to be dropped in 7.0. Maybe you'll want to restyle it to your ideal, if you prefer.


Oleksiy-Yakovenko:
Thanks for that, really good job.

PR #3115, opened 2024-08-24
support relative paths when loading DBPL playlists

rsekman:
related to: https://github.com/DeaDBeeF-Player/deadbeef/issues/1798

when loading an M3U playlist deadbeef supports interpreting the paths relative to the location of the playlist file. this feature is currently missing from the code loading native (DBPL) playlists, but is implemented in this pull request.

this PR will not pass CI due to a segmentation fault in the Playlist test suite. the segmentation fault occurs only if the Playlist test suite is not run first. that is, 
```sh
./scripts/test.sh Playlist AlbumNavigation #passes
./scripts/test.sh AlbumNavigation Playlist #segfaults
```
the segfault occurs here, when  checking if loading the playlist should be handled by a plugin:
https://github.com/DeaDBeeF-Player/deadbeef/blob/a50e3a54261abcfcd8a4e627edc3a8b94ace6aff/src/playlist.c#L2707-L2711
if the Playlist test suite is run first, the array `plug` is empty (`plug[0] == NULL`), but if not it contains an invalid pointer. I do not understand why this happens and don't know how to trace it further.


src/playlist.c
```diff
@@ -2430,14 +2430,21 @@ _plt_load_from_file (playlist_t *plt, ddb_file_handle_t *fp, playItem_t **last_a
     if (ddb_file_read (&cnt, 1, 4, fp) != 4) {
         goto load_fail;
     }
-    
+
+    char dname[PATH_MAX] = "";
```
Oleksiy-Yakovenko, line 2434: It's better to dynamically allocate this one, since stack can get fairly deep with unpredictable memory needs.

src/playlist.c
```diff
@@ -2430,14 +2430,21 @@ _plt_load_from_file (playlist_t *plt, ddb_file_handle_t *fp, playItem_t **last_a
...
+    }
+
     for (uint32_t i = 0; i < cnt; i++) {
         it = pl_item_alloc ();
         if (!it) {
             goto load_fail;
         }
         uint16_t l;
         int16_t tracknum = 0;
+        char true_uri[PATH_MAX];
```
Oleksiy-Yakovenko, line 2447: This should be moved to the scope where it's used, to avoid unnecessary stack memory consumption after this variable is not needed. Also, it's safer to dynamically allocate it.

src/playlist.c
```diff
@@ -2594,7 +2607,9 @@ _plt_load_from_file (playlist_t *plt, ddb_file_handle_t *fp, playItem_t **last_a
                         // some values are stored twice:
                         // once in legacy format, and once in metadata format
                         // here, we delete what was set from legacy, and overwrite with metadata
-                        pl_replace_meta (it, key, value);
+                        if (strcmp (key, ":URI") != 0) {
```
Oleksiy-Yakovenko, line 2610: Why is this change necessary?

Oleksiy-Yakovenko (changes requested):

Oleksiy-Yakovenko:
> this PR will not pass CI due to a segmentation fault in the Playlist test suite. the segmentation fault occurs only if the Playlist test suite is not run first. that is,

This seems like some sort of test setup problem. I'm sure I'll be able to figure it out, but I'm a bit time constrained in the next few weeks.
I'll try to debug it as soon as I can.

src/playlist.c
rsekman: Otherwise a stored `:URI` value will overwrite the value read and interpreted as a relative path before. But it would be better to process this value the same way, I suppose.

src/playlist.c
rsekman: Resolved by factoring the subsequent logic into a function which returns a pointer to a dynamically allocated string; it is the caller's responsibility to `free()` it. This function can then be reused to address the next comment.

src/playlist.c
Oleksiy-Yakovenko: ok, I understand a bit better now.. I misread what the if condition was originally.

src/playlist.c
```diff
@@ -2655,6 +2677,10 @@ _plt_load_from_file (playlist_t *plt, ddb_file_handle_t *fp, playItem_t **last_a
         pl_item_unref (it);
         it = NULL;
     }
+    if (dname) {
```
Oleksiy-Yakovenko, line 2680: nit: this "if" is unnecessary, the `free` will do this check automatically.

rsekman: Fixed; also fixed that `goto`:s could lead to reaching this without `dname` being initialised. (Caught by the Mac CI but passed on Linux, must be some compiler difference?)

Oleksiy-Yakovenko:
I fixed the issue with tests and merged everything to master.

PR #3118, opened 2024-09-01
formatting: run clang-format

rsekman:
Many files in `src/` and `Tests/` were not formatted consistently according to the `.clang-format` in the repo. Perhaps this is because the `.clang-format` contained a duplicate entry that would cause clang-format to crash. This pull requests simply run
```
clang-format -i src/*.h src/*.c Tests/*.c{pp,} Tests/*.h
```

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -3,7 +3,7 @@
 ---
 AccessModifierOffset: -4
 AlignAfterOpenBracket: AlwaysBreak
-AlignArrayOfStructures: None
+AlignArrayOfStructures: Left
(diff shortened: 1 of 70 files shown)


rsekman:
I understand.

Just FYI -- it is possible to exclude particular files with a [`.clang-format-ignore` file](https://clang.llvm.org/docs/ClangFormat.html#clang-format-ignore) and even parts of files from being reformatted by
```
// clang-format off: this code will not be reformatted
...
// clang-format on: reformatting resumes here
```

PR #3119, opened 2024-09-01
Tests: add .gitignore

rsekman:
running the test suite creates the empty file Tests/PresetManagerData/dspconfig, which should be ignored

### DIFF
--- a/Tests/.gitignore
+++ b/Tests/.gitignore
@@ -1,0 +1,1 @@
+PresetManagerData/dspconfig


Oleksiy-Yakovenko:
IMO this fixes a symptom instead of the root cause.
The tests should not be writing there.
I probably just haven't noticed this, since I almost always run tests via Xcode where this problem isn't present.
The correct solution is to copy test assets to some temporary directory, and then execute tests there.
Even better is to provide a mechanism to prevent the code from from writing this config file during tests.

Oleksiy-Yakovenko:
I applied another fix, merged to master. Please let me know if it works for you.

PR #3129, opened 2024-09-16
adplug: update upstream library and add OPL synth options

thomasjepp:
Hi!

I was trying to play back some CMF files using DeaDBeeF and hit some audio quality issues with the adplug plugin.

To resolve, I've updated the plugin to the latest upstream - I've not looked too hard at the upstream changes, they're quite substantial - and added support for several more synth types. I've set the default to Nuked OPL3, which seems to be the highest quality/most accurate. It looks like the updated upstream library supports some extra file formats, but I've not enabled these as I don't have any example files to confirm they're functional.

From my checking this looks okay still for LGPL v2.1 - there are options to use the MAME upstream OPL plugin that *would* affect the license, so I've avoided this.

A full list of changes is below:

* Updated adplug library to https://github.com/adplug/adplug/commit/a6706ba256976eedf73e06d67ad74c5985271c1f (current master as of 2024-09-16)
* Removed surround option - changed to automatically apply based on synth type
* Set default synth sample rate to the same as a real OPL chip
* Added support for more synth types
* Set the default synth to Nuked OPL3

Please let me know if you have any comments or feedback - my C is a little rusty!


thomasjepp:
[libbinio](https://github.com/adplug/libbinio) might need updating, but seems functional for me

I've only tested building this on macOS - not on Linux or Windows.

Oleksiy-Yakovenko:
One important thing that's easy to overlook (and hard to do) is to see whether all patches that were applied on top of previous version of adplug need to be re-applied after updating.
If they are not applied - some bugs will return, and some optimizations will be lost.

dmitrysmagin:
Come on, give this patch a go! Adplug was updated with new music formats and bug fixes.
Most probably, old optimizations are no longer relevant.

Oleksiy-Yakovenko:
I'm temporarily unavailable for large patches. I'll have a good look in about 1-2 weeks.

Oleksiy-Yakovenko:
I'll merge it from another branch, where I'm applying deadbeef patches.

Oleksiy-Yakovenko:
Merged

thomasjepp:
Apologies for not testing on Linux & rebasing deadbeef's patches, real life caught up!

I appreciate you merging this though, thanks!

Issue #3130, opened 2024-09-16
[Mac] Segfault crash when drag and dropping files into playlist from Media Library


carlmylo:
### Steps to reproduce the problem
1. Open DeaDBeeF (through Terminal or normally)
2. Drag items into Playlist from the Media LIbrary
3. Crash

https://github.com/user-attachments/assets/cf9cc592-2e90-441d-b52f-59196f707cb8

### What's going on? Describe the problem in as much detail as possible.
I recently updated and tried to refresh the main playlist since I use that to view all of my library at once.
Dragging and dropping files into the playlist will cause a crash. I've tried adding just a single song, a single album, and a single genre and they all crash.

Adding items in from Finder works fine.

### Information about the software:
devel (6ef4bad)

Deadbeef version:
OS: macOS Monterey 12.7.6 (21H1320)

Below is a Mac .ips file. I had to change the extension to .txt since Github wouldn't let me upload it otherwise but changing it back to .ips should let it open up in the Console application.

[DeaDBeeF-2024-09-16-143238.txt](https://github.com/user-attachments/files/17018940/DeaDBeeF-2024-09-16-143238.txt)

Oleksiy-Yakovenko:
Fixed by PR #3131

Oleksiy-Yakovenko:
I hope to have some time to look into this next week

PR #3131, opened 2024-09-17
fix plt_load_from_buffer

percontation:
plt_load_from_buffer calls _plt_load_from_file with fname=NULL, which basically just unconditionally crashes.

Anyway, drag-and-drop from the sidebar into a playlist was broken on macOS beacause of this. Seems to be fine with this fix.


percontation:
Oh, I just saw #3130, presumably this PR fixes that issue.

Oleksiy-Yakovenko:
Thanks!

PR #3192, opened 2025-05-28
Opus artwork

eko5z:
Beforehand, the Ogg artwork decoder was supposed to get artwork from Opus files, but this always resulted in an error on `ov_open_callbacks` (OV_ENOTVORBIS). Thus, Opus files with embedded artwork did not show any image in DeaDBeeF. This PR creates new decoders, mostly copied from the already existing Ogg ones, and now Opus files with embedded cover art work fine.

Before:
Zrzut ekranu z 2025-05-28 17-17-49

After:
Zrzut ekranu z 2025-05-28 17-18-24

### DIFF
--- a/plugins/artwork/Makefile.am
+++ b/plugins/artwork/Makefile.am
@@ -12,7 +12,7 @@ endif
 if HAVE_OPUS
 opus_def = -DUSE_OPUS=1
 opus_libs = $(OPUS_LIBS)
-opus_sources = artwork_opus.c artwork_ogg.h
+opus_sources = artwork_opus.c artwork_opus.h
 endif
 
 if ARTWORK_USE_VFS_CURL
--- a/plugins/artwork/Makefile.am
+++ b/plugins/artwork/Makefile.am
@@ -9,6 +9,12 @@ ogg_sources = base64.c base64.h artwork_ogg.c artwork_ogg.h
 endif
 endif
 
+if HAVE_OPUS
+opus_def = -DUSE_OPUS=1
+opus_libs = $(OPUS_LIBS)
+opus_sources = artwork_opus.c artwork_ogg.h
+endif
+
 if ARTWORK_USE_VFS_CURL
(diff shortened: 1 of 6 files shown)


plugins/artwork/Makefile.am
```diff
@@ -9,6 +9,12 @@ ogg_sources = base64.c base64.h artwork_ogg.c artwork_ogg.h
 endif
 endif
 
+if HAVE_OPUS
+opus_def = -DUSE_OPUS=1
+opus_libs = $(OPUS_LIBS)
+opus_sources = artwork_opus.c artwork_ogg.h
```
Oleksiy-Yakovenko, line 15: typo: should be artwork_opus.h

Oleksiy-Yakovenko (changes requested):
Thanks, it looks really good!
There's one small change request, and I'll need to do some testing, but otherwise it's good to go.

eko5z:
No problem. Also, is there a way to indent the code by some DeaDBeeF standard? I just used the default Emacs indentation, which might be bad.

plugins/artwork/artwork_ogg.c
```diff
@@ -73,8 +73,7 @@ int
 ogg_extract_art (ddb_cover_info_t *cover) {
     int err = -1;
     if (!strcasestr (cover->priv->filepath, ".ogg")
-        && !strcasestr (cover->priv->filepath, ".oga")
```
Oleksiy-Yakovenko, line 76: I'm not sure this is right. Seems like oga can be used for either vorbis and opus. Why remove?

eko5z: It's still there, I only removed `.opus'. It's just that git is showing the diff in a weird way.

Oleksiy-Yakovenko:
merged! thanks for great job!

eko5z:
My pleasure!

PR #3196, opened 2025-06-01
gtkui: fix auto-naming playlists regression (#3195)

### LINKED ISSUE #3195: gktui: auto-naming playlists regression (5c5d44a)
After commit 5c5d44a adding a folder to a playlist that already has a title will always rename the playlist to the folder name. This is because `gtkpl_add_dirs` checks for the playlist having a title like this
https://github.com/DeaDBeeF-Player/deadbeef/blob/5c5d44afb13a0d0008e859305848dda7686b2a0f/plugins/gtkui/fileman.c#L48-L49
but `plt_get_title` _always_ returns 0 if called with a non-`NULL` buffer for the title:
https://github.com/DeaDBeeF-Player/deadbeef/blob/5c5d44afb13a0d0008e859305848dda7686b2a0f/src/playlist.c#L672-L684 So the check can never fail, and the playlist will *always* be renamed.

rsekman:
`plt_get_title` always returns 0 if called with a non-`NULL` buffer for the title; otherwise it returns the length of the title. This causes #3195, and is unintuitive. We can fix #3195 by always returning the length of the title. The only other place in the codebase that uses the return value from this function is https://github.com/DeaDBeeF-Player/deadbeef/blob/5c5d44afb13a0d0008e859305848dda7686b2a0f/plugins/cocoaui/Utility/PlaylistContextMenu.m#L106 where it is called with a `NULL` buffer.  So this change should be safe.

### DIFF
--- a/src/playlist.c
+++ b/src/playlist.c
@@ -672,15 +672,13 @@ plt_get_idx_of (playlist_t *plt) {
 int
 plt_get_title (playlist_t *p, char *buffer, int bufsize) {
     LOCK;
-    if (!buffer) {
-        int l = (int)strlen (p->title);
-        UNLOCK;
-        return l;
+    int l = (int)strlen (p->title);
+    if (buffer) {
+        strncpy (buffer, p->title, bufsize);
+        buffer[bufsize - 1] = 0;
     }
-    strncpy (buffer, p->title, bufsize);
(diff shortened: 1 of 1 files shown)


Oleksiy-Yakovenko:
The bug is valid, but your understanding of the root cause seems to be incorrect. I'll make a fix.

PR #3224, opened 2025-10-03
Add option to always resume session in paused state (fixes #3222)

### LINKED ISSUE #3222: Always start DeaDBeeF in paused state
Is it possible to start the program with the track stopped despite the state of the previous run?
Now it restores the previous state: if some track was playing on exit, then it resumes instantly, when the program starts; if it was paused, then the program starts silently. I'm aiming to start DeaDBeeF silenced every time it starts.
I checked through the GUI and found no option for this behavior.
Currently the sole way I see to achieve that is switching "resume.paused" to 1 in the config file after each run.

kuba160:
Cocoaui settings entry is missing. I guess translations would need to be updated as well.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -838,7 +838,7 @@ save_resume_state (void) {
     int playtrack = -1;
     int playlist = -1;
     playlist_t *plt = pl_get_playlist (trk);
-    int paused = (output->state () == DDB_PLAYBACK_STATE_PAUSED);
+    int paused = (output->state () == DDB_PLAYBACK_STATE_PAUSED) || (conf_get_int ("resume_always_paused", 0));
     if (trk && plt) {
         playlist = plt_get_idx_of (plt);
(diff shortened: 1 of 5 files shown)


src/main.c
```diff
@@ -838,7 +838,7 @@ save_resume_state (void) {
     int playtrack = -1;
     int playlist = -1;
     playlist_t *plt = pl_get_playlist (trk);
-    int paused = (output->state () == DDB_PLAYBACK_STATE_PAUSED);
+    int paused = (output->state () == DDB_PLAYBACK_STATE_PAUSED) | (conf_get_int ("resume_always_paused", 0));
```
Oleksiy-Yakovenko, line 841: This should be `||` instead of `|`

Oleksiy-Yakovenko (changes requested):
Looks good, I'll need to test before merging though.
Please fix the `|` :)

Oleksiy-Yakovenko (approved):

PR #3067, opened 2024-04-02
Add an "enqueue" action to the context menus

dark-penguin:
Add an "Enqueue" action to the context menus of files and also directories on Linux.

PR #3230, opened 2025-10-18
Add an "enqueue" action to the context menus of files and directories

dark-penguin:
A second attempt at #3067 .

This time, I've managed to build it and make sure that it actually works as intended. :slightly_smiling_face:


Oleksiy-Yakovenko:
Tested and merged, seems to work as intended.

PR #3236, opened 2025-10-26
cocoaui: Add Sidebar toggle button and tracking separator

nevack:
Do not autosave toolbar configuration as:
1. Toolbar user customizations are not allowed in deadbeef right now.
2. Toolbar is built programmatically.

Demo:

https://github.com/user-attachments/assets/5e83e179-409d-4334-a99c-6ade32b5a7d9


nevack:
Partially depends on https://github.com/DeaDBeeF-Player/deadbeef/pull/3234
Without https://github.com/DeaDBeeF-Player/deadbeef/pull/3234 and with "Icon and Text" mode enabled, toolbar can overlap sidebar content.

image

PR #3255, opened 2025-11-04
cocoaui: Modernize DdbTabStrip look - Part 2

nevack:
##### Demo:
Screenshot 2025-11-04 at 16 14 18

I think it will need some refinement for non-Liquid Glass look as there would be a separator line between toolbar and content view.

### DIFF
--- a/plugins/cocoaui/DdbTabStrip/DdbTabStrip.m
+++ b/plugins/cocoaui/DdbTabStrip/DdbTabStrip.m
@@ -39,7 +39,7 @@ static const int text_left_padding = 24;
 static const int text_right_padding = 24;
 
 static const int tab_overlap_size = 0;
-static const int tabs_left_margin = 0;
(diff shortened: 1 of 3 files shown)


plugins/cocoaui/DdbTabStrip/DdbTabStrip.m
```diff
@@ -124,7 +124,7 @@ - (void)updateDrawingConfiguration {
     self.isKeyWindow = self.window.isKeyWindow;
 
     NSMutableParagraphStyle *textStyle = [NSParagraphStyle.defaultParagraphStyle mutableCopy];
-    textStyle.alignment = NSTextAlignmentLeft;
+    textStyle.alignment = NSTextAlignmentCenter;
```
nevack: Fixes text "jumping" when selecting and deselecting tab

PR #3278, opened 2026-01-10
gtkui: Implement mouse forward and back buttons

magusjanus:
Implement use of mouse forward and back buttons to control playback.

* Add handling in `on_mainwin_button_press_event` for buttons 8 and 9/10 (based on [Firefox code](https://github.com/mozilla-firefox/firefox/blob/main/widget/gtk/nsWindow.cpp#L4881-L4893))
* Change playlist's view, header, and tabstrips' button-press events to return `FALSE` in the fallback case, so that events propagate up to `on_mainwin_button_press_event`

I don't have the historical or big-picture context for why the playlist button-press events return `TRUE` in the fallback case, let me know if there's a better way to do this.

I only changed the playlist button-press events because that's the main widget I use, I don't know if I should make the same change for other widgets as well.

### DIFF
--- a/plugins/gtkui/callbacks.c
+++ b/plugins/gtkui/callbacks.c
@@ -522,6 +522,12 @@ on_mainwin_button_press_event (GtkWidget *widget, GdkEventButton *event, gpointe
         if (event->type == GDK_2BUTTON_PRESS) {
             deadbeef->sendmessage (DB_EV_TRACKFOCUSCURRENT, 0, 0, 0);
         }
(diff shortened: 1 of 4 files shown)


Oleksiy-Yakovenko:
Tested: it works on linux, but doesn't work on mac.
event->button is 4 and 5 for back and forward respectively.
I don't know whether it can be done any better though.
(I will merge anyway, since deadbeef has Mac-native UI anyway, which is unaffected)

PR #3295, opened 2026-02-11
gtkui: Build fix

ToadKing:
No idea why it started happening but I started getting this error trying to build on Linux:

```
trkproperties.c:912:15: error: initializing 'char *' with an expression of type 'const char *' discards qualifiers [-Werror,-Wincompatible-pointer-types-discards-qualifiers]
  912 |         char *semicolon = strchr(svalue, ';');
      |               ^           ~~~~~~~~~~~~~~~~~~~
```

Fixing this makes the build succeed.

### DIFF
--- a/plugins/gtkui/trkproperties.c
+++ b/plugins/gtkui/trkproperties.c
@@ -909,7 +909,7 @@ _semicolon_separated_string_for_meta(DB_metaInfo_t *meta) {
 static void
 _iterate_semicolon_separated_substrings(const char *svalue, void (^completion_block)(const char *item)) {
     while (*svalue) {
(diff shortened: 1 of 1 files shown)


ToadKing:
Ah, I think I know what caused it: Arch Linux just upgraded to glibc 2.43 which has [this change](https://sourceware.org/pipermail/libc-alpha/2026-January/174374.html):
> The aforementioned change in ISO C23 of the declaration of bsearch, memchr, strchr, strpbrk, strrchr, strstr, wcschr, wcspbrk, wcsrchr, wcsstr, and wmemchr as const-preserving macros can lead to compilation issues in code not set up for it. In particular, gnulib snapshots may need to be updated.

PR #3319, opened 2026-04-02
Add threshold config options for prev_or_restart action

tanguycano:
The restart-vs-go-to-previous threshold in action_prev_or_restart_cb was hardcoded to 3 seconds. This makes it configurable via playback.prev_restart_threshold (float, in seconds), defaulting to 3 to preserve existing behavior.

Increasing this value is useful when triggering the action via "smart shortcuts" or "touch controls" on Bluetooth headsets, as some of these input schemes require multiple or long key presses that take longer to complete.

### DIFF
--- a/src/actionhandlers.c
+++ b/src/actionhandlers.c
@@ -774,7 +774,8 @@ action_prev_or_restart_cb (struct DB_plugin_action_s *action, ddb_action_context
         deadbeef->pl_item_unref (it);
         if (dur > 0) {
             float pos = deadbeef->streamer_get_playpos ();
-            if (pos > 3) {
+            float threshold = deadbeef->conf_get_float ("playback.prev_restart_threshold", 3);
+            if (pos > threshold) {
                 deadbeef->sendmessage (DB_EV_SEEK, 0, 0, 0);
                 return 0;
             }


Oleksiy-Yakovenko:
Hi, Can you modify the PR to just set it to larger value?
Which value is preferred?
I think 3 seconds was quite arbitrary, and I understand it's not optimal.
But adding a configuration option for that is an overkill.

tanguycano:
I have it set to 10s on my machine, but that's also pretty arbitrary.

Some tracks also start slow, silent or generic enough that you can't actually tell whether you've rewound or skipped to the previous track until many seconds into the playback. It really depends on the content you are listening to.

Therefore, I'd still push for a configuration option, but it obviously is your call. Else, I will just rewrite the constant to a bigger value, such as 10s.

Oleksiy-Yakovenko:
I decided not to wait, and committed the change to 10 sec.
Sorry for not taking your PR as is - I'm just not a fan of secret config options without UI configuration.
(Also, as I stated, this behavior doesn't really need to be configurable)

tanguycano:
No worries, I should have resubmitted it with a 10s hard-coded value anyway.
Thanks for taking the time to make the change.

PR #3335, opened 2026-05-23
cocoaui: let playlist key events fall back to menu shortcuts

fff7d1bc:
After updating to macOS 26.5 and a master DeaDBeeF build, I started seeing the Cocoa UI playback hotkeys behave inconsistently. Shortcuts like `b` for Next and `n` for Play Random would work for a while after launch, then stop working until I restarted the player or reset/re-applied the shortcuts in settings.

The issue seemed tied to playlist focus. After clicking in the playlist or using navigation keys like arrows/Home/End, the playlist view becomes first responder. It handles those navigation keys itself, but for other keys it was just falling through to `[super keyDown:]`.

That seems a bit fragile for DeaDBeeF's plain letter shortcuts, because those shortcuts are Cocoa menu key equivalents. If the playlist does not handle a key, the main menu should still get a chance to handle it.

This change makes the playlist view try:

```
[NSApp.mainMenu performKeyEquivalent:theEvent]
```

before falling back to `[super keyDown:]`.

This keeps the existing playlist navigation behavior, but makes menu-backed shortcuts work even when the playlist is focused. I don't know whether macOS 26.5 changed the AppKit behavior here, or whether it only made an existing edge case easier to hit, but this makes the routing explicit instead of relying on the responder chain to recover later.

Verified with:

```
xcodebuild -project osx/deadbeef.xcodeproj -target DeaDBeeF -configuration Release
```

And during runtime, after making this change I did not had another issue with 'b' or 'n' not working. Before it was happening at random but pretty frequently.

### DIFF
--- a/plugins/cocoaui/Playlist/PlaylistContentView.m
+++ b/plugins/cocoaui/Playlist/PlaylistContentView.m
@@ -1037,6 +1037,9 @@ static int grouptitleheight = 22;
             [self scrollPoint:NSMakePoint(vis.origin.x, (self.frame.size.height - sv.contentSize.height))];
             break;
         default:
+            if ([NSApp.mainMenu performKeyEquivalent:theEvent]) {
+                return;
+            }
             [super keyDown:theEvent];
             return;
         }
@@ -1477,4 +1480,3 @@ static int grouptitleheight = 22;
 }
 
 @end
-


Oleksiy-Yakovenko:
Hi,

Can you describe a way to reproduce the problem in more detail?
At first glance, what you did looks like a hack, it's supposed to work without this change, and also I can't reproduce the issue (I've been running 26.5 since day one)

> the playlist view becomes first responder. It handles those navigation keys itself, but for other keys it was just falling through to [super keyDown:].

Well yeah, that [super keyDown] call is supposed to just continue normal handling, i.e. pass the event up the responder chain, which should eventually reach the menu items.
It should be the same behavior as if keyDown was not overridden.

fff7d1bc:
I have no reliable steps to reproduce as it happens at random but daily for me. What happens is that the keys like `n`, `b` or `v` stop doing anything, no next/random/stop, but if I navigate with arrow keys or mouse and swap track, for a short moment the keys will work, if I go to settings -> hotkeys and there `reset to defaults` it does seems to recovery for a little longer but now when I think about it perhaps me interacting with top menu was the workaround. Otherwise restarting player helps.

It very much is a hack, as I suspect it is a bug inside MacOS update that just landed, I am perfectly okay if you are not feeling comfortable merging it, especially so if you did not hit this bug yourself.

Oleksiy-Yakovenko:
I will watch out for this bug for a bit, maybe I'd be able to reproduce, then we'll see.

Oleksiy-Yakovenko:
I'm not observing any issue like what you are describing, even after running the player for weeks uninterrupted.
If this bug still persists for you, I recommend creating a clean user account on your mac, and / or check which 3rd party software you may have installed, which could interfere with keyboard or menu bar functionality.

fff7d1bc:
Thanks for looking into it, I was able to reproduce the problem on two different macOS boxes and narrowed it down to AltTab update, restarting AltTab fixes deadbeef and recently there was huge update to it, so must be somewhere there that is casing the problem.

PR #3347, opened 2026-06-28
UI improvements

markojurkovic:
## Summary

This updates several parts of the macOS Cocoa UI to use the standard system-defined font size instead of explicitly using smaller control fonts.

The goal is to improve readability and make DeaDBeeF better match the native macOS interface, while continuing to rely on AppKit-provided fonts rather than hard-coded point sizes.

It also changes DSP initialization so that no DSP plugin, such as SuperEQ, is automatically added to the DSP chain. The DSP list remains empty until the user explicitly adds an effect.

## Changes

* Use the default macOS menu font size for the playlist-item context menu.
* Use the same default system font size for playlist rows.
* Use a bold version of the default menu font for playlist column headings.
* Increase the bottom status-bar font to the default system font size.
* Increase the font size of the URL field in **File → Add Location**.
* Adjust the Add Location dialog layout so that the URL field remains vertically aligned with the “Location” label.
* Rename the playlist deletion action from the standard `delete:` selector to a DeaDBeeF-specific selector.

  * This prevents AppKit from automatically adding a Delete icon.
  * It also avoids reserving an otherwise empty icon column for the other context-menu items.

### DSP defaults
Do not automatically insert DSP plugins into the DSP chain.
In particular, SuperEQ is no longer added by default.
Users can still add DSP plugins manually when needed.

This gives users a clean, unmodified signal path by default and avoids enabling audio processing without an explicit user choice.

## Rationale

Several Cocoa UI elements explicitly used `smallSystemFontSize` or a small control-content font. On current macOS versions, this makes playlist content, menus and supporting UI noticeably smaller than equivalent native controls.

Using AppKit’s system font APIs with their default size:

* follows the user’s macOS appearance and accessibility settings;
* avoids hard-coded font sizes;
* improves consistency between DeaDBeeF and other macOS applications;
* keeps the UI adaptable across macOS versions.

The playlist column headings remain bold to preserve their visual hierarchy.

The DSP change follows the principle that audio processing should be opt-in. A newly initialized DSP chain should not contain effects that the user did not explicitly select.

## Scope

These changes affect only the macOS Cocoa UI. Other platform frontends are unchanged.

## Testing

Tested in a local macOS build:

* playlist rows render correctly at the updated font size;
* selected and unselected rows use consistent sizing;
* playlist column headings remain vertically centred and are not clipped;
* the playlist context menu and its submenus use the standard menu font size;
* context-menu items no longer have an unnecessary empty icon column;
* the Delete action and Delete-key shortcut continue to work;
* the status bar remains correctly positioned;
* the Add Location URL field accepts and displays text correctly;
* the URL text is vertically aligned with the “Location” label;
* the Add Location dialog buttons and sheet layout remain unaffected.
* the DSP list is not automatically populated with SuperEQ or another DSP plugin;
* DSP plugins can still be added manually.

### DIFF
--- a/plugins/cocoaui/AppDelegate.m
+++ b/plugins/cocoaui/AppDelegate.m
@@ -622,8 +622,16 @@ main_cleanup_and_quit (void);
 }
 
 - (IBAction)addLocationAction:(id)sender {
+    CGFloat fontSize = [NSFont menuFontOfSize:0].pointSize;
+
+    // Same size as the menus, but with proper text-field metrics.
+    _addLocationTextField.controlSize = NSControlSizeRegular;
+    _addLocationTextField.font = [NSFont systemFontOfSize:fontSize];
+
     _addLocationTextField.stringValue = @"";
-    [_mainWindow.window beginSheet:_addLocationPanel completionHandler:^(NSModalResponse returnCode) {
+
+    [_mainWindow.window beginSheet:_addLocationPanel
+                completionHandler:^(NSModalResponse returnCode) {
         if (returnCode == NSModalResponseOK) {
             NSString *text = [self.addLocationTextField.stringValue stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceAndNewlineCharacterSet]];
 
--- a/plugins/cocoaui/MainWindowController.m
+++ b/plugins/cocoaui/MainWindowController.m
@@ -113,6 +113,11 @@ extern DB_functions_t *deadbeef;
 
 - (void)windowDidLoad {
     [super windowDidLoad];
+    
+    
+    // Match the standard macOS menu font size.
+    self.statusBar.font = [NSFont menuFontOfSize:0];
+    [self.statusBar invalidateIntrinsicContentSize];
 
     // This doesn't work reliably when set in XIB
 #if defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && __MAC_OS_X_VERSION_MAX_ALLOWED >= 101600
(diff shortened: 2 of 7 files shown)


Oleksiy-Yakovenko:
Sorry, but no -- none of these changes are acceptable.
I may agree to revert small font size in the menus. The rest of the changes are just not aligned with how I would prefer it, and some are just breaking functionality (e.g. removing supereq from dsp chain)

markojurkovic:
OK, could you explain how removing the SuperEQ DSP is breaking functionality? The goal is to have the cleanest possible audio signal path for those of us who aim for bit perfect playback for example, and having an EQ component forced to be in the chain, whether one uses it or not, seems unnecessary

Oleksiy-Yakovenko:
Because equalizer would no longer work. There's zero reasons for removing it from dsp chain, since there's no overhead when it's off, but I understands it triggers OCD. However, if it causes EQ to stop working - it will confuse a lot of people, which will trigger a lot of bug reports which would be solvable in only one way.

Oleksiy-Yakovenko:
Most other changes would be ok from UX perspective, if they were configurable, but the code itself is unacceptable mess, please review your own changes before posting pull requests.

markojurkovic:
OK, other than the commented out lines which would need to be removed, not sure what parts are "a mess". But anyway, feel free to take what you want from my PR if you like.

Oleksiy-Yakovenko:
Yes, I meant left over commented out code, non-configurable changes in styling, questionable changes without any necessity, and of course the bot comments as well.

PR #3367, opened 2026-09-11
Let the gtk3 plugin build with clang-23.

frederick-bruckman:
Without this, clang-23.1.1 has this to say (libgio-2.0-dev-2.89.4-2+b1, Debian testing)...

/usr/include/glib-2.0/gio/gdbusactiongroup.h:36:23: error: redefinition of typedef 'GDBusActionGroup' is a C11 feature [-Werror,-Wtypedef-redefinition]
   36 | G_DECLARE_FINAL_TYPE (GDBusActionGroup, g_dbus_action_group, G, DBUS_ACTION_GROUP, GObject)
      |                       ^
/usr/include/glib-2.0/gio/giotypes.h:53:47: note: previous definition is here
   53 | typedef struct _GDBusActionGroup              GDBusActionGroup;
      |                                               ^
1 error generated.


Oleksiy-Yakovenko:
It looks like you're trying to fix the symptom instead of the root cause. I would prefer to keep -std c99, there should not be the need to bump to c11 everywhere. I would be fine if you can come up with a simpler fix, or you could wait until I investigate and come up with my own fix.

frederick-bruckman:
Thanks for your attention. It seems I can't currently reproduce the problem! Nevermind.

